PR #712, opened 2026-05-17
sec: harden credential encryption at rest (audit + KDF + rotation)

chrisgeo:
## Summary
- Links CHAOS-1552.
- Audits SQLAlchemy secret storage and commits `docs/security/credential-encryption-audit-2026-05-17.md`.
- Hardens credential encryption with PBKDF2-HMAC-SHA256, explicit salt support, `v1:` ciphertext prefixing, and legacy v0 decrypt fallback.
- Adds a v0→v1 re-encryption utility plus key rotation runbook.

## Audit findings
- `settings.value` is encrypted when `is_encrypted=true`; risk is caller misuse.
- `integration_credentials.credentials_encrypted` is encrypted via IntegrationCredentialsService.
- `sso_providers.encrypted_secrets` is encrypted for service-managed writes; plaintext compatibility fallback remains for legacy rows.
- Token tables store hashes only, not plaintext tokens.
- Billing/subscription tables store provider IDs/audit state; no credential columns identified.

## Verification
- `pytest tests/core/test_encryption.py -x -q` ✅ 9 passed
- `pytest tests/api/services/ -x -q -k "settings or sso"` ✅ 13 passed
- `lsp_diagnostics` on changed Python files ✅ clean
- `alembic upgrade head` attempted against SQLite and configured Postgres; SQLite hit existing 0001 SQLite-incompatible DDL, Postgres connection was unavailable in local environment. No Alembic migration is included because no plaintext secret column requiring schema/data migration was identified.

SCREENSHOT-WAIVER: backend encryption/docs/tooling only; no rendered frontend output changes.

### DIFF
--- a/docs/security/credential-encryption-audit-2026-05-17.md
+++ b/docs/security/credential-encryption-audit-2026-05-17.md
@@ -1,0 +1,84 @@
+# Credential encryption at rest audit — 2026-05-17
+
+Scope: SQLAlchemy models under `src/dev_health_ops/models/`, with focused review of settings, SSO, credential, token, billing, subscription, and identity-related storage. This audit reframes CHAOS-1552 as hardening: encryption already existed in `dev_health_ops.core.encryption` and callers use it for the primary credential paths.
+
+## Summary by table
+
+| Table | Secret-bearing column(s) | Encrypted at rest? | Risk | Action |
+| --- | --- | --- | --- | --- |
+| `settings` | `value` when `is_encrypted=true` | Yes, through `SettingsService.set(..., encrypt=True)` → `encrypt_value`; read via `SettingsService.get` → `decrypt_value` | Medium | Harden KDF/versioning; operators must set `encrypt=True` for sensitive settings. |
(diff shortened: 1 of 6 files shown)


chrisgeo:
CodeQL alert 856 (`py/clear-text-logging-sensitive-data`) addressed in 5f3d70f1.

The sink was `core/encryption.py:90` — `logger.error("Unsupported encrypted value version: %s", version)`. Even though `version` is just a short prefix string like `"v2"` parsed via `.split()`, CodeQL's dataflow correctly taints any value reachable from a secret column (the four reported sources: `scripts/reencrypt_settings_credentials.py:127`, `api/auth/sso/router.py:1363`, `api/services/sso.py:971` and `:983`).

Fix: drop the log statement; the raised `ValueError("Decryption failed - unsupported ciphertext version")` already conveys the failure to callers and operators without leaking anything ciphertext-derived. Added an explanatory comment so the log statement doesn't get re-added.

Tests still green (`pytest tests/core/test_encryption.py -q` → 9 passed). LSP clean. CodeQL will re-run on this push.

PR #737, opened 2026-05-18
feat(metrics): compute AI workflow impact metrics (CHAOS-1581)

chrisgeo:
## Summary
Adds AI workflow impact metrics computed from `ai_attribution` (CHAOS-1579/1580). Queryable by team, repo, time range, work type. AI vs non-AI baselines tracked separately. Null/unknown attribution is a distinct bucket — never silently treated as human.

## Metrics
- AI-assisted PR ratio
- Agent-created PR count
- AI cycle-time delta
- AI review amplification
- AI rework drag
- AI revert + incident drag
- AI test gap rate
- AI Operating Leverage (decomposed — exposes its components)

## Design
- Migration 036_ai_metrics.sql (ReplacingMergeTree; dims org_id, team_id, repo_id, work_type, date)
- Compute jobs registered into `dev-hops metrics daily`
- Read path uses `ai_attribution_resolved`
- Unknown attribution kept as a separate bucket

## Verification
```
uv run pytest tests/metrics/test_ai_impact*
uv run ruff check
uv run mypy src
```

SCREENSHOT-WAIVER: backend-only.

Closes CHAOS-1581

### DIFF
--- a/src/dev_health_ops/metrics/ai_impact.py
+++ b/src/dev_health_ops/metrics/ai_impact.py
@@ -1,0 +1,364 @@
+from __future__ import annotations
+
+import uuid
+from collections import defaultdict
+from collections.abc import Callable, Mapping, Sequence
+from dataclasses import dataclass
+from datetime import date, datetime, timedelta, timezone
+
+from dev_health_ops.metrics.schemas import (
+    AIImpactMetricsDailyRecord,
+    AIOperatingLeverageComponents,
+    AIPullRequestAttributionRow,
+    CommitStatRow,
+    IncidentRow,
+    PullRequestReviewRow,
+    PullRequestRow,
+)
+
+TeamResolver = Callable[[str, str | None, str | None], tuple[str | None, str | None]]
+
+AI_BUCKETS = {"ai_assisted", "agent_created", "ai_review"}
+NON_AI_BUCKET = "human"
+UNKNOWN_BUCKET = "unknown"
+
+
+@dataclass(frozen=True)
+class _PRFact:
+    repo_id: uuid.UUID
+    number: int
+    bucket: str
+    work_type: str
+    team_id: str | None
+    merged: bool
+    cycle_hours: float | None
+    reviews: int
+    changes_requested: int
+    additions: int
+    deletions: int
+    changed_files: int
+    has_test_change: bool
+    followup_commits: int
+
+
+@dataclass(frozen=True)
+class _Agg:
+    prs_total: int
+    prs_merged: int
+    cycle_avg: float | None
+    reviews_per_pr: float | None
+    changes_requested_per_pr: float | None
+    rework_prs: int
+    rework_rate: float | None
+    followup_commits: int
+    revert_prs: int
+    revert_rate: float | None
+    incidents_count: int
+    incident_rate: float | None
+    test_gap_prs: int
+    test_gap_rate: float | None
+
+
+def _to_utc(value: datetime) -> datetime:
+    if value.tzinfo is None:
+        return value.replace(tzinfo=timezone.utc)
+    return value.astimezone(timezone.utc)
+
+
+def _avg(values: Sequence[float]) -> float | None:
+    if not values:
+        return None
+    return sum(values) / len(values)
+
+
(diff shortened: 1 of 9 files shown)


chrisgeo:
Pushed [`a506b1e9a`](https://github.com/full-chaos/dev-health-ops/pull/737/commits/a506b1e9a) addressing both findings:

**1. CodeQL alert #868 (`py/missing-call-to-init`)** — `ClickHouseDataLoader.__init__` now delegates to `AIImpactClickHouseLoader` via `super().__init__(client, org_id=org_id)` instead of duplicating the field assignments. Removed the now-unused `OrgScopedQuery` import.

**2. live-e2e migration failure (`Code 44: ILLEGAL_COLUMN — Sorting key contains nullable columns`)** — `team_id` in `ai_impact_metrics_daily` is now non-nullable `String` with empty-string sentinel for 'no team association'. This matches the established repo pattern (see `035_ai_attribution.sql` ORDER BY, which uses non-nullable sentinels). Python `str | None` semantics preserved at the boundary: sink writes `''` when record value is `None`, loader maps `''` back to `None` on read.

Local gates clean: `9 passed`, `ruff check` passed, `ruff format --check` passed, `mypy` clean.

chrisgeo:
Rebased onto latest `origin/main` (now includes #735 + #736). Two conflicts, both clean additive overlap with the now-merged #736 (CHAOS-1587 governance):

1. `src/dev_health_ops/metrics/sinks/clickhouse/__init__.py` — kept `AIGovernanceMixin` import + base, added `AIImpactMixin` alongside (alphabetical: attribution → governance → impact).
2. `src/dev_health_ops/metrics/job_daily.py` — kept #736's governance build + write calls, added this PR's AI-impact compute + write right after, both inside the same per-day loop.

Force-pushed with `--force-with-lease`. Local gates re-green: `9 passed`, `ruff check` passed, `ruff format --check` passed, `mypy` clean.

chrisgeo:
Sorry — my previous fix's migration comment broke the runner. The dev-health-ops CH migration runner can't handle `;` inside `--` comments (see prior fix `d19bc1e78`). My comment had `(sentinel; sorting key cannot be Nullable)` which the runner split mid-comment, producing the new SYNTAX_ERROR with unmatched `(`.

Amended the comment to remove the `;` and parentheses around the second clause. No other code changed. Force-pushed.

chrisgeo:
Third fix in this thread — apologies for the noise. Migration runs clean now, but live-e2e exposed two query-shape issues in the AI attribution loader that my earlier batch-edit lost on a hash-mismatch atomic rollback (only the line-178 retry actually landed; lines 48/68/73/92 never got their edits):

1. `attr.subject_id IN (toString(pr.number), concat(toString(pr.repo_id), '#', toString(pr.number)))` — ClickHouse rejects `IN` when the RHS is a tuple of non-constant column expressions (`UNSUPPORTED_METHOD`, code 1). Rewritten as `(attr.subject_id = toString(pr.number) OR attr.subject_id = concat(...))`.
2. `CAST(NULL, 'Nullable(String)') AS team_id` (both UNION branches) — aligned to the new String-with-sentinel schema: `CAST('', 'String') AS team_id`.
3. The accompanying boundary normalization in `load_ai_pr_attributions` (`raw.get('team_id') or None`) now also lands.

Amended the fix commit (force-pushed with `--force-with-lease`). All local gates re-green.

chrisgeo:
Fix #4 — TypeError 'TeamResolver object is not callable'. The compute fn declared `TeamResolver` as a `Callable[[str, str|None, str|None], tuple[str|None, str|None]]` (function alias), but `job_daily` was passing the dataclass-based `TeamResolver` from `providers/teams.py` (member→team via `.resolve()` method). Two type mismatches in one parameter.

Fixed at the call site: wrapped the correct `repo_team_resolver` (RepoPatternTeamResolver) in a lambda matching the expected signature: `lambda _repo_id, repo_name, _identity: repo_team_resolver.resolve(repo_name)`. This is consistent with how `repo_team_resolver` is passed to neighboring compute fns in the same loop (lines 351, 365, 402, 410, 418).

Local gates green. Force-pushed.

PR #765, opened 2026-05-21
test(graphql): EXPLAIN-based SQL validation for every resolver (CHAOS-1752)

chrisgeo:
Adds tests/api/graphql/test_resolver_sql_explain.py — a single
parametrized test that replays every resolver-built parameterized
query through ClickHouse EXPLAIN PLAN against the production DDL,
without executing the data path. EXPLAIN validates parse, table and
column existence, function signatures, aggregation legality, and
parameter binding shape — exactly the surface that the CHAOS-1751
resolver bugs slipped past.

Bugs the test catches by construction
-------------------------------------
* CHAOS-1751 #2 — argMax aliases shadowing the source column they
  reference (ILLEGAL_AGGREGATION, code 184).
* CHAOS-1751 #3 — column references against a table whose schema
  uses different names (UNKNOWN_IDENTIFIER, code 47).

Validated by `git revert HEAD~3..HEAD` simulation on the
compounding_risk resolver — both bugs fire the test as expected.

Mechanism
---------
* `tests/api/graphql/_sql_explain_helpers.py` — `CapturingSink`
  records `(sql, params)` calls without executing them. Slots in
  via `query_dicts`'s `hasattr(sink, "query_dicts")` path.
* `tests/api/graphql/sql_explain_fixtures.py` — `ALL_RESOLVER_SQL_FIXTURES`
  is a per-resolver registry of async functions that invoke each
  SQL-emitting helper with representative arguments (cover every
  WHERE/branch difference).
* `tests/api/graphql/test_resolver_sql_explain.py` — session-scoped
  fixture connects to ClickHouse (`CLICKHOUSE_URI`), applies the
  production DDL once via `ClickHouseMetricsSink.ensure_schema()`,
  and runs `EXPLAIN PLAN {sql}` for every captured query.

Coverage
--------
11 resolver fixtures: compounding_risk, forecast, home, capacity,
work_graph, recommendations, operating_review, security, bus_factor,
data_health, analytics. Suite runs in ~1.5s locally end-to-end
against a fresh CH container.

Pre-existing bugs surfaced and fixed
------------------------------------
The new test immediately caught several latent runtime bugs in
unrelated resolvers. Fixed in-PR because they are surgical and follow
the CHAOS-1751 pattern:

* `home`: `max(ingested_at)` against `investment_metrics_daily`
  (column doesn't exist; should be `computed_at`).
* `operating_review`:
  - `FROM work_item_state_duration_daily` (table is plural:
    `work_item_state_durations_daily`).
  - `WHERE risk_score > 0` clashing with outer `avg(risk_score) AS
    risk_score`; rename the inner alias to `latest_risk_score` and
    move the predicate into HAVING on the inner aggregate.
* `recommendations`: `argMax(window_end, computed_at) AS window_end`
  (and friends) clashing with source columns referenced in WHERE /
  GROUP BY / HAVING. Rename SELECT aliases to `latest_*` and update
  `_row_to_recommendation` accordingly (with backward-compat fallback
  for any callers that still pass un-aliased rows).
* `analytics`: `Dimension.WORK_TYPE` was not in `investment_dims`,
  so non-forced `use_investment=False` requests routed it against
  `investment_metrics_daily` looking for a non-existent
  `work_item_type` column. WORK_TYPE belongs to the
  `work_unit_investments` rollup; add it to the auto-route set.

Remaining work tracked in CHAOS-1754: the `THROUGHPUT`, `CHURN_LOC`,
and `CYCLE_TIME_HOURS` measure mappings on `work_unit_investments`
reference columns that don't exist on that table. Until that lands,
the analytics fixture matrix is narrowed to `count` against
investment dimensions.

CI
--
`.github/workflows/test.yml`: add `CLICKHOUSE_URI` to the test-matrix
env (in addition to the existing `DATABASE_URI`) so the EXPLAIN
fixture picks up the already-provisioned ClickHouse service.

Contract
--------
Every new GraphQL resolver that emits SQL MUST register a fixture in
`ALL_RESOLVER_SQL_FIXTURES`. Documented in
`docs/api/graphql-overview.md` under "Resolver SQL Validation
Contract (CHAOS-1752)".

SCREENSHOT-WAIVER: backend-only change; no rendered UI surface.

Closes CHAOS-1752

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -91,6 +91,7 @@ jobs:
       - name: Run tiered test contract
         env:
           DATABASE_URI: clickhouse://ch:ch@localhost:8123/default
+          CLICKHOUSE_URI: clickhouse://ch:ch@localhost:8123/default
           POSTGRES_URI: postgresql+asyncpg://postgres:postgres@localhost:5432/test_db
           SECONDARY_DATABASE_URI: mongodb://localhost:27017
           COVERAGE_THRESHOLD: "70"
--- a/docs/api/graphql-overview.md
+++ b/docs/api/graphql-overview.md
@@ -305,3 +305,51 @@ See also: [Persisted Queries](persisted-queries.md).
 ## Web client
 
 See: [Web GraphQL Client](web-graphql-client.md).
+
+
+---
+
+## Resolver SQL Validation Contract (CHAOS-1752)
+
+Every GraphQL resolver that emits ClickHouse SQL **must** register a fixture
+in `tests/api/graphql/sql_explain_fixtures.py`. The fixture exercises each
+SQL-emitting helper with representative arguments; `tests/api/graphql/`
+`test_resolver_sql_explain.py` replays the captured queries through
+`EXPLAIN PLAN` against a real ClickHouse with production DDL applied.
+
+EXPLAIN PLAN validates:
+
+- SQL parse
+- Table existence
+- Column existence (catches mistyped column references)
+- Function signature resolution
+- Aggregation legality (catches alias-collision bugs like CHAOS-1751 #2)
+- Parameter binding shape
+
+It does **not** execute the data path, so the full suite runs in <30s.
+
+### Adding coverage for a new resolver
+
+1. Implement the resolver. Keep SQL-emitting work in helper functions or
+   compile functions that accept a sink/client — not embedded in HTTP
+   handlers.
+2. Add a fixture entry to `ALL_RESOLVER_SQL_FIXTURES`:
+
+   ```python
+   async def _fixture_<name>(sink: CapturingSink) -> None:
+       context = FakeGraphQLContext(client=sink, org_id=SAMPLE_ORG_ID)
+       await my_resolver_helper(context, ...representative args...)
+       # Cover every WHERE/JOIN branch the helper can produce.
+
+   ALL_RESOLVER_SQL_FIXTURES = [
+       ...,
+       ("<name>", _fixture_<name>),
+   ]
+   ```
+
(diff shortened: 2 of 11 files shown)


chrisgeo:
## submit-pypi failure — pre-existing upstream conflict, not introduced by this PR

The `Automatic Dependency Submission (Python)` job is failing on `pip-compile` resolution, **not** on anything this PR added.

**Diagnosis**

The latest `opentelemetry-instrumentation-celery` released (`0.45b0`) pins its transitive `opentelemetry-instrumentation` to `==0.45b0`. Every released `opentelemetry-instrumentation-fastapi` from `0.46b0` through `0.63b0` pins the same transitive dep to its matching version (`==0.46b0` … `==0.63b0`). With both `fastapi` and `celery` instrumentation listed under `>=0.45b0` in `pyproject.toml`, the system pip resolver picks the latest of each and explodes on the pinned mismatch. `uv`'s resolver handles it locally by picking matching versions; pip's does not.

**Evidence it's pre-existing / upstream**

- Local `uv run pip-compile pyproject.toml` against this branch succeeds.
- This PR doesn't touch `pyproject.toml` dependencies (only added a `[[tool.mypy.overrides]]` block for sibling test imports).
- PRs #761, #762, #763 all passed `submit-pypi` — the upstream metadata that triggers this only landed recently.

**Impact on merge**

`submit-pypi` is **not** a required status check on `main` (only `test` and `lint` are required per the branch ruleset). All required checks pass. Resolving the underlying conflict is a separate piece of work — either bump `opentelemetry-instrumentation-celery` (waiting on upstream) or pin `opentelemetry-instrumentation-fastapi` to `==0.45b0` in `pyproject.toml`. I can file a follow-up Linear ticket if that's worth doing.

PR #849, opened 2026-06-11
fix(licensing): resolve org tier via Organization.tier fallback (CHAOS-2256)

chrisgeo:
## Summary
Fixes [CHAOS-2256](https://linear.app/fullchaos/issue/CHAOS-2256): an Enterprise org setting a 1h sync interval was rejected with the **community** 24h floor because `TierLimitService` (and `FeatureService`) resolved tier from the `OrgLicense` table only, silently defaulting to COMMUNITY when no row exists.

An org can legitimately be Enterprise with no `OrgLicense` row: tier set on `Organization.tier` at creation / by an admin (`OrganizationService.update`), or a trial before the Stripe plan resolves (`_sync_org_license` returns early when `billing_plan_id` is None). The entitlements path (`gating.get_org_entitlements_from_db`, `_check_org_feature_async`) already falls back to `Organization.tier` — the limits path didn't.

## Change
- New `resolve_org_tier(session, org_id, org_license)` in `api/services/licensing.py`: `OrgLicense.tier` wins when a row exists; else fall back to `Organization.tier`; default COMMUNITY. Mirrors the entitlements path exactly (including invalid-tier warning behavior).
- Used by `TierLimitService.get_limit` / `get_all_limits` and `FeatureService.check_feature_access` — fixes all 7 tier-limit gates in `admin/routers/sync.py` (max_repos, backfill_days, min_sync_interval_hours ×2, max_work_items, backfill trigger) in one place.

## Feature-gate inventory (audit performed for this fix)
- **7 TierLimitService call sites** (all in `admin/routers/sync.py`) — were BUGGY, all fixed by this change.
- **31 `@require_feature` decorators** (SSO, IP allowlist, retention, audit logs) — all async → already get the per-org `_check_org_feature_async` fallback → OK.
- **Entitlements endpoints** (`/api/v1/licensing/entitlements`, `/api/v1/billing/entitlements`) — already fall back to `Organization.tier` → OK.
- **GraphQL `authz.py` `has_feature`** — instance-level JWT (self-hosted LICENSE_KEY), per-deployment by design.
- No unsafe ad-hoc tier checks found elsewhere.

## Tests
- `test_enterprise_min_sync_interval_is_quarter_hour` — with OrgLicense → 0.25
- `test_enterprise_tier_resolves_via_organization_fallback` — **no OrgLicense row** → 0.25 (fails pre-fix)
- `test_enterprise_without_license_can_set_hourly_schedule` — 1h cron → 201 (fails pre-fix with the exact CHAOS-2256 error)
- `test_enterprise_can_set_hourly_schedule`, `test_enterprise_below_floor_still_rejected` (10-min cron → 403)

## Gates
- `ruff format --check` / `ruff check` clean
- `mypy --install-types --non-interactive .` — 0 issues, 1008 files
- `tests/api/admin/test_tier_limits.py` 14 passed; `tests/test_licensing.py` 95 passed; `tests/test_backfill_integration.py` 14 passed

### DIFF
--- a/src/dev_health_ops/api/billing/subscription_service.py
+++ b/src/dev_health_ops/api/billing/subscription_service.py
@@ -418,6 +418,23 @@ class SubscriptionService:
             if customer_id:
                 assign_attr(org_license, "customer_id", customer_id)
 
+        # Keep the denormalized Organization.tier column in lockstep so the
+        # tier-resolution fallback (resolve_org_tier) and any direct readers of
+        # Organization.tier never see a value that diverges from OrgLicense.
+        try:
+            from dev_health_ops.models.users import Organization
+
+            org_result = await self.db.execute(
+                select(Organization).where(Organization.id == org_id)
+            )
+            org = org_result.scalar_one_or_none()
+            if org is not None and str(org.tier) != tier_str:
+                assign_attr(org, "tier", tier_str)
+        except OperationalError as exc:
+            logger.warning(
+                "organizations table unavailable; skipping org tier sync (%s)", exc
+            )
+
         # Static literal labels keyed by tier string to break the taint chain
         # entirely (CodeQL py/clear-text-logging-sensitive-data).
         tier_label = {
--- a/src/dev_health_ops/api/services/licensing.py
+++ b/src/dev_health_ops/api/services/licensing.py
@@ -22,6 +22,7 @@ from dev_health_ops.models.licensing import (
     OrgLicense,
     TierLimit,
 )
+from dev_health_ops.models.users import Organization
 
 if TYPE_CHECKING:
     from sqlalchemy.orm import Session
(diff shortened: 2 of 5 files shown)


chrisgeo:
## Adversarial review (Codex) addressed — commit 1f39d698b

**Finding (medium):** `resolve_org_tier` gives OrgLicense precedence, so a stale OrgLicense row (e.g. community) would keep overriding a newer `Organization.tier` (enterprise) — the version-skew case the original fix didn't cover.

**Resolution — write-path reconciliation, not read-precedence change.** Flipping precedence would desync limits from entitlements again (the CHAOS-2256 acceptance criterion). Instead, the two write paths that could diverge now keep both columns in lockstep:

- `OrganizationService.update(tier=...)` also updates a pre-existing `OrgLicense` row in the same transaction (admin path — Codex's exact scenario). No row → resolution already falls back to `Organization.tier`.
- `SubscriptionService._sync_org_license` also updates the denormalized `Organization.tier` after upserting `OrgLicense` (Stripe path — the inverse skew). `billing/router.py::_persist_license` already did both.

**New tests:**
- `test_stale_org_license_takes_precedence_and_paths_agree` — pins read behavior for legacy diverged rows: OrgLicense wins AND limits/entitlements paths agree on it.
- `test_admin_tier_update_heals_stale_org_license` — admin upgrade updates the stale license row; `get_limit` then returns 0.25.
- `test_subscription_creates_org_license` extended — Stripe upsert syncs `Organization.tier` to enterprise.

Gates: ruff clean, mypy 0 issues (1008 files), tier-limits 16/16, billing+licensing 152 passed. (`test_org_deletion.py` has 2 failures that reproduce on main — pre-existing, unrelated.)

PR #860, opened 2026-06-11
fix(admin): provider-shaped child sync_options in batch sync-config create (CHAOS-2283)

chrisgeo:
## Summary

Fixes CHAOS-2283: `POST /sync-configs/batch` wrote GitHub-shaped child option keys (`repo=<name>`) for **every** provider. GitLab children were therefore created without the integer `project_id` that the sync runtime requires (`sync_runtime.py` raises `Missing GitLab project_id in sync options`; `sync_batch.py` dispatch also keys GitLab children on `int(project_id)`), so they could never sync.

## Changes

- **Provider-conditional child option construction** in `batch_create_sync_configs`:
  - `github`: unchanged (regression-tested) — child carries `repo` plus inherited `owner`; name stays `{owner}/{repo}`.
  - `gitlab`: child carries int `project_id` + `group` (web form sends the namespace as `owner`; API callers may use `group`), with `gitlab_url` passing through from parent options. GitHub-shaped `owner`/`repo`/`search` keys are dropped. Child name uses `path_with_namespace` when known.
- **Payload contract** (documented in the endpoint docstring): `repos` entries may be numeric GitLab project ids (used as-is, no API call) or project names — the web `RepoSelector` sends names from `DiscoveredRepo.name`. Names are resolved to ids by listing the group's projects via the stored credential (same pattern as `GET /credentials/{id}/repos`, which feeds the selector). Unknown/ambiguous names, or name entries without `credential_id`/group, fail fast with 400 instead of persisting children that can never sync.

## Tests

`tests/api/admin/test_sync_configs.py` (+5):
- GitHub regression: exact child `sync_options` dict unchanged (`owner`/`repo`/`schedule_cron` only).
- GitLab: exact child keys (`project_id` int, `group`, `gitlab_url`, no `repo`/`owner`), names from `path_with_namespace`, connector called with credential token + `gitlab_url`.
- GitLab numeric ids: no credential/API call needed.
- GitLab unknown name → 400 naming the repo.
- GitLab names without credential → 400.

## Gates

- `mypy --install-types --non-interactive .` — clean (1011 files)
- `ruff check` / `ruff format` — clean
- `./ci/run_tests.sh unit` — 3957 passed; only failures are the 4 known pre-existing ignores (test_org_deletion ×2, TestCoreCache ×2)

### DIFF
--- a/src/dev_health_ops/api/admin/routers/sync.py
+++ b/src/dev_health_ops/api/admin/routers/sync.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 import uuid
 from typing import Any, Protocol, cast
 
@@ -34,6 +35,8 @@ from dev_health_ops.models.settings import (
 from dev_health_ops.workers.sync_batch import _is_batch_eligible
 
 from .common import get_session
+
+logger = logging.getLogger(__name__)
 
 router = APIRouter()
 
@@ -137,6 +140,7 @@ PROVIDER_SYNC_TARGETS: dict[str, list[str]] = {
 
 NON_REPO_SYNC_PROVIDERS = {"jira", "linear"}
 DEFAULT_SYNC_CRON = "0 * * * *"
+DEFAULT_GITLAB_URL = "https://gitlab.com"
 
 
 def _sync_options_with_top_level_fields(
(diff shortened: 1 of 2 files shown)


chrisgeo:
Addressed both Codex-review HIGH findings in be531e389:

**1. Credential `gitlab_url` now persisted into parent + child options.** `_resolve_gitlab_batch_projects` resolves the effective instance URL up front — `sync_options.gitlab_url` → decrypted credential `url` → credential `config.url` → `https://gitlab.com` (same precedence as `credentials/resolver.py:resolve_gitlab_url` from #863, implemented locally since that helper isn't on this merge-base) — and returns it alongside the project map. The endpoint persists it into `parent_options` (and therefore every child's options) whenever it differs from the public default, so self-hosted children no longer fall back to the gitlab.com runtime default in `sync_runtime.py`. Resolution also happens for all-numeric batches when a `credential_id` is present.

**2. Numeric entries are cross-checked against the group listing.** When a credential + group are available, the listing now also runs for numeric `repos` entries: an entry matching a listed project name/full_name is resolved as a *name* (ambiguity rules apply), so a project literally named `007` resolves to its real project id instead of being coerced to `project_id=7`. Entries matching no listed name keep id semantics. The no-credential all-numeric escape hatch is preserved and now documented in the endpoint docstring; a listing failure on an all-numeric batch logs a warning and falls back to id semantics instead of failing the request.

Tests added (`tests/api/admin/test_sync_configs.py`): credential self-hosted URL persisted into parent + children when parent options lack `gitlab_url`; numeric entry matching a listed name resolves as name; numeric entry matching nothing resolves as id while the listing is still consulted. 35/35 passing; mypy clean.

PR #862, opened 2026-06-11
fix(clickhouse): add org_id to ReplacingMergeTree dedup keys (CHAOS-2290)

chrisgeo:
## Summary

Fixes the CHAOS-2290 tenant-isolation bug class: ReplacingMergeTree tables whose sorting key (dedup key) lacks `org_id`, so identical natural keys across two orgs collapse to one row on a background merge — silent cross-tenant data loss.

### Survey result: the headline table was already fixed

`009_raw_work_items.sql` creates `work_items` / `work_item_transitions` (there is no table literally named `raw_work_items`). Migration `027_add_org_id_to_sorting_keys.py` already rebuilds both with `org_id` first in the key, and the live DB confirms `work_items → (org_id, repo_id, work_item_id)`. **However**, five RMT tables created after 027 repeat the exact same flaw and are broken on live databases:

| Table | Created by | Old key |
|---|---|---|
| `ci_job_runs` | 029_testops_tables.sql | `(repo_id, run_id, job_id)` |
| `test_suite_results` | 029_testops_tables.sql | `(repo_id, run_id, suite_id)` |
| `test_case_results` | 029_testops_tables.sql | `(repo_id, run_id, suite_id, case_id)` |
| `coverage_snapshots` | 029_testops_tables.sql | `(repo_id, run_id, snapshot_id)` |
| `security_alerts` | 032 (033 added the org_id column, but a key cannot be ALTERed) | `(repo_id, alert_id)` |

### Changes

- **`042_rmt_org_id_dedup_keys.py`** — ClickHouse cannot ALTER an existing RMT key, so this reuses 027's shadow-table pattern (`SHOW CREATE TABLE` → create `<table>_new` with `org_id` prepended → `INSERT ... SELECT` → verify → `EXCHANGE TABLES` → drop). It additionally verifies **distinct sorting-key tuple counts** match before the swap (raw row counts can legitimately differ because the copy collapses not-yet-merged same-key duplicate versions — this fired on real local data during verification). `work_items`/`work_item_transitions` are included as idempotent safety nets for any DB whose 027 run was interrupted; the migration skips tables that already have `org_id` first.
- **`tests/test_migration_042_rmt_org_id_dedup.py`** — unit tests (no DB): string-level assertions that every 042 rebuild key starts with `org_id`, plus a whole-catalog guard that parses every `*.sql` migration, applies the 027/042 overrides, and fails if any RMT table's *effective* key lacks a tenant key (`org_id`/`org_id_hash`). This prevents migration 043+ from reintroducing the bug.
- **`tests/test_rmt_org_id_dedup_live.py`** — opt-in `@pytest.mark.clickhouse` regression tests: two rows identical except `org_id`, `OPTIMIZE TABLE ... FINAL`, assert both survive (work_items, work_item_transitions, security_alerts, test_suite_results). Self-cleaning.

### Live verification

Applied against local `dev-health-clickhouse-1` (real data): all 7 tables now report `org_id`-first sorting keys in `system.tables`, `schema_migrations` records `042`, all 4 live tests pass, no shadow/test artifacts left behind.

### Codex-review hardening (second commit)

- **Racing writes (HIGH):** the rebuild no longer drops the old table right after `EXCHANGE`. It first runs a **catch-up pass** — `INSERT INTO <table> SELECT * FROM <table>_new` — re-inserting ALL old rows, which is idempotent under RMT dedup (newest version wins) and preserves any rows written to the source between the snapshot copy and the swap. The old table is dropped only after the catch-up succeeds. **Residual risk:** writes racing the `EXCHANGE` instant itself can land on either side; both sides are preserved, but version ties on the same key resolve arbitrarily. *Ops note: run this migration during a quiet period / with sync workers stopped.*
- **Crash convergence (MEDIUM):** if a run dies after `EXCHANGE` but before catch-up/`DROP`, the rerun's idempotency-skip path now detects the leftover `<table>_new` (which holds the OLD table), runs the catch-up from it, then drops it — no stranded rows, no leftover shadow. Pre-`EXCHANGE` failures still drop their disposable shadow; the old blanket drop-on-any-failure was removed because it would have destroyed the catch-up delta after a post-swap crash.
- **Fail-closed DDL rewrite (MEDIUM):** after creating the shadow, the migration reads `system.tables.sorting_key` for both tables and verifies the shadow's key is exactly `org_id, <old key>` (normalized string compare) **before copying any data** — a regex miss on nested ORDER BY expressions (`cityHash64(...)`, `ifNull(...)`) now aborts and drops the shadow instead of shipping a wrong dedup key.
- **Join predicates (HIGH, data leak):** added `org_id` equality to the `coverage_snapshots ↔ ci_pipeline_runs` and `git_commits ↔ git_commit_stats` joins in `metrics/loaders/clickhouse.py`. The `ci_job_runs ↔ ci_pipeline_runs` and `test_case_results ↔ test_suite_results` joins are fixed in **PR #857 (CHAOS-2293)**, which also reworks the alias filter handling in those exact hunks — deliberately left untouched here; merge order between #857 and #862 doesn't matter.
- **Tests:** live dedup tests now cover all rebuilt TestOps tables (`ci_job_runs`, `test_case_results`, `coverage_snapshots` added; 7 live tests total), plus FakeClient unit tests for catch-up ordering, post-swap-failure shadow retention, crash convergence, fail-closed key verification, and pre-swap cleanup. The full rebuild → catch-up → convergence flow was also exercised end-to-end against the local ClickHouse with a scratch table (cleaned up).

### Reader audit flags (NOT fixed here — follow-up candidates)

Per the org_id-in-every-join-predicate checklist:

- `src/dev_health_ops/audit/completeness.py` — `build_work_items_query` / `build_transitions_query` / `build_git_commits_query` aggregate with **no org_id filter at all** (cross-tenant counts).
- `src/dev_health_ops/api/graphql/resolvers/security.py` — every `security_alerts` query joins `INNER JOIN repos r ON sa.repo_id = r.id` **without** `AND sa.org_id = r.org_id`; org scoping rides solely on `r.org_id`. Now that cross-org alert rows survive dedup (this fix), a colliding `repo_id` would attach another org's alerts to this org's repo row.
- `src/dev_health_ops/work_graph/investment/queries.py` — `fetch_work_items` / `fetch_parent_titles` only filter by org when the optional `org_id=""` arg is set (unscoped fallback).
- `src/dev_health_ops/metrics/compute_capacity.py` — `get_backlog_size` org filter is conditional (unscoped fallback).
- `src/dev_health_ops/work_graph/builder.py` — `WHERE org_id` applied only `if self.config.org_id` (unscoped fallback; also interpolates org_id via f-string).

Non-RMT (plain MergeTree) tables missing org_id in their keys (e.g. `report_provenance`, 030/031 testops daily/benchmark tables) are a perf/index concern, not a dedup-loss one — left out deliberately.

### Gates

- `uv run mypy --install-types --non-interactive .` — clean (1014 files)
- `./ci/run_tests.sh unit` (unsandboxed) — 3963 passed; only the 4 known-flaky ignores failed (test_org_deletion ×2, TestCoreCache ×2)
- `uv run pytest tests/test_migration_042_rmt_org_id_dedup.py -q` — 11 passed
- Live `pytest -m clickhouse tests/test_rmt_org_id_dedup_live.py` — 7 passed

Linear: CHAOS-2290

### DIFF
--- a/src/dev_health_ops/metrics/loaders/clickhouse.py
+++ b/src/dev_health_ops/metrics/loaders/clickhouse.py
@@ -101,6 +101,7 @@ class ClickHouseDataLoader(AIImpactClickHouseLoader, DataLoader):
         FROM git_commits AS c
         LEFT JOIN git_commit_stats AS s
           ON (s.repo_id = c.repo_id) AND (s.commit_hash = c.hash)
+         AND (s.org_id = c.org_id)
         WHERE c.committer_when >= {{start:DateTime}} AND c.committer_when < {{end:DateTime}}
         {repo_filter}
         {org_filter_c}
@@ -499,6 +500,7 @@ class ClickHouseDataLoader(AIImpactClickHouseLoader, DataLoader):
         FROM coverage_snapshots AS c
         INNER JOIN ci_pipeline_runs AS p
           ON (p.repo_id = c.repo_id) AND (p.run_id = c.run_id)
+         AND (p.org_id = c.org_id)
         WHERE p.started_at >= {{start:DateTime}} AND p.started_at < {{end:DateTime}}
         {repo_filter}
         {self._org_filter(alias="p")}
--- a/src/dev_health_ops/migrations/clickhouse/042_rmt_org_id_dedup_keys.py
+++ b/src/dev_health_ops/migrations/clickhouse/042_rmt_org_id_dedup_keys.py
@@ -23,20 +23,46 @@ key using the same shadow-table pattern as migration 027 (Altinity pattern):
     1. SHOW CREATE TABLE to get full DDL (preserves settings, indexes, the
        RMT version column, partitioning, etc.)
     2. Modify DDL: rename to ``<table>_new``, prepend org_id to ORDER BY
-    3. INSERT INTO <table>_new SELECT * FROM <table>
-    4. Verify the distinct new-key tuple counts match between original and
+    3. Verify via system.tables that the shadow's sorting key is exactly
+       ``org_id, <old sorting key>`` — abort (and drop the shadow) on any
+       mismatch, so a regex miss on exotic DDL fails closed instead of
+       silently shipping the wrong dedup key
+    4. INSERT INTO <table>_new SELECT * FROM <table> (snapshot copy)
+    5. Verify the distinct new-key tuple counts match between original and
        shadow (raw row counts may legitimately differ: the copy can collapse
        not-yet-merged same-key duplicate versions, which is RMT semantics,
        not data loss)
-    5. EXCHANGE TABLES <table> AND <table>_new (atomic swap)
-    6. DROP TABLE <table>_new (which now holds the old structure)
+    6. EXCHANGE TABLES <table> AND <table>_new (atomic swap)
+    7. CATCH-UP: INSERT INTO <table> SELECT * FROM <table>_new — the shadow
+       now holds the OLD table, including any rows written between the
+       snapshot (step 4) and the swap (step 6). Re-inserting ALL old rows is
+       idempotent under RMT semantics: rows already copied dedup away on the
+       (now org_id-first) key with the newest version winning, while rows
+       that landed after the snapshot survive. This shrinks the write-loss
+       window from the whole copy duration to the EXCHANGE instant, which is
+       atomic.
+    8. DROP TABLE <table>_new — only after the catch-up succeeded.
+
+Residual operational risk: writes that race the EXCHANGE itself can target
+either side of the swap; both sides are preserved (catch-up re-inserts the
+old side), but version ties on the same key resolve arbitrarily. Ops note:
+run this migration during a quiet period / with sync workers stopped.
+
+Crash convergence: if a run crashes after EXCHANGE but before the catch-up
+or DROP, the main table already has the org_id-first key, so a rerun takes
+the idempotency-skip path. That path checks for a leftover ``<table>_new``
+shadow and, if present, performs the catch-up INSERT + DROP before
+declaring the table done — no rows stranded, no shadow left behind. A crash
+*before* EXCHANGE leaves a disposable shadow that the next run drops and
+recreates.
 
 ``work_items`` and ``work_item_transitions`` are included defensively: on
 any database where migration 027 already ran they are skipped by the
 idempotency check (org_id already first in ORDER BY), but a database whose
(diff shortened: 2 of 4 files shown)


chrisgeo:
Codex-review findings addressed in 8b63ad882:

**1. Racing writes (HIGH)** — after `EXCHANGE TABLES`, the migration now runs a catch-up `INSERT INTO <table> SELECT * FROM <table>_new` before dropping the old table. Re-inserting all old rows is idempotent under RMT dedup (newest version wins), so rows written between the snapshot copy and the swap survive. Loss window shrinks to the atomic EXCHANGE instant; residual risk + quiet-period ops note documented in the migration docstring and PR body.

**2. Cross-org join leak (HIGH)** — added `org_id` equality to the `coverage_snapshots ↔ ci_pipeline_runs` and `git_commits ↔ git_commit_stats` joins in `metrics/loaders/clickhouse.py`. The `ci_job_runs` and `test_case_results` joins are fixed in PR #857 (CHAOS-2293), which also reworks the alias filter handling in those hunks — left untouched here to avoid a conflict; merge order between #857 and #862 doesn't matter.

**3. Crash convergence (MEDIUM)** — the idempotency-skip path now detects a leftover `<table>_new` (crash after EXCHANGE, before catch-up/DROP — at that point it holds the OLD table), runs the catch-up from it, then drops it. The previous blanket drop-shadow-on-any-failure was removed: post-EXCHANGE it would have destroyed the catch-up delta. Pre-EXCHANGE failures still clean up their disposable shadow.

**4. Fail-closed ORDER BY rewrite (MEDIUM)** — after creating the shadow, the migration verifies via `system.tables.sorting_key` that the shadow's key is exactly `org_id, <old key>` (normalized) before copying any data; mismatch ⇒ drop shadow + abort. Regex misses on nested expressions (`cityHash64(...)` etc.) can no longer ship a wrong dedup key.

**Tests** — live dedup tests now cover all rebuilt TestOps tables (7 live tests, all passing against local CH); new FakeClient unit tests cover catch-up ordering, shadow retention on post-swap failure, convergence, key-verify abort, and pre-swap cleanup (11 unit tests). Full flow also exercised end-to-end on live ClickHouse via a scratch table (cleaned up).

**Gates** — mypy clean (1014 files); `./ci/run_tests.sh unit` 3963 passed with only the 4 sanctioned failures (test_org_deletion ×2, TestCoreCache ×2).

PR #867, opened 2026-06-11
feat(deploy): explicit migrate step in all production stacks (CHAOS-2304)

chrisgeo:
## Summary

CHAOS-2268/#850 gated ClickHouse auto-migration behind `AUTO_RUN_MIGRATIONS` (default `true`), but no production deploy stack shipped a migrate step — api/workers all still ambient-migrated on boot. Migration 042 (#862) does shadow-table rebuilds, so concurrent worker-driven migration in prod is race-prone. This PR ends worker-driven ambient migration in every production stack by adding an explicit one-shot migrate step (reference pattern: workspace `compose.yml`) and setting `AUTO_RUN_MIGRATIONS=false` on app services.

### docker-compose (`deploy/docker-compose/compose.production.yml`)
- One-shot `migrate` service (same image): `dev-hops migrate postgres && dev-hops migrate clickhouse`; Alembic step auto-skips when `POSTGRES_URI` is unset (this stack's Postgres is external/optional).
- `api`/`billing-edge`/`worker` gate via `depends_on.condition: service_completed_successfully`; `AUTO_RUN_MIGRATIONS: "false"` on api/worker.
- `POSTGRES_URI` documented as direct-to-5432, never a transaction-mode pooler (DDL — see `docs/ops/database-connection-pooling.md`).

### kubernetes (`deploy/kubernetes/`)
- New `migrate-job.yaml` Job (`restartPolicy: Never`, `backoffLimit: 2`, same image + `envFrom` configmap/secret as api), added to `kustomization.yaml`.
- **Job, not initContainer**: api/worker run multiple replicas under HPAs — per-pod initContainers would run the same DDL concurrently, recreating the race. Jobs don't gate Deployments natively, so the required `kubectl wait --for=condition=complete` ordering is documented in the manifest comment and a new `deploy/kubernetes/README.md`.
- `AUTO_RUN_MIGRATIONS: "false"` added to the shared configmap (covers api + worker via `envFrom`).

### helm (`deploy/helm/dev-health`)
- `templates/migrate-job.yaml`: `pre-install,pre-upgrade` hook Job, `hook-delete-policy: before-hook-creation,hook-succeeded`, toggle `migrations.hook.enabled` (default `true`).
- Hook ships its own ConfigMap/Secret copies (hook-weight -5): pre-install hooks run **before** the chart's regular resources exist, so the Job can't reference them on a fresh install. With `secrets.create=false` the external secret is referenced directly. Config/secret data factored into `dev-health.configData`/`dev-health.secretData` helpers — rendered output verified byte-identical to main except the intended new key.
- Shared configmap computes `AUTO_RUN_MIGRATIONS` from the toggle (`false` when hook enabled, `true` when disabled), overridable via `config.AUTO_RUN_MIGRATIONS`.

### docker-swarm (`deploy/docker-swarm/stack.yml`)
- `migrate` service with `deploy.restart_policy.condition: none` (one-shot), hardened (`no-new-privileges`, `read_only` + tmpfs); `AUTO_RUN_MIGRATIONS: "false"` on api/worker.
- Swarm has no `depends_on` ordering: README documents that operators must verify `docker service logs <stack>_migrate` before first deploy / after image updates, and how to re-run via `docker service update --force`.

## Validation
- `docker compose -f deploy/docker-compose/compose.production.yml config -q` (dummy env) — OK; verified the interpolated `migrate` service entrypoint.
- `docker compose -f deploy/docker-swarm/stack.yml config -q` — OK (pre-existing obsolete-`version` warning only).
- `kubectl kustomize deploy/kubernetes/` — OK.
- `helm template` (default + `migrations.hook.enabled=false` + values-quickstart) and `helm lint` — OK; configmap/secret render-diffed against main.

## Tests
- `tests/test_compose_config.py`: new assertions for all four stacks (migrate step exists, compose gating wired, `AUTO_RUN_MIGRATIONS=false` present, helm hook annotations, kustomization wiring) — 9 passed.
- `mypy --install-types --non-interactive .` — clean (1015 files).
- `./ci/run_tests.sh unit` — 4039 passed; only the 4 known-ignorable failures (test_org_deletion ×2, TestCoreCache ×2).

### DIFF
--- a/deploy/docker-compose/compose.production.yml
+++ b/deploy/docker-compose/compose.production.yml
@@ -48,6 +48,40 @@ services:
     networks:
       - dev-health
 
+  # ===========================================================================
+  # Database migrations (one-shot — runs before app services)
+  # ===========================================================================
+  # Applies PostgreSQL (Alembic, when POSTGRES_URI is set) and ClickHouse
+  # schema migrations, then exits 0. App services gate on this completing
+  # successfully via depends_on.condition: service_completed_successfully and
+  # run with AUTO_RUN_MIGRATIONS=false so they never ambient-migrate
+  # (CHAOS-2304 — migration 042+ does shadow-table rebuilds that are not safe
+  # to run concurrently from workers).
+  #
+  # POSTGRES_URI must point DIRECTLY at Postgres (port 5432), never at a
+  # transaction-mode pooler (PgBouncer/RDS Proxy): migrations run raw DDL.
+  # See docs/ops/database-connection-pooling.md.
+  migrate:
+    image: ${DEV_HEALTH_IMAGE:-ghcr.io/your-org/dev-health-ops:latest}
+    container_name: dev-health-migrate
+    restart: "no"
+    entrypoint:
+      - sh
+      - -c
+      - >-
+        if [ -n "$$POSTGRES_URI" ]; then dev-hops migrate postgres; fi
+        && dev-hops migrate clickhouse
+    environment:
+      # Direct Postgres URI (5432), NOT pgbouncer — empty skips the Alembic step.
+      POSTGRES_URI: ${POSTGRES_URI:-}
+      CLICKHOUSE_URI: clickhouse://${CLICKHOUSE_USER:-ch}:${CLICKHOUSE_PASSWORD:-ch}@clickhouse:8123/${CLICKHOUSE_DB:-default}
+      LOG_LEVEL: ${LOG_LEVEL:-INFO}
+    depends_on:
+      clickhouse:
+        condition: service_healthy
+    networks:
+      - dev-health
+
   api:
     image: ${DEV_HEALTH_IMAGE:-ghcr.io/your-org/dev-health-ops:latest}
     container_name: dev-health-api
@@ -70,6 +104,9 @@ services:
       JIRA_BASE_URL: ${JIRA_BASE_URL:-}
       JIRA_EMAIL: ${JIRA_EMAIL:-}
       JIRA_API_TOKEN: ${JIRA_API_TOKEN:-}
+      # Schema is applied by the one-shot `migrate` service (gated via
+      # depends_on) — the API must never run migrations itself (CHAOS-2304).
+      AUTO_RUN_MIGRATIONS: "false"
     ports:
       - "${API_PORT:-8000}:8000"
     depends_on:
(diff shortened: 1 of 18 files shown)


chrisgeo:
## Codex review follow-ups addressed (c82e2fe)

**Finding 1 (HIGH) — migrate Job had no resolvable ClickHouse DSN:** `dev-hops migrate clickhouse` resolves `CLICKHOUSE_URI` / `--analytics-db` (`resolve_sink_uri`), not `DATABASE_URI`. Added `CLICKHOUSE_URI` to `dev-health-secrets` (same DSN); the Job already `envFrom`s that secret, so it now boots correctly first time. `DATABASE_URI` untouched for app compat.

**Finding 2 (HIGH) — naive `kubectl apply -k` could serve on an unmigrated schema:**
- Added a read-only wait primitive to the CLI: `dev-hops migrate clickhouse status --check` exits 1 while any migration is pending, 0 when current. No DDL, ever — `status` was already read-only but always exited 0, so a new flag was needed.
- api and worker Deployments now carry a `wait-for-migrations` initContainer that polls `status --check` every 5s (60 attempts per run, then exit 1, relying on kubelet restart backoff as the overall timeout). Concurrent replicas polling cannot race: the one-shot Job remains the only migration runner.
- The README's explicit `kubectl wait` flow stays the recommended path; the initContainer is documented as the safety net. The check is **ClickHouse-only** — Postgres is external/optional in this stack and Alembic has no equally cheap read-only pending-check wired into `dev-hops`; README says so explicitly.
- Bonus hardening (scanner-driven): api/worker pods now set `runAsNonRoot` + `allowPrivilegeEscalation: false`, matching migrate-job.yaml.

**Finding 3 (MEDIUM) — tests:**
- `test_compose_config.py`: secret defines `CLICKHOUSE_URI` and it round-trips through `resolve_sink_uri`; migrate Job envFrom sees the secret; api/worker initContainer contract (present, runs `status --check`, every `dev-hops` invocation is the read-only probe, envFrom includes the secret).
- `test_migrate_commands.py`: `--check` routing + exit codes (pending → 1, current → 0, fresh DB with no `schema_migrations` table → 1, no migration files → 0, plain `status` keeps exit 0; client closed in all paths).

**Nit:** operator recovery note added next to the `migrate` service in `compose.production.yml` (failed run + `restart: "no"` leaves api/billing-edge/worker gated — logs, re-run, then `up -d`), mirroring the swarm/k8s READMEs.

**Gates:** mypy clean (1015 files); targeted pytest 62 passed; `kubectl kustomize deploy/kubernetes/` builds; full `ci/run_tests.sh unit`: 4050 passed, only the 4 known pre-existing failures (test_org_deletion ×2, TestCoreCache ×2).

PR #928, opened 2026-06-15
feat(workers): daily scheduled investment materialization (CHAOS-2439)

chrisgeo:
## Summary

Rollout-robustness follow-on to the merged **CHAOS-2427** (#925, work-graph theme/subcategory filter).

Investment materialization — which populates `work_unit_membership`, the table the work-graph theme/subcategory filter reads — is currently **event-driven only**: it runs post-sync via the `run_work_graph_build` → `run_investment_materialize` chain in `_dispatch_post_sync_tasks` (`workers/sync_runtime.py`). There is no fixed-cadence beat entry. So **idle-sync orgs** and the **post-deploy window** leave `work_unit_membership` empty, stranding theme filters in the `MEMBERSHIP_NOT_MATERIALIZED` degraded state indefinitely.

This adds a **daily floor-cadence safety net** that re-runs the same chain per org.

## Changes

**1. Beat entry** (`workers/config.py`) — `run-investment-materialize-daily`, daily at **01:15 UTC** (clear of `run-daily-metrics` 01:00 and `run-release-impact-daily` 01:30):
```python
"run-investment-materialize-daily": {
    "task": "dev_health_ops.workers.tasks.dispatch_investment_materialize",
    "schedule": crontab(hour=1, minute=15),
    "options": {"queue": "default"},
},
```

**2. Dispatch task** (`workers/work_graph_tasks.py::dispatch_investment_materialize`) — fans out per active org the **SAME immutable `build → materialize` chain** used post-sync:
```python
build_sig = celery_app.signature("...run_work_graph_build", kwargs={"org_id": org_id}, queue="metrics")
materialize_sig = celery_app.signature("...run_investment_materialize", kwargs={"org_id": org_id}, queue="metrics", immutable=True)
chain(build_sig, materialize_sig).apply_async()
```
Chain, **not** parallel — materializing against a stale/empty work graph is exactly the race **CHAOS-2374** avoided; the immutable link keeps the build's return value out of materialize's positional args. Idempotent and safe to coexist with the post-sync dispatch (materialization is keyed on `computed_at` in a ReplacingMergeTree — a re-run just refreshes the latest run).

**3. Org selection** — mirrors the other daily dispatchers via `_discover_active_org_ids()` (the existing helper in `recommendations_tasks.py`: active orgs from Postgres as source of truth, `["default"]` fallback for single-tenant installs). Then **narrows to orgs that actually have `work_graph_edges`** in ClickHouse so empty / never-synced orgs are not churned nightly (fail-open if the probe errors — the per-org chain is itself a no-op for an empty org, so a transient probe failure must not skip the safety net). Retries on org-enumeration failure rather than reporting a silent empty success (mirrors `dispatch_release_impact`). No schema change, no migration.

**4. Log fix** (`work_graph/runner.py`) — the materialization-complete log printed `Components/Records/Quotes` only, omitting the `memberships` count `materialize_investments()` already returns, so a real run that wrote membership looked like it hadn't. Now logs `... Quotes=%d Memberships=%d`.

## Org-selection source mirrored

`_discover_active_org_ids()` (`workers/recommendations_tasks.py`) — the same active-org enumeration used by the recommendations daily safety-net job and documented as mirroring `dispatch_scheduled_metrics`. Narrowed to ClickHouse `work_graph_edges` presence for the "has data" gate (mirrors how `_discover_team_ids` scopes to tables that hold data).

## Tests

`tests/test_scheduled_investment_dispatch.py` (follows `test_post_sync_investment_dispatch.py` style — patches the Celery `chain`/`signature` factories and the org/data sources, asserts the dispatch contract without a live broker/ClickHouse):
- Beat entry registered with expected name + schedule (01:15) and non-collision with the 01:00/01:30 neighbours.
- Dispatch queues a `build → materialize` **chain** per org in chained order (asserts chain composition, immutable link, metrics-queue org-scoping — **not** parallel).
- Skips orgs without `work_graph_edges`; clean no-op when none have data.
- Chain shape **matches the post-sync chain contract** byte-for-byte (idempotent coexistence).
- Retries on org-enumeration failure (no silent empty success).
- Org-with-data probe fails open on a ClickHouse error.

**Full `mypy --install-types --non-interactive .` clean over 1073 files; targeted suite 68 passed.**

Refs CHAOS-2439. Parent: CHAOS-2427.

---

## Review-round fixes (Codex #928) — commit 25bdf90e8

**1. [HIGH] Stopped gating the rebuild on the build's own OUTPUT table.**
The dispatcher previously narrowed the fan-out to orgs that already have `work_graph_edges` rows — but that table is the *output* the build creates. Any active org with raw synced data but no edges yet (a failed prior post-sync build, a truncated/migrated-empty graph table, or a brand-new tenant that synced but never built) would be **skipped forever** while the task still reported success — skipping exactly the tenants the safety net exists to repair.

**New gating: none — dispatch the `build → materialize` chain for EVERY active org.** `_orgs_with_work_graph_data` is removed entirely. This is the preferred option from the review: the build is a cheap no-op for an org with no source data (it scans a 30-day window and yields zero edges) and `materialize_investments` then short-circuits on zero components (`{"components": 0, ...}`), so fanning out to all active orgs is correct and inexpensive. The rule is simply: never gate a rebuild on its own output.

**2. [MEDIUM] Strict org discovery so a DB outage retries instead of silently succeeding.**
`_discover_active_org_ids()` caught its own Postgres exceptions and returned `["default"]`, so the retry-wrapped dispatcher treated a multi-tenant DB outage as a clean empty success and never retried the once-daily run. Added a `strict` flag:
```python
def _discover_active_org_ids(*, strict: bool = False) -> list[str]:
    try:
        ... query active orgs ...
    except Exception:
        logger.exception("Failed to enumerate active organizations")
        if strict:
            raise          # CHAOS-2439: retryable DB outage, not "no orgs"
        org_ids = []
    return org_ids or ["default"]   # ["default"] only on a SUCCESSFUL empty query
```
The dispatcher calls `_discover_active_org_ids(strict=True)` so a Postgres failure raises and triggers its existing `self.retry(...)`. The `["default"]` fallback now applies only to the *positively-detected* single-tenant case (successful query, zero active orgs). The existing recommendations-job caller keeps the non-strict default — behaviour unchanged.

**Tests** (updated `tests/test_scheduled_investment_dispatch.py`): an active org with raw inputs but **zero `work_graph_edges` IS dispatched** (finding-1 regression) and the module exposes no output-table probe helper; single-tenant `["default"]` is dispatched; the dispatcher calls discovery with `strict=True`; strict discovery **raises** on a DB error → retry fires; strict discovery still returns `["default"]` on a successful empty table; non-strict still falls back on error. **Full `mypy --install-types --non-interactive .` clean over 1073 files; targeted suite 70 passed.**

---

## Review round-2 fix (Codex #928) — commit cac5e09e9

**[HIGH] Forward the resolved `db_url` to both child signatures.**
`dispatch_investment_materialize` resolved `db_url` but enqueued `run_work_graph_build` and `run_investment_materialize` with only `org_id`. Both children fall back to `_get_db_url()` when `db_url` is absent, so an explicit override — `dispatch_investment_materialize(db_url="clickhouse://override")` for a manual/backfill run — silently ran the build+materialize against the workers' **ambient** ClickHouse while still returning a clean `dispatched` result. The sibling dispatchers forward `db_url`; this one dropped it.

```python
build_sig = celery_app.signature(
    "...run_work_graph_build",
    kwargs={"db_url": db_url, "org_id": org_id}, queue="metrics",
)
materialize_sig = celery_app.signature(
    "...run_investment_materialize",
    kwargs={"db_url": db_url, "org_id": org_id}, queue="metrics", immutable=True,
)
```
The scheduled (no-override) path forwards the same value `_get_db_url()` already resolves, so behaviour is unchanged when no override is supplied; the immutable link on the materialize sig is preserved.

**Tests:** `dispatch_investment_materialize.run(db_url="clickhouse://override")` → both child signatures receive `db_url="clickhouse://override"` and `org_id` (and never collapse to the ambient default); the scheduled path forwards the resolved default to both children; the chain-shape test updated for the new kwargs. **Full `mypy --install-types --non-interactive .` clean over 1073 files; targeted suite 62 passed.**

---

## Round-3 redesign (user decision) — commit 4edbc831d: daily job is now NO-LLM

**The daily scheduled job must NOT re-run LLM materialization** (cost + category drift). The daily beat now runs a CHEAP, no-LLM **membership backfill** that projects `work_unit_membership` from the theme/subcategory distributions ALREADY persisted in `work_unit_investments` by the post-sync LLM path. **The post-sync `build -> run_investment_materialize` (LLM) chain is UNCHANGED** — it still categorizes new data.

### Shared-helper extraction
The threshold + `is_dominant` projection logic moved to a new module `work_graph/investment/membership.py`:
- `lexical_argmax`, `membership_categories`, and `build_membership_records(unit_nodes, work_unit_id, theme_distribution, subcategory_distribution, categorization_status, computed_at, org_id)`.

`materialize.py` now calls `build_membership_records` (keeping `_lexical_argmax` / `_membership_categories` as thin aliases for existing imports), so the **LLM materializer and the no-LLM backfill emit byte-for-byte identical rows** for the same persisted distributions — proven by a test that diffs the backfill output against `build_membership_records`.

### New backfill (`work_graph/investment/backfill.py` + `run_membership_backfill` task), per org, NO LLM
1. Rebuild work-graph connected components from `work_graph_edges`; compute `work_unit_id` + `unit_nodes` (reuse `work_unit_id()` so hashing matches the materializer).
2. Read EXISTING `work_unit_investments` **latest-per-`work_unit_id`** (`argMax` on `computed_at`, same semantics as `api/queries/work_unit_investments.py`) for `theme_distribution_json` / `subcategory_distribution_json`. **No categorizer/LLM call.**
3. Project membership via the shared `build_membership_records`, stamping a fresh `computed_at` (the resolver's per-node latest-run guard then supersedes stale rows).
4. Write via `sink.write_work_unit_memberships`.

**Skip-on-churned-component:** a unit whose CURRENT component hash has no matching `work_unit_investments` row — because edges churned since the last categorization, so the node's new `work_unit_id` was never categorized — is **skipped**. The post-sync full `build -> run_investment_materialize` (LLM) chain covers those (it runs on new data). The backfill only refreshes membership for components that ALREADY have a categorization; it never invents categories. Documented in the module docstring.

### Beat repoint
`run-membership-backfill-daily` (01:15 UTC) → `dispatch_membership_backfill`, which fans out the no-LLM chain `run_work_graph_build -> run_membership_backfill` per active org. The prior fixes are KEPT: **strict org discovery** (`strict=True` → raise → retry on a Postgres outage) and **`db_url` threaded to BOTH child sigs**. (`dispatch_investment_materialize` is removed; the old beat entry `run-investment-materialize-daily` is gone.)

### Tests
- `tests/test_membership_backfill.py` (new): daily backfill invokes **NO categorizer/LLM provider** (asserts both are never called); backfill rows **equal `build_membership_records`** for the same distributions (shared-helper consistency); idle org with existing investments → membership populated, no LLM; **skip-on-churned-component** (matched component written, churned one skipped); latest-per-unit `argMax` query shape; empty-graph clean no-op.
- `tests/test_scheduled_membership_backfill.py` (renamed from the dispatch test): chain order `build -> backfill`, immutable link, metrics-queue org-scoping, **`db_url` forwarded to both children** (override + scheduled default), strict discovery → retry, beat name/schedule.

**Full `mypy --install-types --non-interactive .` clean over 1077 files; targeted suite 90 passed.**

---

## Round-4 fix (Codex adversarial review) — commit 69014d4aa: tombstone-on-skip

**[HIGH] Stale-on-skip: churned nodes retained obsolete theme/subcategory membership.**

When a component's `work_unit_id` has no matching `work_unit_investments` row (edges churned since last LLM run), the backfill previously wrote **nothing** for those nodes. The resolver's per-node latest-run guard selects membership rows whose `computed_at` equals the max for that `(org_id, node_type, node_id)`, so any **OLD** membership rows from a prior component remained the latest and kept matching theme/subcategory filters with stale categories. The daily job reported success while leaving stale data.

### Fix: tombstone-on-skip (no LLM)

Instead of writing nothing, the backfill now writes a **TOMBSTONE** for each node in a skipped (churned/uncategorized) component:

```python
WorkUnitMembershipRecord(
    category_kind="theme"|"subcategory",
    category="",          # TOMBSTONE_CATEGORY sentinel — never a real category
    weight=0.0,
    is_dominant=0,
    categorization_status="tombstone",
    computed_at=<current run timestamp>,   # FRESH — supersedes stale rows
    ...
)
```

Two tombstone rows per node (one per `category_kind`). The `computed_at` matches the rest of the run, so the per-node latest-run guard picks up the tombstone and displaces all older rows.

### Tombstone invariants (verified by analysis and test)

| Invariant | Mechanism |
|---|---|
| Tombstoned node not returned by theme/subcategory filter | `(m.category_kind, m.category) IN %(category_tuples)s` — all real categories are non-empty strings, `''` never matches |
| Tombstoned node annotated as `theme=null / subcategory=null` | Tombstone has `is_dominant=0` → excluded from annotation query's `AND m.is_dominant = 1`; no entry → `None` |
| Tombstoned org NOT misread as `MEMBERSHIP_NOT_MATERIALIZED` | Tombstones are live rows with fresh `computed_at`; degraded probe sees `membership_rows > 0` → not degraded |
| `materialize.py` unchanged | Only the backfill skip path tombstones. The LLM materializer re-categorizes. Post-sync chain replaces tombstones with real rows. |

### Stats dict

Added `tombstones` key to the returned stats dict (total tombstone rows written). Backward-compatible — callers pass stats through as `{"status": "success", "stats": stats}`.

### Regression test (`test_tombstone_on_skip_regression`)

Covers all three invariants end-to-end:

**(a) Backfill emission** — churned component with no investments row → tombstone rows written for all nodes with fresh `computed_at`; `category=''`, `weight=0.0`, `is_dominant=0`, `categorization_status="tombstone"`.

**(b) Resolver filter** — mock returns empty edge list (simulating ClickHouse correctly finding no edges because `category=''` ∉ `{('theme', 'quality')}`); asserts `result.edges == []` and `degraded_reason is None` (tombstones are live rows → `membership_rows > 0` → not degraded); asserts SQL uses `(m.category_kind, m.category) IN %(category_tuples)s` with non-empty category tuples.

**(c) Annotation** — annotation query excludes `is_dominant=0` rows; tombstoned node returns no annotation rows → `edge.theme is None`, `edge.subcategory is None`.

**Full `mypy --install-types --non-interactive .` clean over 1074 files; targeted suite 72 passed.**

### DIFF
--- a/src/dev_health_ops/metrics/sinks/clickhouse/core.py
+++ b/src/dev_health_ops/metrics/sinks/clickhouse/core.py
@@ -188,18 +188,38 @@ class ClickHouseCore(BaseMetricsSink):
             row[0] for row in (getattr(applied_result, "result_rows", []) or [])
         )
 
-        # Collect all migration files
+        # Collect all migration files in apply order.
         migration_files = sorted(
             list(migrations_dir.glob("*.sql")) + list(migrations_dir.glob("*.py"))
         )
 
+        # Fast-path: short-circuit only when EVERY on-disk migration is already
+        # applied (full-set completeness check via all_migrations_applied —
+        # CHAOS-2440).  A latest-filename-only check is unsound here because of
+        # inserted / mixed-ordering migrations (023b_ between 023_ and 024_,
+        # duplicate numeric prefixes): the DB could hold the latest row yet be
+        # missing an intermediate migration.  The full-set check stays O(n) in
+        # memory over the SELECT we already ran — no per-file DB query, no log
+        # loop — so it remains fast and quiet while being correct.
+        from dev_health_ops.migrations.clickhouse import all_migrations_applied
+
+        if migration_files and all_migrations_applied(
+            (p.name for p in migration_files), applied_versions
+        ):
+            logger.debug(
+                "ClickHouse schema is current (%d migrations applied) — "
+                "skipping migration loop",
+                len(migration_files),
+            )
+            return
+
         for path in migration_files:
             version = path.name
             if version in applied_versions:
-                logger.info(f"Skipping already applied migration: {version}")
+                logger.debug("Skipping already applied migration: %s", version)
                 continue
 
-            logger.info(f"Applying migration: {version}")
+            logger.info("Applying migration: %s", version)
 
             if path.suffix == ".sql":
                 try:
--- a/src/dev_health_ops/migrations/clickhouse/__init__.py
+++ b/src/dev_health_ops/migrations/clickhouse/__init__.py
@@ -10,7 +10,37 @@ into its own statement and break a migration with a ClickHouse SYNTAX_ERROR
 
 from __future__ import annotations
 
-__all__ = ["strip_line_comments", "split_sql_statements"]
+from collections.abc import Iterable
+
+__all__ = [
+    "strip_line_comments",
+    "split_sql_statements",
+    "all_migrations_applied",
+]
+
+
+def all_migrations_applied(
+    migration_filenames: Iterable[str],
+    applied_versions: Iterable[str],
+) -> bool:
+    """Return True iff every on-disk migration is already recorded as applied.
+
+    This is the single source of truth for the migration runners' fast-path
+    short-circuit (CHAOS-2440). It must be a FULL-SET completeness check, not a
+    "latest filename applied" check: this repo has inserted / mixed-ordering
+    migrations (e.g. ``023b_dora_metrics.sql`` sorts between ``023_*`` and
+    ``024_*``, and duplicate numeric prefixes like ``002_phase2_metrics.sql`` /
+    ``002_teams.sql`` exist). A database can hold the row for the
+    lexicographically-latest migration while still missing an *intermediate*
+    one — a latest-only check would falsely report "current" and silently skip
+    the missing migration, causing schema drift.
+
+    Comparing the full set of on-disk filenames against the applied set is still
+    O(n) in memory over the single ``SELECT version FROM schema_migrations`` that
+    the runner already issues — no per-file DB query, no logging loop — so the
+    fast-path stays both fast and quiet while being correct.
+    """
+    return set(migration_filenames) <= set(applied_versions)
 
 
 def strip_line_comments(sql: str) -> str:
--- a/src/dev_health_ops/storage/clickhouse.py
+++ b/src/dev_health_ops/storage/clickhouse.py
@@ -135,14 +135,36 @@ class ClickHouseStore:
                 row[0] for row in (getattr(applied_result, "result_rows", []) or [])
             )
 
-            # Collect all migration files
+            # Collect all migration files in apply order.
             migration_files = sorted(
                 list(migrations_dir.glob("*.sql")) + list(migrations_dir.glob("*.py"))
             )
 
+            # Fast-path: short-circuit only when EVERY on-disk migration is
+            # already applied (full-set completeness check via
+            # all_migrations_applied — CHAOS-2440).  A latest-filename-only
+            # check is unsound here because of inserted / mixed-ordering
+            # migrations (023b_ between 023_ and 024_, duplicate numeric
+            # prefixes): the DB could hold the latest row yet be missing an
+            # intermediate migration.  The full-set check stays O(n) in memory
+            # over the SELECT we already ran — no per-file DB query, no log
+            # loop — so it remains fast and quiet while being correct.
+            from dev_health_ops.migrations.clickhouse import all_migrations_applied
+
+            if migration_files and all_migrations_applied(
+                (p.name for p in migration_files), applied_versions
+            ):
+                logger.debug(
+                    "ClickHouse schema is current (%d migrations applied) — "
+                    "skipping migration loop",
+                    len(migration_files),
+                )
+                return
+
             for path in migration_files:
                 version = path.name
                 if version in applied_versions:
+                    logger.debug("Skipping already applied migration: %s", version)
                     continue
 
                 if path.suffix == ".sql":
--- a/src/dev_health_ops/work_graph/investment/backfill.py
+++ b/src/dev_health_ops/work_graph/investment/backfill.py
@@ -1,0 +1,232 @@
+"""No-LLM membership backfill (CHAOS-2439).
+
+The daily scheduled job must NOT re-run LLM categorization (cost + category
+drift). Instead it cheaply PROJECTS ``work_unit_membership`` from the theme /
+subcategory distributions ALREADY persisted in ``work_unit_investments`` by the
+post-sync LLM materializer.
+
+Per org, with NO categorizer/LLM call:
+
+1. Rebuild the work-graph connected components from ``work_graph_edges`` (reuse
+   ``_build_components`` + ``work_unit_id`` so the unit hashing matches the
+   materializer exactly).
+2. Read the LATEST ``work_unit_investments`` row per ``work_unit_id`` (argMax on
+   ``computed_at``, the same latest-per-unit semantics as
+   ``api/queries/work_unit_investments.py``) for those unit ids to recover the
+   persisted ``theme_distribution_json`` / ``subcategory_distribution_json``.
+3. Project membership rows via the SHARED ``build_membership_records`` helper, so
+   the rows are byte-for-byte identical to what the LLM materializer would emit
+   for the same distributions. A fresh ``computed_at`` is stamped; the resolver's
+   per-node latest-run guard then supersedes any stale rows.
+4. Write via ``sink.write_work_unit_memberships``.
+
+SKIP-ON-CHURNED-COMPONENT: a unit whose CURRENT component hash has no matching
+``work_unit_investments`` row is skipped. That happens when edges churned since
+the last categorization (the node moved to a new component, so its new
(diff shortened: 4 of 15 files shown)


chrisgeo:
Non-blocking note from re-verifying CHAOS-2433 (run-identity consistency between work_unit_investments and work_unit_membership):

`backfill.py` writes membership with a fresh `computed_at = now()`, intentionally newer than the investments it backfills. This is **safe** under the reader's per-node `max(computed_at)` guard, so no regression. But it moves further from the idealized run-equality model in CHAOS-2433 (`investments.run == membership.run`).

If you want to keep strict run-equality viable, consider stamping the **investments' `computed_at`** on backfilled membership rows instead of `now()`. Tracked as CHAOS-2443 (the run-equality reader join + partial-divergence probe in CHAOS-2433 #2/#4 were shipped as a per-node design, not as written). Live is currently clean (org a78c1a6a: 157/157 run-identity agreement).

chrisgeo:
Superseded by #933 (CHAOS-2433). The daily no-LLM membership backfill was rebuilt on the run_id/coverage protocol and absorbed into #933 (run-scoped reads, completion markers, retention, tombstones removed). Closing in favor of #933.

PR #930, opened 2026-06-15
perf(cli): quiet + fast-path migration ensure (CHAOS-2440)

chrisgeo:
## Summary

Fixes CHAOS-2440. Every CLI command that opens a ClickHouse sink was running the full migration pass and logging ~50 `Skipping already applied migration: X` lines per invocation.

- **Fast-path short-circuit**: after the single cheap `SELECT version FROM schema_migrations` that was already required, check `sorted(*.sql + *.py)[-1].name in applied_versions`. If the latest file on disk is already applied, return immediately — skip the per-file loop entirely. The check reuses the exact same `sorted(*.sql + *.py)` ordering as the apply loop, so exotic names like `023b_dora_metrics.sql` sort correctly between `023_*` and `024_*`.
- **Quiet**: the per-migration `"Skipping already applied migration: X"` log lines are lowered from `INFO` → `DEBUG` in `core.py`, and a new `DEBUG`-level skip log is added to `storage/clickhouse.py` (previously silent but still iterated). Genuinely-applying and error logs remain at `INFO`/`ERROR`.

## Files changed

- `src/dev_health_ops/metrics/sinks/clickhouse/core.py` — `_apply_sql_migrations`: fast-path + lower skip log + fix f-string logging to `%` format
- `src/dev_health_ops/storage/clickhouse.py` — `_ensure_tables`: fast-path + debug-level skip log

## Tests

9 new unit tests in `tests/test_migration_fast_path.py` covering both runners:
- Up-to-date DB → fast-path fires, no `INSERT INTO schema_migrations`, `query()` called once, no INFO-level skip logs
- DB one migration behind → pending migration is applied and recorded, in order
- Mixed `.sql`/`.py` sort order validated against real migration corpus (`023b_` naming confirmed)

## Reviewer notes

- The fast-path key is `migration_files[-1].name` where `migration_files = sorted(*.sql + *.py)`. This is **byte-lexicographic on the full filename** (Python default sort on `Path`). Since all files share the same directory, this is effectively filename sort — identical to what the apply loop iterates. No new ordering logic introduced.
- The `_run_clickhouse_upgrade` CLI path calls `ensure_schema(force=True)` which calls `_apply_sql_migrations` directly — the fast-path applies there too (correct: fresh run of `dev-hops migrate clickhouse` on an already-current DB is now silent).
- `AUTO_RUN_MIGRATIONS=false` still gates before `_apply_sql_migrations` is called at all — no change to that behavior.
- The live-e2e migration path (`dev-hops migrate clickhouse upgrade`) is unchanged in correctness: when migrations ARE pending, all pending files still apply in sorted order.

### DIFF
--- a/src/dev_health_ops/metrics/sinks/clickhouse/core.py
+++ b/src/dev_health_ops/metrics/sinks/clickhouse/core.py
@@ -188,18 +188,32 @@ class ClickHouseCore(BaseMetricsSink):
             row[0] for row in (getattr(applied_result, "result_rows", []) or [])
         )
 
-        # Collect all migration files
+        # Collect all migration files using the same sort order as the apply loop
+        # so the fast-path "latest" comparison is consistent.
         migration_files = sorted(
             list(migrations_dir.glob("*.sql")) + list(migrations_dir.glob("*.py"))
         )
 
+        # Fast-path: if the DB has already applied every file on disk, skip the
+        # per-file loop entirely.  We compare the last element of the sorted
+        # file list (latest migration on disk) against applied_versions, which
+        # is the single cheapest check possible (O(1) set lookup after the
+        # already-necessary full SELECT).  This avoids ~50 log lines and the
+        # iteration overhead on every CLI invocation when the schema is current.
+        if migration_files and migration_files[-1].name in applied_versions:
+            logger.debug(
+                "ClickHouse schema is current (latest: %s) — skipping migration loop",
+                migration_files[-1].name,
+            )
+            return
+
         for path in migration_files:
             version = path.name
             if version in applied_versions:
-                logger.info(f"Skipping already applied migration: {version}")
+                logger.debug("Skipping already applied migration: %s", version)
                 continue
 
-            logger.info(f"Applying migration: {version}")
+            logger.info("Applying migration: %s", version)
 
             if path.suffix == ".sql":
                 try:
--- a/src/dev_health_ops/metrics/sinks/clickhouse/core.py
+++ b/src/dev_health_ops/metrics/sinks/clickhouse/core.py
@@ -188,22 +188,28 @@ class ClickHouseCore(BaseMetricsSink):
             row[0] for row in (getattr(applied_result, "result_rows", []) or [])
         )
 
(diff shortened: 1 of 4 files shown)


chrisgeo:
## Update: addressed Codex adversarial review (2x HIGH — unsound fast-path)

Codex correctly flagged the original fast-path as unsound: it used `migration_files[-1].name in applied_versions` (latest-only), which falsely assumes "latest applied => all applied". This repo has inserted / mixed-ordering migrations (`023b_*` between `023_` and `024_`, duplicate numeric prefixes like `002_phase2_metrics.sql` / `002_teams.sql`), so a DB could hold the latest row while missing an **intermediate** migration — the fast-path would return "current" and silently skip the gap (schema drift).

### Corrected condition

Fast-return now only fires on a **full-set completeness check**:

```python
def all_migrations_applied(migration_filenames, applied_versions) -> bool:
    return set(migration_filenames) <= set(applied_versions)
```

Still O(n) in memory over the single existing `SELECT version FROM schema_migrations` — no per-file DB query, no logging loop — so it stays fast AND quiet, but is now correct.

### Shared helper (no triplicated logic)

Extracted into one helper `all_migrations_applied()` in `src/dev_health_ops/migrations/clickhouse/__init__.py` (alongside the existing `split_sql_statements`), imported and used by BOTH runners (`core.py` `_apply_sql_migrations` and `storage/clickhouse.py` `_ensure_tables`).

### Tests added

- 5 unit tests for `all_migrations_applied` including the missing-middle regression and "extra applied rows on disk-removed files are ignored".
- **Missing-middle integration test for BOTH runners**: latest IS applied but a middle migration is ABSENT -> fast-path does NOT fire and exactly the missing migration is applied/recorded (asserts `recorded_versions == [missing.name]`).
- Existing up-to-date / quiet / ordering / one-behind tests retained.

### Verification

- `mypy --install-types --non-interactive .` -> Success, no issues in 1073 files
- Migration + storage suite: 106 passed / 1 skipped (migration tests) + 111 passed (storage tests). New fast-path file: 16 passed.

PR #933, opened 2026-06-15
feat(work-graph): run_id/coverage protocol for membership supersedure (CHAOS-2433)

chrisgeo:
## Summary

Fixes three simultaneous failure modes in `work_unit_membership` by replacing the per-node `max(computed_at)` recency guard with a **run_id / completion-marker protocol**:

- **Concurrency race** (round-5 bug): a half-written materializer run has rows in ClickHouse but no marker → the prior complete run (e.g. from a daily backfill) remains visible until the marker lands, then readers atomically switch to the new run.
- **Split/merge stale rows**: nodes from old components carry their old `run_id`; the new complete run has a new `run_id`; old rows are simply invisible — no explicit tombstones needed.
- **Partial-write divergence**: a run with rows but no marker is never selected; there is no "partially true" membership state.

This PR supersedes #928 (branch `chaos-2439`) and **removes the tombstone mechanism entirely**. The run_id protocol makes tombstones unnecessary: a churned node is absent from the latest complete run, which the resolver treats as "no membership" (annotation null, not filterable).

### Schema / Migration
- **Migration 047**: `ADD COLUMN run_id String DEFAULT ''` to `work_unit_membership`; `CREATE TABLE work_unit_membership_runs (org_id, run_id, completed_at) ENGINE = ReplacingMergeTree(completed_at) ORDER BY (org_id, run_id)` as the completion-marker table.
- `WorkUnitMembershipRecord` gains `run_id: str = ""` (backward-compat default).
- New `WorkUnitMembershipRunRecord` dataclass for the completion marker.

### Write side
- **Materializer**: stamps `run_id` (reuses `categorization_run_id`) on every membership row; writes `work_unit_membership_runs` marker as the **last step** after all rows.
- **Backfill** (new `backfill.py`, ported from chaos-2439, rebuilt on run protocol): generates its own `uuid4().hex` run_id; writes marker last; churned components are skipped (not tombstoned).
- Sink: `write_work_unit_memberships` now includes the `run_id` column; new `write_membership_run()` writes the marker.

### Read side (3 resolver paths)
Replaced `_LATEST_RUN_PER_NODE_SUBQUERY` with `_LATEST_COMPLETE_RUN_SUBQUERY`:
```sql
SELECT argMax(run_id, completed_at) AS latest_run_id
FROM work_unit_membership_runs
WHERE org_id = %(org_id)s
```
Applied with `INNER JOIN latest_run ON m.run_id = latest_run.latest_run_id AND latest_run.latest_run_id != ''` in:
1. Filter semijoin (theme/subcategory EXISTS clause)
2. Dominant-annotation lookup (`_batch_resolve_membership`)
3. Degraded-state probe (`_detect_membership_degraded_reason`)

`_MEMBERSHIP_TABLES` now covers both `work_unit_membership` and `work_unit_membership_runs` for graceful-degrade on missing-table errors.

### Workers / scheduling
- New `run_membership_backfill` Celery task wrapping `backfill_memberships`.
- New `dispatch_membership_backfill` task fans out `build → backfill` chain per active org with `strict=True` org discovery (`_discover_active_org_ids` gained a `strict=` parameter; strict mode re-raises on Postgres failure rather than silently returning `["default"]`).
- Beat schedule: `run-membership-backfill-daily` at 03:30 UTC.

## Tests

- **`test_work_unit_membership.py`**: `test_membership_rows_carry_run_id` + `test_completion_marker_written_after_membership_rows` (write-order invariant).
- **`test_membership_backfill.py`**: no-LLM proof, shared-helper consistency, run_id stamping, marker-last write order, no marker on empty run, churned components skipped (no tombstones), dispatcher chain, strict discovery, beat schedule entry.
- **`test_work_graph_theme_filter.py`**: all 3 reader paths assert `work_unit_membership_runs` / `argMax(run_id, completed_at)` / `m.run_id = latest_run.latest_run_id` patterns; new `TestRunIdConcurrencyRace` class (4 tests):
  - Incomplete materializer run invisible; prior complete backfill run visible.
  - After materializer marker written → resolver switches to new run.
  - No complete run → degraded when investments exist.
  - Split/merge stale node absent from new run.
- **`test_work_graph_membership_split_merge_live.py`**: rewritten to use run_id protocol; added live ClickHouse `test_concurrency_race_incomplete_run_invisible`.

All 217 work_graph/membership tests pass; live ClickHouse tests pass; mypy 0 errors.

## Supersedes / removes

- Supersedes PR #928 (chaos-2439) — absorbs the no-LLM backfill, rebuilt on run protocol.
- Removes tombstone mechanism: no `TOMBSTONE_CATEGORY` sentinel, no `_build_tombstone_records`, no `tombstones` stat.

Closes CHAOS-2433.

### DIFF
--- a/src/dev_health_ops/api/graphql/resolvers/work_graph.py
+++ b/src/dev_health_ops/api/graphql/resolvers/work_graph.py
@@ -310,34 +310,39 @@ def _row_to_edge(
 
 
 # work_unit_membership is multi-membership (one row per node-category) on a
-# ReplacingMergeTree keyed by (org, node, category_kind, category).
+# ReplacingMergeTree keyed by (org, node, category_kind, category, run_id).
 #
-# STALENESS IS SCOPED PER NODE, NOT PER WORK UNIT. work_unit_id is a hash of the
-# connected component, so when edge churn moves a node into a new component the
-# OLD work_unit_id is never re-emitted — a per-work_unit_id "latest run" guard
-# would keep that dead unit's rows alive forever (the node would keep matching
-# obsolete categories, and annotation would see multiple is_dominant rows). A
-# node belongs to exactly one component per run and emits its rows with one
-# computed_at, so restricting to rows whose computed_at equals the max
-# computed_at for that (org_id, node_type, node_id) makes the node's most recent
-# materialization always supersede its prior component's rows. This fixes
-# split/merge and below-threshold drop-off in one guard.
+# RUN_ID / COMPLETION-MARKER PROTOCOL (CHAOS-2433):
+# Every membership write (materializer or backfill) stamps a single run_id on
+# ALL rows of the run, then writes ONE completion-marker row to
+# work_unit_membership_runs (org_id, run_id, completed_at) as the LAST step.
+# A run is COMPLETE only when its marker exists. Readers select the latest
+# complete run for the org:
 #
-# ORPHANS: a node removed from the graph entirely is never re-emitted, so its
-# last-known rows linger. The materialization job can run repo/team-scoped (see
-# workers/work_graph_tasks.py) — it is NOT guaranteed to be a global org sweep
-# each run — so a "latest global run per org" backstop would wrongly hide nodes
-# whose repo was not in the most recent scoped run, and tombstoning vanished
-# nodes would require a full-graph diff a scoped run cannot produce. We
-# therefore intentionally let an orphaned node retain its last categories until
-# it is re-materialized (or its edges age out of work_graph_edges, at which
-# point no edge references it anyway). This matches how the rest of the work
-# graph treats vanished nodes (no tombstones).
-_LATEST_RUN_PER_NODE_SUBQUERY = """
-        SELECT node_type, node_id, max(computed_at) AS max_computed_at
-        FROM work_unit_membership
+#   argMax(run_id, completed_at) FROM work_unit_membership_runs WHERE org_id=?
+#
+# and scope ALL membership reads to rows whose run_id equals that value.
+#
+# This protocol fixes three failure modes at once (superseding per-node
+# max(computed_at)):
+#   1. CONCURRENCY RACE: a materializer in-flight has written some membership
+#      rows but not its marker; the prior COMPLETE backfill run is still visible.
+#      The materializer's marker write atomically switches readers to the new run.
+#   2. SPLIT/MERGE STALE: nodes from a prior component that are absent from the
+#      latest complete run simply have no rows in that run — not filterable and
+#      annotation returns None, exactly "no membership row" semantics.
+#   3. PARTIAL-WRITE DIVERGENCE: a run with rows but no marker is never selected.
+#
+# NO TOMBSTONES: tombstones (category='') are not needed because a churned node
+# that is absent from the latest complete run has no rows in scope — the reader
+# sees exactly the same result as "node has no membership" without any sentinel.
+#
+# When NO complete run exists for the org (table empty or all runs incomplete),
+# readers treat it as no membership at all — degraded state if investments exist.
+_LATEST_COMPLETE_RUN_SUBQUERY = """
+        SELECT argMax(run_id, completed_at) AS latest_run_id
+        FROM work_unit_membership_runs
         WHERE org_id = %(org_id)s
-        GROUP BY node_type, node_id
 """
 
 
@@ -349,9 +354,10 @@ async def _batch_resolve_membership(
     """Batch-lookup the dominant theme/subcategory per edge endpoint in ONE query.
 
     Returns {(node_type, node_id) -> {"dominant_theme": ..., "dominant_subcategory": ...}}
-    built from the is_dominant=1 rows of each node's latest run (stale rows from
-    a prior component excluded via the per-node latest-run join — see
-    _LATEST_RUN_PER_NODE_SUBQUERY). Scoped to org_id.
+    built from the is_dominant=1 rows of the latest COMPLETE run (scoped via
+    work_unit_membership_runs — see _LATEST_COMPLETE_RUN_SUBQUERY). Nodes absent
+    from the latest complete run return no entry (annotation null). Scoped to
+    org_id.  When no complete run exists, returns {} (no annotation).
     """
     from dev_health_ops.api.queries.client import query_dicts
 
@@ -377,21 +383,20 @@ async def _batch_resolve_membership(
         membership_rows = await query_dicts(
             client,
             f"""
-            WITH latest AS ({_LATEST_RUN_PER_NODE_SUBQUERY})
+            WITH latest_run AS ({_LATEST_COMPLETE_RUN_SUBQUERY})
             SELECT
                 m.node_type AS node_type,
                 m.node_id AS node_id,
                 m.category_kind AS category_kind,
                 m.category AS category
             FROM work_unit_membership AS m
-            INNER JOIN latest
-                ON m.node_type = latest.node_type
-               AND m.node_id = latest.node_id
-               AND m.computed_at = latest.max_computed_at
+            INNER JOIN latest_run
+                ON m.run_id = latest_run.latest_run_id
             WHERE m.org_id = %(org_id)s
               AND m.node_type IN %(node_types)s
               AND m.node_id IN %(node_ids)s
               AND m.is_dominant = 1
+              AND latest_run.latest_run_id != ''
             """,
             {
                 "org_id": org_id,
@@ -442,21 +447,20 @@ def _theme_membership_exists_clause() -> str:
     A correlated semi-join pushed INTO the edge query so the membership filter,
     the repo_id/edge_type/source filters, AND the LIMIT all execute in one
     ClickHouse plan (no unbounded Python IN set; the before-LIMIT guarantee
-    holds because this lives in the edge WHERE). Per-node latest-run guard
-    excludes stale rows from a prior component (split/merge safe). The
-    uniqExact HAVING enforces "member of EVERY requested (kind, category)", so a
-    theme+subcategory filter requires membership in BOTH. An edge matches when
-    EITHER endpoint satisfies the subquery.
+    holds because this lives in the edge WHERE). Run-scoping via
+    work_unit_membership_runs (CHAOS-2433) ensures only the latest COMPLETE run's
+    rows are considered. The uniqExact HAVING enforces "member of EVERY requested
+    (kind, category)", so a theme+subcategory filter requires membership in BOTH.
+    An edge matches when EITHER endpoint satisfies the subquery.
     """
     return f"""
         EXISTS (
             SELECT 1
             FROM work_unit_membership AS m
-            INNER JOIN ({_LATEST_RUN_PER_NODE_SUBQUERY}) AS latest
-                ON m.node_type = latest.node_type
-               AND m.node_id = latest.node_id
-               AND m.computed_at = latest.max_computed_at
+            INNER JOIN ({_LATEST_COMPLETE_RUN_SUBQUERY}) AS latest_run
+                ON m.run_id = latest_run.latest_run_id
             WHERE m.org_id = %(org_id)s
+              AND latest_run.latest_run_id != ''
               AND (
                 (m.node_type, m.node_id) = (work_graph_edges.source_type, work_graph_edges.source_id)
                 OR (m.node_type, m.node_id) = (work_graph_edges.target_type, work_graph_edges.target_id)
@@ -533,14 +537,14 @@ MEMBERSHIP_NOT_MATERIALIZED = "MEMBERSHIP_NOT_MATERIALIZED"
 
 async def _detect_membership_degraded_reason(client: Any, org_id: str) -> str | None:
     """Return MEMBERSHIP_NOT_MATERIALIZED when a theme filter yielded nothing
-    because work_unit_membership is unpopulated for an org that HAS categorized
-    work units, else None.
+    because work_unit_membership has no complete run for an org that HAS
+    categorized work units, else None.
 
-    Degraded iff: the org has >= 1 work_unit_investments row AND ZERO
-    work_unit_membership rows. The membership count is scoped to each node's
-    latest run (a row is "live" only if its computed_at equals the node's max)
-    so leftover rows from a prior component never mask a truly empty table —
-    consistent with the per-node staleness used everywhere else (CHAOS-2430).
+    Degraded iff: the org has >= 1 work_unit_investments row AND no complete
+    membership run exists in work_unit_membership_runs (i.e. no marker has been
+    written yet, meaning the run_id / completion-marker protocol has not produced
+    a complete run — CHAOS-2433). Counts membership rows in the latest complete
+    run only (zero rows in the latest complete run = no usable membership).
     Also emits the rollout warning log (observability retained). Never raises:
     on probe failure it returns None so the request is unaffected.
     """
@@ -549,21 +553,15 @@ async def _detect_membership_degraded_reason(client: Any, org_id: str) -> str | 
     try:
         rows = await query_dicts(
             client,
-            """
+            f"""
             SELECT
                 (
                     SELECT count()
                     FROM work_unit_membership AS m
-                    INNER JOIN (
-                        SELECT node_type, node_id, max(computed_at) AS max_computed_at
-                        FROM work_unit_membership
-                        WHERE org_id = %(org_id)s
-                        GROUP BY node_type, node_id
-                    ) AS latest
-                        ON m.node_type = latest.node_type
-                       AND m.node_id = latest.node_id
-                       AND m.computed_at = latest.max_computed_at
+                    INNER JOIN ({_LATEST_COMPLETE_RUN_SUBQUERY}) AS latest_run
+                        ON m.run_id = latest_run.latest_run_id
                     WHERE m.org_id = %(org_id)s
+                      AND latest_run.latest_run_id != ''
                 ) AS membership_rows,
                 (
                     SELECT count()
@@ -584,10 +582,10 @@ async def _detect_membership_degraded_reason(client: Any, org_id: str) -> str | 
     investment_rows = int(rows[0].get("investment_rows") or 0)
     if membership_rows == 0 and investment_rows > 0:
         logger.warning(
-            "Theme filter returned no edges and work_unit_membership is empty "
-            "for org %s while work_unit_investments has %d rows — the "
-            "post-migration investment materialization rerun that populates "
-            "work_unit_membership has likely not run yet (CHAOS-2430).",
+            "Theme filter returned no edges and no complete membership run "
+            "exists for org %s while work_unit_investments has %d rows — "
+            "the post-migration investment materialization rerun that populates "
+            "work_unit_membership has likely not run yet (CHAOS-2430/2433).",
             org_id,
             investment_rows,
         )
@@ -595,7 +593,7 @@ async def _detect_membership_degraded_reason(client: Any, org_id: str) -> str | 
     return None
 
 
-_MEMBERSHIP_TABLE = "work_unit_membership"
+_MEMBERSHIP_TABLES = frozenset({"work_unit_membership", "work_unit_membership_runs"})
 
 # ClickHouse names the actually-missing table in an "Unknown table ...
 # identifier '<name>'" clause. The full error ALSO echoes the entire failing
@@ -623,19 +621,18 @@ def _unknown_table_names(text: str) -> set[str]:
 
 def _is_missing_membership_table_error(exc: BaseException) -> bool:
     """True ONLY when a ClickHouse missing-table (code 60) error names
-    ``work_unit_membership`` as the unknown table.
+    ``work_unit_membership`` OR ``work_unit_membership_runs`` as the unknown table.
 
-    During a rolling deploy or before the membership migration has run, the
-    filtered edge query's EXISTS subquery references ``work_unit_membership``
-    before it exists; the driver raises a DatabaseError carrying ``code == 60``
-    with an ``UNKNOWN_TABLE`` server message whose identifier clause names the
-    offending table. We require the code-60/UNKNOWN_TABLE signal AND that the
-    reported-unknown identifier IS ``work_unit_membership`` — parsed from the
-    identifier clause, NOT the echoed SQL (which lists every table in the
-    query). So a code-60 error for any OTHER table (e.g. a missing
-    ``work_graph_edges`` or another schema regression on the filtered path)
-    re-raises and surfaces loudly instead of masquerading as the benign degraded
-    state.
+    During a rolling deploy or before migration 047 has run, the filtered edge
+    query or the degraded probe references these tables before they exist; the
+    driver raises a DatabaseError carrying ``code == 60`` with an
+    ``UNKNOWN_TABLE`` server message whose identifier clause names the offending
+    table. We require the code-60/UNKNOWN_TABLE signal AND that the
+    reported-unknown identifier IS one of the membership tables — parsed from the
+    identifier clause, NOT the echoed SQL (which lists every table in the query).
+    So a code-60 error for any OTHER table (e.g. a missing ``work_graph_edges``
+    or another schema regression on the filtered path) re-raises and surfaces
+    loudly instead of masquerading as the benign degraded state.
     """
     text = str(exc)
     is_unknown_table = (
@@ -645,7 +642,7 @@ def _is_missing_membership_table_error(exc: BaseException) -> bool:
     )
     if not is_unknown_table:
         return False
-    return _MEMBERSHIP_TABLE in _unknown_table_names(text)
+    return bool(_unknown_table_names(text) & _MEMBERSHIP_TABLES)
 
 
 def _empty_edges_result(degraded_reason: str | None = None) -> WorkGraphEdgesResult:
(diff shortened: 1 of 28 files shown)


chrisgeo:
## Codex review: 3 HIGHs resolved (commit 831e2edd2)

All three findings were real edge cases of the run_id / completion-marker protocol. Fixed on this branch; mypy clean (630 files), work_graph unit suite 222 passed / 2 skipped, 4 live ClickHouse regressions pass. All OR'd-join + INSERT...SELECT SQL was verified against ClickHouse 26.5 before committing.

### 1. ALL-SKIPPED backfill can't supersede → empty-but-complete marker rule

**Rule:** the completion marker is published whenever the org **has work-graph components**, even when the run emits **zero** membership rows. An empty complete run is how the no-tombstone design retires churned nodes — it becomes the latest complete run and the previous run's stale rows stop matching. Only a **genuine no-component org** (no edges → early return) writes no marker. The materializer got the same treatment (a run that produced zero rows still publishes). `backfill.py` / `materialize.py` now gate the marker on "components were evaluated", not on "rows were written".

### 2. Org-wide latest-run marker unsafe under scoped runs → org-wide-only publishing

**Resolution chosen (the simplest acceptable one from the finding):** membership-publishing runs are **org-wide only**. A repo/team-scoped materialize or repo-scoped backfill writes its rows but does **not** publish the org-wide marker — otherwise it would become the org's latest complete run while covering only that scope, blanking every other repo for unscoped reads. Verified the post-sync dispatcher (`_dispatch_post_sync_tasks`) and the daily backfill dispatcher are both already org-wide, so the normal paths always publish; only manual/CLI `repo_ids`/`team_ids` runs skip. Added `MembershipBackfillConfig.is_org_wide` and the equivalent guard in the materializer.

### 3. Migration orphans existing rows → legacy seed marker + reader fallback

**Approach (lowest-risk, no heavy mutation):** new **migration 048** seeds one synthetic `__legacy__` completion marker per org that has pre-existing membership rows, via `INSERT INTO work_unit_membership_runs SELECT org_id, '__legacy__', max(computed_at) FROM work_unit_membership GROUP BY org_id`. No rewrite of the existing (run_id='') rows. The reader's run-scope join — shared across all 3 read paths via `_RUN_SCOPE_JOIN_ON` — recognises the `__legacy__` marker and matches the pre-existing `run_id=''` rows:

```
m.run_id = latest_run.latest_run_id
  OR (latest_run.latest_run_id = '__legacy__' AND m.run_id = '')
```

So existing membership stays filterable/annotated the instant 047+048 land. **Retirement:** the legacy marker's `completed_at` is the max historical timestamp (in the past); every real run publishes with `completed_at = now()` (strictly greater), so `argMax` selects the real run the moment one completes and the legacy path retires automatically — no cleanup, the legacy marker simply stops being the latest. Documented in the 048 docstring.

### New regression tests
- `test_membership_backfill.py`: empty-but-complete marker, genuine-no-component no-marker, scoped-run-no-marker, `is_org_wide` flag.
- `test_work_unit_membership.py`: materializer scoped-run-no-marker.
- `test_work_graph_theme_filter.py`: `TestLegacyRolloutFallback` — legacy `'__legacy__'` + `run_id=''` fallback present in all 3 reader paths (filter / annotation / degraded probe).
- `test_work_graph_membership_split_merge_live.py` (live ClickHouse): `test_empty_complete_run_supersedes_previous` (finding #1), `test_legacy_rows_readable_then_superseded_by_real_run` (finding #3).

chrisgeo:
## Round-2 review: 1 HIGH + 1 MEDIUM resolved (commit 30feba2e4)

mypy clean (630 files); work_graph unit suite 223 passed / 2 skipped; 42 migration tests pass; 5 live ClickHouse tests pass. The OPTIMIZE FINAL eviction was reproduced directly against ClickHouse 26.5 (old key drops the complete row, new key keeps both) before committing.

### 1. [HIGH] run_id missing from the dedup key → background merges break the protocol — rebuild migration 049

**Root cause confirmed:** `work_unit_membership` is `ReplacingMergeTree(computed_at) ORDER BY (org_id, node_type, node_id, category_kind, category)` (migration 046). run_id is a column (047) but **not** in the sort key, so a merge / `OPTIMIZE FINAL` collapses same-(org,node,category) rows across runs to the max-`computed_at` version — a newer **incomplete** run evicts the prior **complete** run's row, and the resolver scoped to the complete run_id then reads nothing.

**Fix — new migration 049** rebuilds the table with run_id appended:
```
ORDER BY (org_id, node_type, node_id, category_kind, category, run_id)
```
keeping `ReplacingMergeTree(computed_at)` (now it only dedups idempotent re-writes of the *same* run). Uses the **shadow-table + EXCHANGE TABLES** pattern from migration 042:
1. `SHOW CREATE TABLE` → rewrite name + ORDER BY
2. verify the shadow sort key is **exactly** old-key + `, run_id` (fail closed on a regex miss)
3. `INSERT SELECT *` (legacy `run_id=''` rows copied verbatim)
4. distinct-new-key-count check (new key is a superset → lossless)
5. `EXCHANGE TABLES` (atomic) → catch-up `INSERT` → `DROP` shadow

Idempotent (skips if run_id already last) and crash-convergent (leftover shadow → catch-up before skip). **Lossless** because 047/048/049 apply in the same migration pass before any non-empty run_id is written, so all pre-049 rows share `run_id=''` and cannot collapse across runs on the old key.

**Test (live):** `test_optimize_final_preserves_prior_complete_run_row` inserts the same node/category under a complete run + a newer incomplete run (no marker), runs `OPTIMIZE TABLE work_unit_membership FINAL`, then asserts the complete run's row is still present and the resolver scoped to it still returns the node. Fails on the 046 key, passes after the 049 rebuild.

### 2. [MEDIUM] Degraded probe mislabels a valid empty complete run — gate on marker existence

**Probe change:** previously keyed on `count(membership rows in latest run) > 0`, which flagged the intentionally-empty all-skipped supersede run as `MEMBERSHIP_NOT_MATERIALIZED`. The probe now **counts `work_unit_membership_runs` markers** and returns `MEMBERSHIP_NOT_MATERIALIZED` only when **no complete marker exists** while `investments > 0`. A marker present — even with zero membership rows — is a genuine empty (`degraded_reason=None`). The probe no longer joins `work_unit_membership` at all.

**Tests:** `test_empty_complete_run_is_not_degraded` (marker present, zero rows → None) and `test_degraded_state_when_membership_unmaterialized` (no marker + investments → degraded); all probe-feeding tests updated to the `complete_run_markers` gate.

### Bonus fix
Migration 048 now **skips gracefully** when `work_unit_membership` is absent (fresh DB / dry-run mock) instead of failing closed — a no-membership DB is a valid no-op. Fixes `test_clickhouse_migrations_dry_run_non_default_db`.

chrisgeo:
## Round-3 review: 2 HIGHs resolved (commit 78fdc7a7c)

mypy clean (630 files); work_graph + fixtures + dispatch unit suites 211 passed / 2 skipped; 7 live ClickHouse tests pass. All argMax/overlap behaviour verified against live ClickHouse.

### completed_at source (finding #1)
The marker's `completed_at` is now stamped at the **actual marker write** with `datetime.now(timezone.utc)`, captured **after** all membership rows are persisted — not the run-start `computed_at`. The membership **rows** keep their run-start `computed_at`; only the **marker** carries the completion timestamp. So `argMax(run_id, completed_at)` reflects **completion order**: a run that starts earlier but finishes later now publishes the greater timestamp and wins, eliminating the early-stamp race.

### Writer unification (finding #2 — chose the recommended option)
The no-LLM full-coverage **projection** (`backfill.py`) is now the **sole writer** of `work_unit_membership` rows **and** completion markers in both paths:
- **post-sync:** `build -> materialize (LLM; writes work_unit_investments ONLY) -> project-membership (full coverage; writes membership + marker)`
- **daily:** `build -> project-membership` (unchanged)

The materializer's membership-row + marker write is **removed** — it persists `work_unit_investments` only. Because the projection iterates the **full current work graph** (no `from_ts`/`to_ts` window), its org-wide marker is legitimately full-coverage, so a date-windowed materialize can never publish partial coverage that blanks out-of-window components. Post-sync stays fresh: the projection runs right after materialize persists the new investments.

Also fixed the fixtures path (`generators/investments.py` + `runner.py`): seeded membership rows now carry a `run_id` and the runner publishes a matching completion marker, so fixture/e2e theme filtering works under the run-scoped resolver (previously seeded rows had `run_id=''` and no marker → invisible).

### #925 assumption check — NOT broken
Removing the materializer's membership write does not break merged #925 (CHAOS-2427). That work only needs `work_unit_membership` populated so theme filters leave the `MEMBERSHIP_NOT_MATERIALIZED` degraded state. The projection now populates it (with full coverage) in **both** the post-sync and daily chains, so the guarantee holds and is in fact stronger (full-coverage vs windowed). The only observable change is the materialize stats dict dropping its `"memberships"` key — no caller reads it.

### Tests
- **Unit (backfill):** `test_backfill_marker_completed_at_is_completion_time_not_run_start`, `test_backfill_overlap_later_finisher_publishes_later_marker`, `test_backfill_projects_full_current_coverage_no_time_window`.
- **Unit (materializer):** rewritten to assert investments-only — writes NO membership rows and NO marker; a sink that raises on either method is never called.
- **Unit (post-sync dispatch):** updated to the 3-step `build -> materialize -> project` chain (project is immutable, org-scoped, on the metrics queue, runs last).
- **Live ClickHouse:** `test_overlap_later_marker_wins_over_earlier_marker` (a run whose rows are NEWER but whose marker is OLDER LOSES to a run whose rows are older but marker is newer — selection follows the marker) and `test_full_coverage_projection_does_not_blank_out_of_window_nodes` (an out-of-window component stays filterable under the full-coverage projection marker).

chrisgeo:
## Round-4 review: 1 HIGH + red CI mypy resolved (commit aec10f056)

CI-style full-tree mypy (`mypy --install-types --non-interactive .`, **1076 source files**) is clean. Runner suite 9 passed; membership/dispatch unit suites 48 passed; 7 live ClickHouse tests pass.

### CLI projection wiring (finding #1)
After the round-3 writer-unification, `materialize_investments` writes `work_unit_investments` only — the no-LLM projection owns membership rows + the marker. The post-sync Celery chain got the 3rd projection step, but the operator entry point `run_investment_materialization` (`work_graph/runner.py`) still called `materialize_investments` alone, so `dev-hops investment materialize` left theme filters reading the stale/previous marker.

Fix: after a successful **org-wide full-coverage** materialization, the CLI now invokes `backfill_memberships(...)` **synchronously** (a direct function call, not a Celery dispatch, so the CLI is complete on return) to publish a matching full-coverage completion marker. Per the coverage rule, a **repo/team-scoped or date-windowed** manual materialize refreshes investments only and does **not** publish an org-wide marker — the org-wide daily projection republishes full coverage.

Gating detail: `--window-days` now defaults to `None` (the 30-day fallback is applied in code), so a bare `investment materialize` is detected as full-coverage while `--from`/`--to`/`--window-days` marks the run windowed. `is_full_org_wide = not repo_ids and not team_ids and not explicitly_windowed`.

### CI mypy
Removed the two unused `# type: ignore[override]` comments at `tests/work_graph/test_work_unit_membership.py` lines 379/384 (the FakeSink override signatures are compatible). Ran the **full CI invocation from the repo root** (not a src-only subset) and fixed the one additional error it surfaced (a `[var-annotated]` on a new test helper dict). Full tree is now clean.

### Tests
- `test_cli_org_wide_materialize_runs_membership_projection` — org-wide run invokes the projection with an org-wide config (`repo_ids=None`, `is_org_wide=True`); marker published.
- `test_cli_repo_scoped_materialize_skips_org_marker` / `test_cli_team_scoped_materialize_skips_org_marker` / `test_cli_date_windowed_materialize_skips_org_marker` — scoped/windowed runs refresh investments only, projection NOT called.
- `test_cli_materialize_failure_skips_projection` — a materialize failure skips the projection and returns rc=1.

chrisgeo:
## Round-5 review: 1 HIGH (unbounded growth) resolved (commit ec2e2b17b)

CI-style full-tree mypy (`mypy --install-types --non-interactive .`, **1076 source files**) is clean. Unit suite 235 passed / 2 skipped; **11 live ClickHouse tests pass** (4 new retention regressions). The `ALTER ... DELETE`, `Array(String)` param, and in-flight survival were verified directly against ClickHouse 26.5.

### Retention mechanism
New sink method `prune_membership_runs(org_id, keep=2)`:
1. Reads the org's completion markers from `work_unit_membership_runs` ordered by `completed_at DESC`, keeps the latest 2 run_ids.
2. `ALTER TABLE work_unit_membership DELETE WHERE org_id = ? AND run_id IN <dropped>` (lightweight async mutation).
3. `ALTER TABLE work_unit_membership_runs DELETE WHERE org_id = ? AND run_id IN <dropped markers>`.

Both deletes are scoped to `org_id` + the explicit dropped run_id set → idempotent and safe to run concurrently across orgs (each org only ever touches its own run_ids). `backfill_memberships` calls it **immediately after publishing the org-wide marker** (org-wide runs only — scoped runs publish no marker, so they don't prune). Retention is **best-effort**: a prune failure logs and is swallowed so it never fails the projection (the marker is already correct; the next prune is idempotent and catches up).

### keep-latest-2 rationale
Keep the current latest run **plus one prior**, so a reader/overlap mid-flight against the immediately-previous complete run — an `argMax` that resolved to it microseconds before a newer marker landed — is not pulled out from under it. Keeping 1 would risk deleting the run a concurrent reader is actively scanning. (This is the read-side complement to the round-3 completion-time marker fix.)

### in-flight-safety guard
The candidate delete set is sourced **exclusively** from `work_unit_membership_runs` (markered runs). A markerless in-flight run — rows written, no marker yet, about to become the next generation — is **never** in the delete set, so a concurrent retention pass can never delete its rows. (The delete is `run_id IN <markered-runs-to-drop>`, never a blanket "not in keep set".) Legacy: the `__legacy__` marker maps to `run_id=''` rows for the row-delete; it ages out naturally once 2 real runs exist.

### Tests
- **Unit:** prune runs after the marker (keep=2, org-scoped, order `[memberships, run_marker, prune]`); empty all-skipped run still prunes; scoped run does NOT prune; prune failure is non-fatal (projection still succeeds).
- **Live ClickHouse:** after 4 runs only the latest 2 runs' rows **and** markers remain; a markerless in-flight run survives a concurrent prune; the resolver still returns the latest run **and** the prior complete run survives (keep=2 margin) while the 2 oldest are gone; row count stays bounded (`<= keep*nodes`) across 6 successive prune-after-publish cycles (would be `6*nodes` unpruned).

chrisgeo:
## Round-6 review: 1 HIGH (migration robustness) resolved (commit b5bc1fa1d)

CI-style full-tree mypy (`mypy --install-types --non-interactive .`, **1078 source files**) is clean. Migration + membership + backfill + GraphQL suites 286 passed / 2 skipped; **11 live ClickHouse tests pass** (the live fixture re-applies 049 with the new post-rebuild verification against the real DB).

### Fail-closed change
The root cause was `049._table_exists()` catching **any** exception from the `system.tables` probe and returning `False` ("table absent") → `upgrade()` returned early → the runner recorded 049 as applied. A transient probe failure or a restricted ClickHouse user then left `work_unit_membership` keyed **without** run_id while the app started run-scoped reads/writes, reintroducing the round-2 background-merge eviction.

Fix:
- **049 `_table_exists()` no longer swallows exceptions** — an unexpected `system.tables` probe error now PROPAGATES, so `upgrade()` RAISES and the migration is NOT recorded as applied (retryable). The table is treated as absent ONLY when a **successful** probe returns zero rows (the genuine fresh-DB / dry-run no-op path).
- **Added a final post-rebuild verification**: after `EXCHANGE` + catch-up, re-read the **live** main-table sorting key and RAISE unless run_id is last — proving the rebuild actually landed before the runner can mark 049 applied. (`_sorting_key` already raises if `system.tables` returns no rows, so this path fails closed too; the pre-copy shadow-key verification is unchanged and still fails closed.)
- **Applied the same discipline to 048's sibling probe** (a swallowed error would mark the legacy-seed migration applied without seeding). **047 is pure SQL** — its errors already propagate via the runner, no change needed.

The genuine zero-row skip (idempotent re-run / dry-run mock) path is preserved, so `test_clickhouse_migrations_dry_run_non_default_db` still passes (the mock's `result_rows=[]` is a successful empty probe, not an error).

### Tests (fake client, no live DB)
- **049:** existence probe raises an unexpected error → `upgrade` RAISES and no rebuild is attempted; restricted-user error propagates; successful zero-row probe → genuine skip; post-rebuild key missing run_id → RAISES (`post-rebuild verification failed`); healthy rebuild completes (EXCHANGE issued); already-migrated → idempotent skip.
- **048:** probe error propagates (no seed INSERT); zero-row → genuine skip; present tables → marker seeded.

chrisgeo:
## Round-7 assessment: FALSE POSITIVE — confirmed by trace + new regressions (commit d96d26805)

Your reasoning is correct on all four points. The claimed coverage hole cannot occur. This is a **test-only** change (no source change); I added definitive regressions so it can't silently regress.

### Trace (verified against the code)
1. **The backfill investment read is NOT date-windowed.** `backfill._fetch_latest_distributions` (`backfill.py:143-177`) is `SELECT ... FROM work_unit_investments WHERE org_id = %(org_id)s AND work_unit_id IN %(work_unit_ids)s GROUP BY org_id, work_unit_id` with `argMax(..., computed_at)` — latest-per-unit, **no `from_ts`/`to_ts`/`computed_at` bound**. The `work_unit_ids` set comes from `_build_components_for_backfill(fetch_work_graph_edges(...))`, and `fetch_work_graph_edges` (`queries.py:17-51`) has **no date filter** either. So the backfill covers the full current graph and reads the latest investment row regardless of age. ✔ unwindowed.
2. **A component is skipped ONLY when its `work_unit_id` has no investment row** (`backfill.py:243-250`: `if persisted is None: skipped; continue`). That means "never categorized" → correctly has no theme → absent from theme filters is correct, not "hiding valid work." ✔
3. **A previously-visible component becomes skipped only via CHURN** — `work_unit_id(...)` is a hash of the component's node set; an unchurned component keeps the same id and matches its existing investment row. A churned id has no row → absent. That's the deliberate round-2 churned→uncategorized→absent (run-scoping over recency) behavior. ✔
4. **Codex's fix contradicts round-2.** Round-2 finding #1 *requires* publishing an empty-but-complete marker even when all components are skipped (to retire stale rows). "Suppress marker on any skip" would break the empty-complete supersede. Rejected. ✔

**The decisive point:** per the round-3 writer-unification, the materializer writes `work_unit_investments` **only** — its 30-day window governs which units get *re-categorized* this run, not which old investment rows persist (`work_unit_investments` is a per-unit ReplacingMergeTree, never truncated). The backfill then reads *all* of them unwindowed, so an out-of-window-but-previously-categorized component still has its row and **is projected (covered)**.

### Regressions added
- **Unit** (`test_membership_backfill.py`): the investment-read SQL carries no temporal window (only org_id + work_unit_id); and a previously-categorized **unchurned** component is matched/covered while a genuinely-uncategorized one is skipped — with the org-wide marker still published.
- **Live ClickHouse** (`test_work_graph_membership_split_merge_live.py`): seed `work_graph_edges` + a `work_unit_investments` row with `computed_at = 2024-01-01` (far outside any 30-day window), run the **real** `backfill_memberships`, then assert the resolver returns the component under its theme (`matched == 1`, edge present, not degraded) — proving the org-wide marker covers it.

CI-style full-tree mypy clean (1078 files); membership/backfill/GraphQL/work_graph suites 249 passed / 2 skipped (incl. 12 live ClickHouse tests).

chrisgeo:
## CI red (test-matrix) fixed — real bug, not a gating issue (commit 86dc94caa)

### The actual failing test
Pulled `gh run view 27579877858 --log-failed`. CI showed **exactly one** failure across all Python versions:
```
tests/test_teams.py::test_clickhouse_store_teams
TypeError: '>' not supported between instances of 'MagicMock' and 'int'
= 1 failed, 4733 passed, 43 skipped, 50 deselected =
```
The leading hypothesis (ungated live tests) is **ruled out**: my live tests carry the `clickhouse` marker and are correctly deselected — the log shows `50 deselected`, and the test-matrix lane actually *has* a ClickHouse service anyway. The live lane is separate and unaffected.

### Root cause — a real bug from my round-6 fail-closed change
`test_clickhouse_store_teams` drives **all** ClickHouse migrations through `ClickHouseStore.__aenter__ → _ensure_tables` with a bare `MagicMock` client. Pre-round-6, `048/049._table_exists` wrapped the probe in `try/except: return False`, which swallowed the `MagicMock > int` TypeError (treated as "absent → skip"). My round-6 change removed that `try/except` to fail closed on real probe errors — but it also removed the mock tolerance, so the migration now crashes under a MagicMock client. (A naive `int()` coercion is worse: `int(MagicMock())` returns `1`, so the table looks "present" and 049 proceeds into the rebuild and crashes in the DDL regex.)

### Fix — type-strict, exception-safe interpretation
New shared `_count_gt_zero(result_rows)`: a successful probe is "present" ONLY when `result_rows` is a real `list`/`tuple` whose first cell is a genuine `int` (>0, bool excluded). Any other shape (a MagicMock `result_rows`, a non-numeric cell) is treated as **absent**, not raised on — because the probe *call* succeeded. A genuine query **error** still raises from `client.query()` and propagates, so round-6's fail-closed-on-DB-error guarantee is preserved (the existence-probe-raises and post-rebuild-verification tests still pass). Applied to both 048 and 049.

### The other locally-observed failures are NOT branch regressions
While reproducing the unit lane locally I also saw `test_org_deletion`, `test_cli_preflight`, and `test_core_extraction` fail — but the CI log confirms all of these **PASSED in CI** (CI was `1 failed, 4733 passed`). They are local-environment artifacts: stale local Postgres data, a missing/invalid 32-byte `LICENSE_PRIVATE_KEY`, and full-suite test-ordering pollution (`test_core_extraction` passes in isolation). No change needed.

### Verification
- **Full-tree CI mypy** (`mypy --install-types --non-interactive .`): clean, **1078 files**.
- **Unit lane** (CI-equivalent, `-m "not benchmark and not clickhouse"`): `test_teams` now passes; green except the pre-existing local-env-only artifacts above (which pass in CI).
- **Live lane** (`-m clickhouse`): my 12 `test_work_graph_membership_split_merge_live.py` tests all pass (they self-seed). The 8 other live failures locally are pre-existing seeded-demo-data-dependent tests (`test_flow_matrix_live`, `test_recommendations_tombstone`, `test_resolver_sql_explain[operating_review]`) unrelated to this PR.
- New regression: `test_magicmock_client_is_treated_as_absent_not_crash` (the exact CI-red shape) + type-strict `_count_gt_zero` unit tests in both migration test files.

chrisgeo:
## Final review HIGH (legacy fallback over-match) fixed (commit b55d4bd29)

### The bug
Migration 048 seeds a `__legacy__` marker and leaves pre-existing rows at `run_id=''`. The legacy branch matched **all** `run_id=''` rows. But the OLD pre-protocol key was `(org, node, category_kind, category)` with run_id NOT in it, so a node that was dominant category A at T1 and re-materialized to dominant B at T2 has **both** rows persisted (distinct category values → distinct keys → never deduped), all `run_id=''`. The OLD reader scoped each node to `max(computed_at)`. The blanket match resurfaced stale A alongside current B → wrong filters + multiple `is_dominant` rows per node, for idle/migrated orgs during the rollout window.

### The legacy-vs-real conditional (option b — no data mutation)
`_RUN_SCOPE_PREDICATE`:
```
(latest_run.latest_run_id != '__legacy__' AND m.run_id = latest_run.latest_run_id)   -- REAL run: run_id equality (one generation per run_id)
OR (latest_run.latest_run_id = '__legacy__' AND m.run_id = '' AND m.computed_at = lnm.legacy_max_computed_at)  -- LEGACY: that node's latest pre-migration row only
```
`lnm` is a `LEFT JOIN` to an inline `(org, node) → max(computed_at) WHERE run_id=''` derived table (`_LEGACY_NODE_MAX_JOIN`) — the exact OLD per-node-latest behavior, NOT a blanket `run_id=''` match. The `latest_run` join becomes a single-row cross join (`ON 1=1`) with the predicate + legacy-node-max join moved into the WHERE.

### 3-path application
- **Filter EXISTS semi-join** (`_theme_membership_exists_clause`) — fixed.
- **Dominant-annotation lookup** (`_batch_resolve_membership`) — fixed.
- **Degraded probe** (`_detect_membership_degraded_reason`) — **no change needed**: it was redesigned in round-2 to gate purely on marker existence (`count(work_unit_membership_runs)`), so it does NOT join membership rows and the over-match cannot occur there. (Noting this explicitly since the review said "all three" — the third reads no membership rows.)

Verified the EXISTS semi-join and annotation query both produce correct results with the per-node-max guard against ClickHouse 26.5.

### Regression results
Live ClickHouse `test_legacy_per_node_max_excludes_stale_dominant`: a node seeded with OLD dominant A (`maintenance @ T1`) and NEWER dominant B (`feature_delivery @ T2`), both `run_id=''`, only the `__legacy__` marker present. Asserts under legacy:
- filter for A (`maintenance`) → **no node**,
- filter for B (`feature_delivery`) → returns the node,
- annotation reports **exactly one** dominant (B), no stale-A duplicate (via the resolver AND a direct per-node-max DB query),
- then a real run supersedes legacy entirely (A and B both gone, `quality` wins).

Unit tests pin the per-node-max guard SQL (`legacy_max_computed_at`, `m.computed_at = lnm.legacy_max_computed_at`) in both the filter and annotation paths.

**CI-style full-tree mypy clean (1078 files); theme-filter + membership + backfill + work_graph + live membership suites 244 passed / 2 skipped (13 live ClickHouse tests).** This should be the last item.

PR #938, opened 2026-06-16
fix(api): serialize ExplainResponse before caching (CHAOS-2456)

chrisgeo:
## Summary

The Redis-backed explain cache silently failed on **every** write, so explanations were recomputed against ClickHouse on every `/api/v1/explain` request (the cache was dead, not stale).

`build_explain_response()` cached the raw `ExplainResponse` (a Pydantic `BaseModel`) via `cache.set(cache_key, response)`. `RedisBackend.set()` serializes with `json.dumps(value)`, which cannot encode a `BaseModel`, so the write threw `Object of type ExplainResponse is not JSON serializable`, was swallowed by the broad `except`, and only logged the message.

```
{"level": "WARNING", "name": "dev_health_ops.core.cache", "message": "Redis set failed: Object of type ExplainResponse is not JSON serializable"}
```

## Changes

- **`services/explain.py`** — cache `response.model_dump(mode="json")`; on cache hit return `ExplainResponse.model_validate(cached)`, preserving the `-> ExplainResponse` return contract. Mirrors the existing `services/home.py` pattern exactly.
- **`core/cache.py`** — enrich the swallowed "Redis set failed" warning with the cache `key`, `value_type`, `ttl`, and `exc_info=True` so this class of failure is traceable from logs alone (the original line had no traceable context).
- **`tests/api/services/test_explain_cache_serialization.py`** — regression test: proves the explain payload round-trips through the JSON backend, documents that a raw `ExplainResponse` is not serializable, and asserts the enriched warning includes key + value type + traceback.

## Verification

- LSP clean on all 3 files
- `ruff format --check` + `ruff check` clean
- New regression tests: 3 passed
- Existing cache/explain tests: 25 passed (no regressions)

SCREENSHOT-WAIVER: backend-only change (cache serialization + logging); no rendered frontend output is affected.

Closes CHAOS-2456

### DIFF
--- a/src/dev_health_ops/api/services/explain.py
+++ b/src/dev_health_ops/api/services/explain.py
@@ -124,7 +124,7 @@ async def build_explain_response(
     cache_key = filter_cache_key("explain", org_id, filters, extra={"metric": metric})
     cached = cache.get(cache_key)
     if cached is not None:
-        return cached
+        return ExplainResponse.model_validate(cached)
 
     config = _METRIC_CONFIG.get(metric, _METRIC_CONFIG["cycle_time"])
     start_day, end_day, compare_start, compare_end = time_window(filters)
@@ -239,7 +239,7 @@ async def build_explain_response(
         },
     )
 
-    cache.set(cache_key, response)
+    cache.set(cache_key, response.model_dump(mode="json"))
     return response
 
 
--- a/src/dev_health_ops/api/services/explain.py
+++ b/src/dev_health_ops/api/services/explain.py
@@ -1,7 +1,10 @@
 from __future__ import annotations
 
+import logging
 from collections.abc import Callable
 from typing import TypedDict
+
+from pydantic import ValidationError
 
 from ..models.filters import MetricFilter
 from ..models.schemas import Contributor, ExplainResponse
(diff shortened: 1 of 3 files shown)


chrisgeo:
## Adversarial review follow-up (Codex, gpt-5.5)

Ran `/codex:adversarial-review` against this PR. Verdict was **needs-attention** with three medium findings — all valid, all now addressed in `f62be21`:

1. **Cache log leak (I introduced it):** the enriched "Redis set failed" warning logged the raw cache key, which `filter_cache_key` builds from the serialized filter payload (`_org_id` + user-controlled scope/repo/developer filters). Now routed through a `_safe_key_label()` helper that logs only the prefix + a short sha256 digest — repeated failures still correlate, no tenant data in logs.
2. **Brittle cache hit:** `model_validate` on a stale-schema/corrupt entry raised, which the endpoint surfaced as a 503 until TTL expiry. Now wrapped in `try/except ValidationError` → degrades to a recompute (miss-on-invalid).
3. **Weak test:** the original round-trip test mirrored the fix but never exercised `build_explain_response`, so a revert of the service line would still pass. Added service-level regression tests that drive `build_explain_response` with mocked queries + a fake Redis-backed cache (asserts JSON is stored and the 2nd identical call is a cache hit), plus a recompute-on-invalid-entry test.

Verification: lsp clean, ruff clean, 30 tests pass (5 new/updated + 25 existing, no regressions).

PR #1059, opened 2026-06-26
fix(github): retry transient 5xx at REST request level (CHAOS-2704)

chrisgeo:
## Summary

GitHub REST calls were built with `retry=None`, so a single transient `503`/`502`/`504` failed the sync unit immediately — while the GraphQL path already retried via `@retry_with_backoff`. This restores symmetry on the REST path and is the main driver of the GitHub-sync 503 floods (CHAOS-2704).

## Changes

- Build the PyGithub client with a bounded `urllib3` `Retry` scoped to **idempotent reads** (`GET`/`HEAD`/`OPTIONS`) and **`502`/`503`/`504` only**, with bounded exponential backoff and an explicit HTTP timeout.
- `respect_retry_after_header=False` is **intentional**: urllib3 v2 would otherwise transport-retry `413`/`429`/`503` carrying a `Retry-After` even when absent from `status_forcelist`, stealing rate-limit handling away from `RateLimitException` → the worker deferral path, and sleeping for an unbounded `Retry-After` inside a single `RateLimitGate` call. Rate-limit/`4xx` stays owned by `RateLimitGate` / `_raise_github_exception`.
- `raise_on_status=False` so a persistent final `5xx` still surfaces as PyGithub's `GithubException` → `APIException`, unchanged (no behavior regression when GitHub is genuinely down).
- Mutations are never retried (read-only allowlist).

Env-tunable: `GITHUB_HTTP_MAX_RETRIES` (3), `GITHUB_HTTP_BACKOFF_FACTOR` (1.0), `GITHUB_HTTP_BACKOFF_MAX` (30), `GITHUB_HTTP_TIMEOUT_SECONDS` (30).

## Testing

- New unit tests: retry config (status_forcelist/allowed_methods/env overrides/`backoff_max`), `429 + Retry-After` is **not** retried, `503` is retried, `503` exhaustion behavior, timeout helper, and both `Github(...)` constructor branches receive a non-None `retry` + `timeout`.
- Updated two pre-existing `test_github_pr_social_batch.py` tests that asserted the old `retry=None` contract.
- `ci/local_validate.sh` gates 1-3 green (ruff format/check, mypy, full unit suite). ClickHouse stage skipped — zero ClickHouse/attribution/migration surface.

## Review notes

- Design + edge cases (respect_retry_after_header, backoff bound) reviewed by Oracle; must-fixes applied.
- Follow-up filed: GitHub App token exchange is single-shot (CHAOS-2706).

SCREENSHOT-WAIVER: backend-only change, no UI/render surface.

Closes CHAOS-2704
Closes CHAOS-2706

### DIFF
--- a/src/dev_health_ops/connectors/utils/github_app.py
+++ b/src/dev_health_ops/connectors/utils/github_app.py
@@ -8,13 +8,24 @@ from typing import Any
 import jwt
 import requests
 
+from dev_health_ops.connectors.utils.retry import retry_with_backoff
+
 GITHUB_API_BASE_URL = "https://api.github.com"
 JWT_TTL_SECONDS = 600
 INSTALLATION_TOKEN_REFRESH_WINDOW_SECONDS = 300
+TOKEN_EXCHANGE_MAX_RETRIES = (
+    3  # total attempts (initial + retries), per retry_with_backoff
+)
+TOKEN_EXCHANGE_INITIAL_DELAY_SECONDS = 1.0
+TOKEN_EXCHANGE_MAX_DELAY_SECONDS = 10.0
 
 
 class GitHubAppAuthError(RuntimeError):
     """Raised when GitHub App authentication cannot produce an access token."""
+
+
+class GitHubAppTransientError(GitHubAppAuthError):
+    """Transient GitHub App auth failure (5xx / network) that is safe to retry."""
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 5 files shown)


chrisgeo:
CHAOS-2706

Scope update: this PR now also includes **CHAOS-2706** — a bounded retry for the GitHub App installation-token exchange (`connectors/utils/github_app.py`), the single-shot gap surfaced during this PR's Oracle review. New `GitHubAppTransientError` marks 5xx/network failures retryable (3 attempts, exp backoff capped 10s); 4xx (401/403/404) never retried; an exhausted transient still raises `GitHubAppAuthError` so callers are unaffected. Oracle-reviewed (SHIP-WITH-NITS, nits applied); full local gate green (5797 passed). Closes CHAOS-2706. (commit 61159b1c5)

PR #1080, opened 2026-06-29
fix(sync): scope backfill work-item units to their source (CHAOS-2725)

chrisgeo:
## Summary

Fixes backfills attempting to sync **all items** from a sync config (projects + work items) instead of only the scoped subset. The admin/API backfill path already routes through the planner's `source × dataset × window` scoping, but the provider layer ignored the per-unit source — so a scoped backfill still fanned out.

Linear: CHAOS-2725

## Root cause

Scoped work-item units **did** carry `repo_name`, but the provider interpretation dropped it:

- **Linear** treated work-item units as `IngestionContext(repo=None)` → all teams.
- **GitHub / GitLab** repo discovery fanned out org-wide when no `repo_id` was present, ignoring `repo_name`.

(Jira already scoped correctly via `jira_project_keys`.)

## What changed

- `metrics/job_daily.py` — repo discovery now applies `repo_name` as an exact ClickHouse `repos.repo` filter when no `repo_id` is available, stopping the org-wide `_discover_repos` fan-out for scoped units.
- `metrics/job_work_items.py` — threads `repo_name` into the Linear `IngestionContext.repo`, so Linear syncs only that team instead of all teams.

Jira and GitLab provider behavior is unchanged.

## Scope boundary

This is the surgical source-scoping fix for the admin/API planner path. The structural reference/discovery-tier rework (fetch teams/orgs/projects/cycles once per run) remains owned by EPIC CHAOS-2719 (scope A). The legacy CLI runner path that bypasses planner scoping is tracked separately in CHAOS-2726.

## Testing

- New regression tests: `tests/test_backfill_integration.py` (repo_name discovery filter), `tests/test_dataset_adapters.py` (Linear source scope adapter).
- `bash ci/local_validate.sh` green: ruff, mypy, full unit suite (5982 passed / 44 skipped), live argMax proof. Provider matrix `{jira,gitlab,github,linear} × {teams,projects,members,issues}` stays green.

SCREENSHOT-WAIVER: backend-only change (metrics/ ingest scoping); no web UI render.

### DIFF
--- a/src/dev_health_ops/metrics/job_daily.py
+++ b/src/dev_health_ops/metrics/job_daily.py
@@ -119,9 +119,15 @@ def discover_repos(
     try:
         query = "SELECT id, repo, settings, provider FROM repos"
         params: dict[str, str] = {}
+        filters: list[str] = []
         if org_id:
-            query += " WHERE org_id = {org_id:String}"
+            filters.append("org_id = {org_id:String}")
             params["org_id"] = org_id
+        if repo_name:
+            filters.append("repo = {repo_name:String}")
+            params["repo_name"] = repo_name
+        if filters:
+            query += " WHERE " + " AND ".join(filters)
         rows = primary_sink.client.query(query, parameters=params).result_rows
         return [
             DiscoveredRepo(
--- a/src/dev_health_ops/metrics/job_work_items.py
+++ b/src/dev_health_ops/metrics/job_work_items.py
@@ -526,7 +526,7 @@ def run_work_items_sync_job(
             )
             ctx = IngestionContext(
                 window=IngestionWindow(updated_since=since_dt, active_until=until_dt),
-                repo=None,
+                repo=repo_name,
                 org_id=uuid.UUID(org_id) if org_id else None,
             )
             fetched_items = 0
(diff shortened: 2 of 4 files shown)


chrisgeo:
Closing — not merging. Adversarial review (codex, gpt-5.5) returned **needs-attention / no-ship**:
- [high] The fix overloads `repo_name` as provider identity, which **silently breaks GitLab unit backfills**: GitLab sources use numeric `project_id` as external_id while repos are stored as `path_with_namespace`, so `repos.repo = '123'` matches zero repos and the unit writes no data (job_daily.py:126-128).
- [medium] GitHub Projects-V2 still fans out outside the scoped repo unit — the GitHub branch unconditionally merges all `GITHUB_PROJECTS_V2` items (job_work_items.py:460-471).

Codex: it *'masks the reference-tier issue rather than fixing it at the unit boundary.'*

Decision: fold this into **CHAOS-2719** scope C (source-scoping) + scope A (reference tier), where provider-aware source resolution (GitLab project_id/path, GitHub full names, Jira keys, Linear team keys) belongs — instead of shipping a leaky point-fix we'd later rip out.

Branch `fix/chaos-2725-scoped-backfill` is **retained** as a reference for the real CHAOS-2719 PR work. Tracking: CHAOS-2725.

PR #1123, opened 2026-07-02
fix(sync): redact credential material from legacy error-text columns (CHAOS-2766)

chrisgeo:
## Summary

Live verification of CHAOS-2758 (evidence comment on CHAOS-2742) demonstrated that a provider exception whose message embeds an Authorization header (`403 rate limited -- Authorization: Bearer ghp_FAKE...`) persists **verbatim** into the pre-existing `sync_run_units.error` column. The new observation store's `reason` field is allow-listed (`_normalized_rate_limit_reason`), but this legacy free-form error column had no equivalent protection.

- Adds a single shared `sanitize_error_text` helper (`src/dev_health_ops/sync/error_sanitize.py`) — a **redaction pass**, not an allow-list, since these columns are free-form operator diagnostics rather than a closed enum. Strips/masks Authorization/Proxy-Authorization headers, bare Bearer tokens, HTTP Basic auth blobs, GitHub PAT prefixes (`ghp_`/`gho_`/`ghu_`/`ghs_`/`ghr_`/`github_pat_`), `glpat-`, `xox[baprs]-`, and `token=`/`private_token=`/`api_key=`/`access_token=`/`secret=` key/value pairs, then applies a length cap (redaction always runs *before* truncation).
- Wires the helper into every real persistence site across `workers/sync_units.py`, `workers/reference_discovery.py`, and `sync/dispatch_outbox.py` — DB writes to `sync_run_units.error`, `sync_runs.error`, `sync_run_reference_discovery.error`, `sync_dispatch_outbox.last_error`, plus the matching task-return-payload `error` fields.
- Adds an AST-based contract guard test (`tests/test_error_sanitize_guard.py`, mirroring the CHAOS-2757 docs-drift-guard precedent) that fails if any future raw `str(exc)`/bare `f"{exc}"` write bypasses the helper, with documented exemptions for the two classification-only helpers (`_classify_error`, `_is_retryable_discovery_error`) that never persist the raw text.
- Documents the design decision in `docs/providers/rate-limit-policy.md` (new "Legacy error-text columns" section), including an explicit scope note on what's intentionally *not* covered (`MetricCheckpoint.error`, `InvestmentBatchJob.error` — different subsystems, different exception provenance).

Scope is intentionally limited to the sync-run/unit execution ledger (the tables CHAOS-2742 hardens); it does not touch the observation store's `reason` allow-list semantics, and does not touch diagnostic `logger.*(extra={"error": str(exc)})` call sites (different risk surface, out of this ticket).

## Test plan

- [x] `tests/test_error_sanitize.py` — table-driven pattern coverage for every documented secret shape, the exact live-verify-case-B regression (fake bearer token → redacted, not persisted), ordinary-text-passes-through-readably, and truncate-after-redact.
- [x] `tests/test_error_sanitize_guard.py` — AST contract guard: no raw exception stringification outside logging across all 6 target files.
- [x] New regression tests in `tests/test_sync_units.py` and `tests/test_reference_discovery_stage.py` mirroring the live-verify case through the real Celery task entrypoints.
- [x] Updated 6 pre-existing test assertions that hardcoded pre-redaction error text to the new `ClassName: message` format.
- [x] `env -i HOME=$HOME PATH=$PATH USER=$USER TMPDIR=$TMPDIR SCRATCH_DB=ci_local_validate_2766 bash ci/local_validate.sh` → **GATE PASSED** (ruff format/check, mypy, full unit suite 6288 passed / 44 skipped, live ClickHouse argMax proof).

Closes CHAOS-2766.

### DIFF
--- a/docs/providers/rate-limit-policy.md
+++ b/docs/providers/rate-limit-policy.md
@@ -329,6 +329,123 @@ deleted outright, no archival path.
 **Cooldown gating reads this table back before dispatch** — see the next
 section.
 
+## Legacy error-text columns (`sync_run_units.error` and siblings)
+
+Shipped in [CHAOS-2766](https://linear.app/fullchaos/issue/CHAOS-2766). Live
+verification of CHAOS-2758 above (evidence comment on CHAOS-2742) found that
+the *observation store*'s `reason` allow-list did its job, but a
+provider exception whose message embeds an `Authorization` header (e.g.
+`403 rate limited -- Authorization: Bearer ghp_FAKE...`) still landed
+**verbatim** in the pre-existing, free-form `sync_run_units.error` column —
+these legacy columns predate CHAOS-2742 and were never brought under the same
+discipline.
+
+**Redaction, not an allow-list — deliberately.** `reason` above is a closed,
+normalized enum with no diagnostic-text mandate, so allow-listing a fixed
+vocabulary is correct there. `sync_run_units.error` and its siblings
+(`sync_runs.error`, `sync_run_reference_discovery.error`,
+`sync_dispatch_outbox.last_error`) are free-form, operator-facing
+diagnostics — the entire point of persisting them is to help debug a failed
+sync without re-running it. Collapsing them to a category string would defeat
+that purpose. Instead, `dev_health_ops.sync.error_sanitize.sanitize_error_text`
+(one shared helper, imported at every persistence site — no per-site regexes)
+strips/masks only the specific credential-shaped substrings that must never
+reach the database:
+
+- `Authorization` / `Proxy-Authorization` headers, any scheme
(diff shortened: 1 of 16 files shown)


chrisgeo:
Resolving the 8 Gitleaks bot threads: all findings reference commits (8223aafa2, fb102936a, 75e636f7e) that no longer exist on this branch — they were fixture-iteration commits containing SYNTHETIC test values for the redaction test suite (never real secrets; nothing to rotate). The branch history was rewritten to a single clean commit (293244aa2) precisely so no commit in the PR range introduces a secret-shaped literal; fixtures are now runtime-constructed with neutral identifiers. Verified: local gitleaks with CI's exact invocation reports 'no leaks found' on the current range, and the Gitleaks Secret Scan check is green on the current head.

PR #1125, opened 2026-07-02
feat(external-ingest): schema discovery + JSON Schema export (CHAOS-2692)

chrisgeo:
## Summary

Implements CHAOS-2692 (wave 2 of the CHAOS-2690 external customer-push ingestion epic): schema discovery and JSON Schema export for the frozen `external-ingest.v1` wire contract shipped by CHAOS-2691.

- **`schema_registry.py`**: builds a versioned JSON Schema bundle from CHAOS-2691's Pydantic models (`BatchEnvelope` + the 9 `RECORD_KIND_MODELS`) via `pydantic.json_schema.models_json_schema()` — never hand-written, never redeclared (CC17/D5). One document with `envelope`/`recordKinds` `$ref`s into a collision-free `$defs`, computed once per process (`@lru_cache`) and forced eagerly at router import time so a bad example or generation regression fails app startup, not a customer's first request.
- **`router.py`**: reworks `GET /schemas` + `GET /schemas/{schema_version}` to serve the registry's bundle, with ETag (`sha256` over the full served representation, not just the schema shape) and `If-None-Match` → `304`. Both routes stay public (D2) and rate-limited (`INGEST_READ_LIMIT`).
- **`examples/*.json`**: one canonical, realistic payload example per record kind — the epic's single fixture home (CC18), loaded via `importlib.resources`, attached to each `recordKinds` entry.
- **`export_schemas.py`**: static export mirroring `graphql/export_schema.py`'s pattern, checked into `docs/api/external-ingest/v1/schema.json`, enforced by a no-drift pytest.
- **`docs/architecture/adr-005-external-ingest-schema-discovery.md`**: records the versioning/caching/static-export decisions, including two adversarial-review rounds' findings and fixes (below).
- 21 new/updated tests across `tests/api/external_ingest/` + targeted updates to `tests/api/test_external_ingest_router.py` (schema response shape changed from inlined per-kind schemas to `$ref`-indexed).

## Adversarial review findings (2 rounds, all resolved)

1. **[HIGH]** `GET /schemas*`'s rate-limit key reused `get_ingest_token_key`, which hashes any `Authorization: Bearer <value>` header into its own bucket — but these routes never validate that header (public by design), so a caller could rotate a random bearer string per request to bypass `INGEST_READ_LIMIT` entirely. **Fixed**: a local IP-only key function scoped to just these two routes (no edit to `middleware/rate_limit.py`).
2. **[MEDIUM]** ETag was hashed over the schema document only; the dynamic `limits` field was appended outside the hash, so a `maxRecordsPerBatch`/`maxBodyBytes` env change could return a stale `304`. **Fixed**: the served ETag is now computed per-request over the full response body.
3. **[MEDIUM]** The exported envelope schema doesn't tie `records[].payload` to `kind` (that link is Python-only, in the frozen `RecordEnvelope` model) — a customer could validate a full batch against `envelope` alone and pass a payload the server's `/validate` rejects. **Addressed** via explicit documentation in the bundle's `description` + a locking regression test, per master-spec D9's already-scoped per-record-`$ref` validation path (a hand-rolled `oneOf`-by-kind schema would violate D5).
4. **[MEDIUM, round 2]** `schemaVersion`/`kind` are typed `str` (not `Literal`) in the frozen model, so the naive generated schema didn't constrain them — a customer could certify `schemaVersion: "external-ingest.v99"` or an unknown `kind` that the server rejects. **Fixed**: post-generation `const`/`enum` tightening in the registry, sourced from data it already owns (`SUPPORTED_SCHEMA_VERSIONS`, `RECORD_KIND_MODELS`), not hand-writing a schema.

## Test plan

- [x] `env -i ... SCRATCH_DB=ci_local_validate_2692 bash ci/local_validate.sh` — GATE PASSED (ruff format/check, mypy, full unit suite 6343 passed/44 skipped, live-ClickHouse argMax proof)
- [x] Codex adversarial review — 2 rounds, all findings resolved (see above)
- [x] Live verification (`uvicorn` on `127.0.0.1:8692`):
  - `GET /schemas` lists `external-ingest.v1` + all 9 record kinds + limits
  - `GET /schemas/external-ingest.v1` returns the bundled document with correct `ETag`/`Cache-Control` headers
  - `If-None-Match` round-trip → `304`; stale/unknown version → `404`
  - Restarting with a different `EXTERNAL_INGEST_MAX_RECORDS` confirms a stale client ETag no longer 304s (round-1 fix verified live)
  - Independent `jsonschema` validator (ephemeral install, not committed) cross-checked against `POST /validate`: all 9 examples validate; a known-bad `commit.v1` payload is rejected by both the per-kind `$ref` schema and `/validate` with matching diagnosis; an unsupported `schemaVersion`/unknown `kind` at the envelope level is now rejected by the independent validator (round-2 fix verified live)

### DIFF
--- a/docs/api/external-ingest/v1/schema.json
+++ b/docs/api/external-ingest/v1/schema.json
@@ -1,0 +1,1518 @@
+{
+  "$defs": {
+    "BatchEnvelope": {
+      "additionalProperties": false,
+      "properties": {
+        "idempotencyKey": {
+          "maxLength": 255,
+          "minLength": 1,
+          "title": "Idempotencykey",
+          "type": "string"
+        },
+        "records": {
+          "items": {
+            "$ref": "#/$defs/RecordEnvelope"
+          },
+          "minItems": 1,
+          "title": "Records",
+          "type": "array"
+        },
+        "schemaVersion": {
+          "const": "external-ingest.v1",
+          "title": "Schemaversion",
+          "type": "string"
+        },
+        "source": {
+          "$ref": "#/$defs/SourceDescriptor"
+        },
+        "window": {
+          "anyOf": [
+            {
+              "$ref": "#/$defs/IngestWindow"
+            },
+            {
+              "type": "null"
+            }
+          ],
+          "default": null
+        }
+      },
+      "required": [
+        "schemaVersion",
+        "idempotencyKey",
+        "source",
+        "records"
+      ],
+      "title": "BatchEnvelope",
+      "type": "object"
+    },
+    "CommitV1": {
+      "additionalProperties": false,
+      "properties": {
+        "authorEmail": {
+          "anyOf": [
+            {
+              "type": "string"
+            },
+            {
+              "type": "null"
+            }
+          ],
+          "default": null,
+          "title": "Authoremail"
+        },
+        "authorName": {
+          "anyOf": [
+            {
+              "type": "string"
+            },
+            {
+              "type": "null"
+            }
+          ],
+          "default": null,
+          "title": "Authorname"
+        },
+        "authorWhen": {
+          "format": "date-time",
+          "title": "Authorwhen",
+          "type": "string"
+        },
+        "committerEmail": {
+          "anyOf": [
+            {
+              "type": "string"
+            },
+            {
+              "type": "null"
+            }
+          ],
+          "default": null,
+          "title": "Committeremail"
+        },
+        "committerName": {
+          "anyOf": [
+            {
+              "type": "string"
+            },
+            {
+              "type": "null"
+            }
+          ],
+          "default": null,
+          "title": "Committername"
+        },
+        "committerWhen": {
+          "anyOf": [
+            {
+              "format": "date-time",
+              "type": "string"
+            },
+            {
+              "type": "null"
+            }
+          ],
+          "default": null,
+          "title": "Committerwhen"
+        },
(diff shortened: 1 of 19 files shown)


chrisgeo:
**Orchestrator verification record (pre-merge)** — CHAOS-2692. Local gate: GATE PASSED under env -i (6343 passed / 0 failed) after both codex rounds. Codex: round 1 — 1 HIGH fixed (public GET /schemas* limiter was bearer-rotation-bypassable; now IP-only key local to the two public routes, middleware untouched), 2 MEDIUM (ETag now hashed over full response body incl. live limits — stale-304 killed; envelope-vs-kind validation gap documented + regression-locked per master-spec D9/D5, ratified). Round 2 — 1 new MEDIUM fixed (exported schema accepted invented schemaVersion/kind; post-generation const/enum tightening from SUPPORTED_SCHEMA_VERSIONS/RECORD_KIND_MODELS, frozen schemas.py untouched). Line-level review: file list exactly in-scope; router changes limited to GET handlers + local key func; import-time bundle generation fails fast on bad fixtures. Live verification: ETag/304 round-trip incl. limits-change invalidation; independent jsonschema cross-check agrees with POST /validate on all 9 examples + a bad payload + the const/enum rejections. Merging under the approved integration-merge policy once CI settles.

PR #1127, opened 2026-07-02
feat(external-ingest): real ingest-token auth + credential lifecycle (CHAOS-2712)

chrisgeo:
## Summary

Replaces CHAOS-2691's interim, flag-gated `require_ingest_scope` body with real, DB-backed `IngestToken` authorization (master-spec CC14 — the epic's hard pre-GA blocker).

- Resolves `fcpush_` bearer tokens via `sha256` against `external_ingest_tokens`, builds `IngestAuthContext(org_id, scopes, token_id, source)`.
- 401 `invalid_token` on missing/malformed/unknown/revoked/expired; 403 `insufficient_scope` on missing scope.
- Last-used tracking (`last_used_at`/`last_used_ip`) in an isolated Postgres session, decoupled from the request's own session, so it survives regardless of what happens downstream.
- Ingest-auth failures are audited (`INGEST_AUTH_FAILED`) with commit-before-raise discipline — a regression test proves the audit row survives the 401/403 it documents.
- **Deletes `EXTERNAL_INGEST_INSECURE_AUTH` and the `X-Org-Id` header path entirely.**
- Fixes the admin `webhookMode` two-layer contract (adr-004): schema accepts the full 3-value enum (no 422 on `fullchaos_hosted`); router rejects it with 400 before persisting.

### Adversarial-review fixes (2 rounds, both closed)

1. **[high] Cross-source write hole**: `POST /batches` never checked a source-bound `ingest:write` token's source against the payload's declared source — only `org_id` was checked. Added `require_matching_source` (called from `router.py::accept_batch` after envelope parsing, since the auth dependency resolves before the body is available) → 403 `source_mismatch`/`source_disabled`.
2. **[medium→high] Unthrottled auth-failure floods**: FastAPI resolves `Depends()` (the auth dependency) before the route's `@limiter.limit(...)` decorator body runs, so a 401/403 never reaches the existing per-token limiter. Added a two-layer IP-keyed throttle in `auth.py`: an atomic `hit()`-based attempt ceiling (`INGEST_AUTH_ATTEMPT_IP_LIMIT=100/minute`, consumed before any DB work — closes a TOCTOU race the first-pass `test()`-then-`hit()` design had, since the DB lookup between those two calls is itself an `await` point) plus a stricter `test()`/`hit()` failure-only bucket (`INGEST_AUTH_FAILURE_IP_LIMIT=30/minute`) layered behind it.

Both fixes were verified with a live server + real minted tokens, and a concurrency regression test (`asyncio.gather` of 110 concurrent invalid-token requests) proves the number of DB lookups actually performed is bounded exactly at the attempt ceiling.

## Test plan

- [x] `tests/api/external_ingest/test_auth.py` (new): direct-dependency-style unit tests (401/403 matrix, audit persistence, last-used bump, org isolation, no-caching-on-revoke) + a handful of end-to-end httpx tests through the real router + the two-round adversarial-review regressions (source binding, concurrent-flood throttle).
- [x] `tests/api/test_ingest_rate_limit_key.py`: rewritten for the new validated-token-id keying (was raw-bearer-text hashing).
- [x] `tests/api/test_external_ingest_router.py`: removed obsolete interim-auth tests; added source-binding coverage; `TEST_CTX` updated to the final `IngestAuthContext` shape.
- [x] `tests/api/admin/test_customer_push.py`: fixed a pre-existing test asserting the *wrong* (422) behavior for `webhookMode=fullchaos_hosted`; added coverage for the corrected 400 two-layer behavior.
- [x] `env -i ... SCRATCH_DB=ci_local_validate_2712 bash ci/local_validate.sh` — green (lint, mypy, full unit suite 6348 tests, live-ClickHouse argMax proof).
- [x] Codex adversarial review — 2 rounds, both findings fixed and re-verified.
- [x] Live verification against a scratch Postgres DB + real minted `fcpush_` tokens: schema:read/ingest:write success paths, 401 matrix (missing/unknown/revoked/expired), 403 matrix (insufficient_scope, source_mismatch), audit-row persistence via `psql`, `last_used_at` tracking, revoke-takes-effect-on-next-request (no caching), `EXTERNAL_INGEST_INSECURE_AUTH=1` confirmed to have zero effect even when explicitly set on the server, and a live concurrent-burst curl flood confirming the atomic throttle.

Docs updated in the same changeset: `docs/architecture/customer-push-authz.md` (new sections on source binding + the two-layer throttle), `docs/architecture/adr-003-external-ingest-rest-boundary.md` and `docs/architecture/external-ingest-rest-contract.md` (marked the interim-auth sections resolved).

### DIFF
--- a/docs/architecture/adr-003-external-ingest-rest-boundary.md
+++ b/docs/architecture/adr-003-external-ingest-rest-boundary.md
@@ -114,21 +114,30 @@ matching rule alone to prevent a Linear collision.
 
 ## Decision 4: Interim auth is a mechanically-gated stopgap, not a real credential system
 
-CHAOS-2691 ships `api/external_ingest/auth.py::require_ingest_scope` with a
-body that accepts **any** bearer token + `X-Org-Id` header combination —
-solely so the REST contract is end-to-end testable before CHAOS-2696/2712
-land real token issuance and validation. This is gated, not silently
-insecure:
+**RESOLVED by CHAOS-2712** — `require_ingest_scope` now resolves real,
+DB-backed `IngestToken` rows (sha256-hashed `fcpush_` bearer tokens against
+`external_ingest_tokens`); the `EXTERNAL_INGEST_INSECURE_AUTH` flag and the
+`X-Org-Id` header path described below no longer exist in any form. This
+section is kept as the historical record of the interim state CHAOS-2691
+shipped and the release-blocker rationale that gated it; see
+[docs/architecture/customer-push-authz.md](customer-push-authz.md) for the
+real auth model.
 
-- The dependency **hard-fails `503 auth_not_configured`** unless the
-  deployment sets `EXTERNAL_INGEST_INSECURE_AUTH=1` — a flag intended only
+CHAOS-2691 originally shipped `api/external_ingest/auth.py::require_ingest_scope`
+with a body that accepted **any** bearer token + `X-Org-Id` header
+combination — solely so the REST contract was end-to-end testable before
+CHAOS-2696/2712 landed real token issuance and validation. This was gated,
+not silently insecure:
+
+- The dependency **hard-failed `503 auth_not_configured`** unless the
+  deployment set `EXTERNAL_INGEST_INSECURE_AUTH=1` — a flag intended only
   for local compose/CI, never a deployed environment.
-- Every request accepted under the flag logs a WARNING naming the org_id,
-  so interim-mode traffic is visible in ops before CHAOS-2696/2712 land.
-- **CHAOS-2696/2712 landing (real DB-backed `IngestToken` validation) is a
+- Every request accepted under the flag logged a WARNING naming the org_id,
+  so interim-mode traffic was visible in ops before CHAOS-2696/2712 landed.
+- **CHAOS-2696/2712 landing (real DB-backed `IngestToken` validation) was a
   hard pre-GA blocker** for this epic, not routine follow-up work. Merging
-  the `chaos-2690-external-ingest` integration branch to `main` remains
-  additionally gated on CHAOS-2712.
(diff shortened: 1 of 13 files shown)


chrisgeo:
**Orchestrator verification record (pre-merge)** — CHAOS-2712. Local gate: GATE PASSED under env -i (6348 passed / 0 failed). Codex adversarial review: 2 rounds, 2 HIGH both fixed — (1) source-bound ingest:write tokens were never matched against the payload source on POST /batches (org-only check); require_matching_source now enforces 403 source_mismatch/source_disabled at accept time; (2) pre-auth flood protection: round-2 caught that the round-1 test()-then-hit() pattern was non-atomic across the DB-lookup await point — replaced with an atomic hit()-first attempt ceiling (100/min/IP) consumed before ANY DB work, with a concurrency regression test bounding DB lookups at exactly the ceiling under 110 concurrent requests. Line-level review: interim EXTERNAL_INGEST_INSECURE_AUTH flag + X-Org-Id path fully deleted (org_id now UUID-by-construction from the token row, closing the wave-1 deferred item); commit-before-raise audit discipline with regression test; last_used bump isolated from request-txn rollback; IngestAuthContext changes additive (frozenset scopes, source), Depends call sites unchanged; webhookMode two-layer contract fixed per adr-004 (schema accepts 3-value enum, router 400s fullchaos_hosted). Live verification: real minted tokens against scratch PG — full 401/403 matrix, audit persistence via psql, revoke-on-next-request, dead-flag proof, live concurrent flood confirming the atomic throttle. Coordination note: the require_matching_source call site in accept_batch folds into CHAOS-2695's canonical accept sequence in wave 4. Merging under the approved integration-merge policy once CI settles; combined-state gate on the integration branch follows the merge (shared router.py with merged #1125).

PR #1134, opened 2026-07-02
feat(external-ingest): bounded recomputation planner + recompute status (CHAOS-2699)

chrisgeo:
## Summary

Bounded recomputation planner for customer-push ingestion (CHAOS-2690, master-spec CC21):

- **`external_ingest/recompute.py`**: `plan_recompute()` (pure) + `dispatch_recompute()` (Celery signature builder) + `schedule_or_coalesce()` (public primitives-only seam, per the synthesizer reconciliation on brief-2699-recompute.md — `RecomputeScope` stays internal). Per-repo `run_daily_metrics -> run_work_graph_build` chains capped at 25 repos / 14 days; ONE scoped `dispatch_investment_materialize_partitioned(force=False)` call per flush, never with both `repo_ids`/`team_ids` empty (D4 hard invariant); D8 org-wide day-bounded fallback for repo-less work-item batches; never reuses `dispatch_daily_metrics_partitioned`/`run_daily_metrics_batch` (D6, negative-space-tested); dispatch failures never raise (D13).
- **`external_ingest/recompute_status.py`**: direct-SQL status persistence (D11), sqlite-portable (no `ANY(:array)`/`RETURNING`, per-ingestion_id UPDATE loop instead).
- **`workers/external_ingest_recompute.py`**: `flush_external_ingest_recompute` Celery task, own module — see "Integrator wiring" below.
- **Migration 0034**: `down_revision="0033"` (chains onto CHAOS-2694's status-store migration). Adds 5 `recompute_*` columns to `external_ingest_batches` + new `external_ingest_recompute_jobs` table.
- **`api/external_ingest/status.py`**: extends `GET /batches/{ingestionId}` with the `recompute` response block (pinned enum: `not_applicable | pending | dispatched | skipped_no_scope | failed`).
- Architecture doc: `docs/architecture/external-ingest-bounded-recompute.md` (D1-D15).

### Fixed after two rounds of Codex adversarial review

1. **Atomic debounce merge** — `schedule_or_coalesce()`'s pending-blob read-merge-write + guard SETNX now runs inside a Valkey `WATCH`/`MULTI` optimistic transaction (was a racy plain `GET`→`SET` that could silently drop one caller's scope under concurrency). Covered by `test_concurrent_callers_do_not_lose_either_ingestion_id`.
2. **Atomic flush read** — the flush task now reads the pending blob via `GETDEL` (one atomic command) instead of separate `GET`+`DELETE`, and no longer explicitly deletes the guard key (its own TTL governs it) — closes a window where a late-arriving newer debounce window could be silently erased by an older flush's cleanup.
3. **Nullable `celery_task_id`** — a per-repo daily-metrics job's captured Celery task id can legitimately be `None` (chain `AsyncResult.parent` unset); the column was `NOT NULL`, so that path raised an `IntegrityError` that silently rolled back the *entire* status/job-log write even though the Celery jobs had already dispatched. Column is now nullable in both the migration and the ORM model.

Two findings were evidence-refuted (documented in the architecture doc, not fixed in this PR):
- Flush task not registered by a fresh worker — deliberate, CC20 pins Celery wiring (including which modules a worker imports) as CHAOS-2693's sole responsibility this wave; the exact one-line fix is in the `INTEGRATOR TODO` comment at the top of `workers/external_ingest_recompute.py`.
- No production caller wires `schedule_or_coalesce()` into the worker — explicitly CHAOS-2697's job per the brief's dependency direction (2697 depends on this PR's seam, not the reverse); 2697 hasn't landed yet.

## Integrator wiring (for CHAOS-2693 / whoever merges this wave)

Add exactly this one line to `workers/config.py`'s `late_ack_excluded_tasks` tuple:

```python
"dev_health_ops.workers.tasks.flush_external_ingest_recompute",
```

No `workers/tasks.py` re-export needed (task name is pinned via the decorator's `name=` kwarg). No `task_queues`/compose change needed — it reuses the existing `default` queue's worker coverage (`queue="default"` already on the task decorator).

## Test plan

- [x] `SCRATCH_DB=ci_local_validate_2699 bash ci/local_validate.sh` — PASSED (lint, mypy, full unit suite 6494 passed, live-ClickHouse argMax proof), env-isolated (`env -i`).
- [x] 71 new unit tests across 6 files: planner (pure), dispatch (+ D14 Celery kwargs-contract via `inspect.signature`, D6 negative-space grep, `parent=None` chain-result path), debounce (fakeredis, incl. concurrent-callers race), flush task (GETDEL, guard preservation), status-SQL persistence (incl. `task_id=None`), migration 0034 idempotent up/down/up.
- [x] Migration 0034 up/down/up proof against a real scratch Postgres DB (`ci_2699`, not `devhealth`) — `psql \d` confirmed columns/table both directions.
- [x] Debounce coalescing + SETNX guard expiry/re-arm proven against real dev Valkey with a throwaway test org (Celery `apply_async` mocked to avoid firing real dispatches at live workers running in dev compose).
- [x] Dispatch shapes (25-repo/14-day caps, empty-scope investment refusal, D8 fallback) proven with real `plan_recompute`/`dispatch_recompute` logic, Celery signature/chain/send_task captured rather than fired.
- [x] `GET /batches/{ingestionId}` proven end-to-end via a real `uvicorn` process against scratch Postgres (auth dependency overridden — CHAOS-2696's token provisioning is out of scope for this check) — confirmed the `recompute` block renders with the pinned enum, scope, and job list.
- [x] No `postgres` pytest marker added (epic-wide convention, per CHAOS-2694's sqlite-portable direct-SQL precedent).
- [x] `tests/test_compose_config.py` stays green untouched (no queue/compose changes).

### DIFF
--- a/docs/architecture/external-ingest-bounded-recompute.md
+++ b/docs/architecture/external-ingest-bounded-recompute.md
@@ -1,0 +1,404 @@
+# Bounded recomputation planner (CHAOS-2699)
+
+Part of the [customer-push ingestion epic](adr-003-external-ingest-rest-boundary.md)
+(CHAOS-2690). A customer-pushed batch that lands new git/work-item facts
+must trigger the *existing* metrics/work-graph/investment pipeline for the
+affected slice -- but never a full-org recompute, and never so eagerly that
+a burst of small batches fires the pipeline once per batch. This doc
+records the D1-D15 design decisions behind
+`src/dev_health_ops/external_ingest/recompute.py`,
+`recompute_status.py`, and `workers/external_ingest_recompute.py`.
+
+## The planner is a free function, not a class (D1)
+
+Mirrors the documented house preference (`workers/queues.py` -- the
+`SyncDispatchPolicy` class was designed but deliberately never built) and
+the working analog `_dispatch_post_sync_tasks`
+(`workers/post_sync_dispatch.py`). `plan_recompute()` is pure (scope in,
+plan out, no I/O); `dispatch_recompute()` is the impure Celery-signature
+builder; `schedule_or_coalesce()` is the public debounce seam.
+
+## Public seam is primitives, not a shared dataclass (CC21)
+
+`RecomputeScope` is internal to `recompute.py`. The function every other
+sub-issue calls is:
+
+```python
+schedule_or_coalesce(
+    *, org_id, source_system, source_instance, ingestion_id,
+    repo_ids: set[str], team_ids: set[str],
+    window_start: datetime | None, window_end: datetime | None,
+    record_kinds: set[str],
+)
+```
+
+This decouples CHAOS-2699 from CHAOS-2697/2698's own scope-accumulation
+data structures -- 2697 maps whatever `AffectedScope` it builds while
+normalizing a batch into these kwargs at the end of processing. No sibling
+issue imports `RecomputeScope` or calls `RecomputeScope.merge_record()`.
+
+## Debounce via Valkey SETNX, not a Celery beat poll (D3)
+
+Debounce here needs "coalesce writes that land within N seconds, fire
+once" -- a SETNX-guarded scheduled flush needs no new beat entry, unlike
+the existing 30s-cadence stream-consumer beat pattern (which drains a
+*stream*, a different shape of problem). Two keys per
+`(org_id, source_system, source_instance)`:
+
+- `external-ingest:recompute:pending:{org}:{system}:{instance}` -- the
+  accumulating scope blob (JSON), widened (union of repo/team/kind sets,
+  min/max window) on every call within the debounce window.
+- `external-ingest:recompute:scheduled:{org}:{system}:{instance}` -- a
+  SETNX guard, TTL == the debounce window. Only the caller that acquires it
+  schedules the flush; every other caller within the window just widens the
+  pending blob.
+
+Client is a plain non-blocking Valkey client (`_get_redis_client()`, same
+shape as `api/product_telemetry/streams.py:get_redis_client()`) -- **not**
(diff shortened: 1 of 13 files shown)


chrisgeo:
**Orchestrator verification record (pre-merge)** — CHAOS-2699. Local gate: GATE PASSED under env -i (mypy + full suite). Codex adversarial review: 2 rounds — R1: 2 HIGH concurrency races fixed and regression-tested (schedule_or_coalesce GET-merge-SET overwrite → Valkey WATCH/MULTI optimistic txn w/ retry, proven by a two-client fakeredis race test; flush GET-then-DELETE window losing a fresher pending blob → atomic GETDEL); R2 re-verified both fixes and surfaced 1 more HIGH fixed (NOT NULL celery_task_id caused IntegrityError that silently rolled back the entire status/job-log write when a chain parent id was legitimately None — batch stranded at not_applicable; now nullable in migration + model, tested end-to-end). Two findings correctly evidence-refuted (production caller lands with CHAOS-2697 per dependency direction; worker wiring is CHAOS-2693's file ownership this wave). Line-level review: file list exactly in-scope — no workers/config.py, no tasks.py, no compose (hot-file directive honored); flush task registered in its own module via decorator name= kwarg on the existing default queue; status.py extension uses the pinned enum. Live verification: migration 0034 up/down/up on scratch PG; live valkey debounce coalescing + expiry re-arm on a test org; celery dispatch shapes verified against inspect.signature (kwargs-drift guard); GET batch detail shows the recompute block. Integration note: the single late_ack_excluded_tasks line is being routed into CHAOS-2693's config.py changeset (its file ownership) rather than hand-applied at merge. Merging under approved policy once CI settles.

PR #1154, opened 2026-07-03
fix(webhooks): re-raise RateLimitException to the Celery retry path (CHAOS-2829)

chrisgeo:
## Summary

Follow-up to CHAOS-2803 (#1151). GitHub/GitLab/Jira webhook handlers caught `RateLimitException` in the generic `except Exception` clause and returned `{"processed": False}`. The outer task treats that as a permanent failure and does not retry -- and since the provider will not resend the webhook, a rate-limited sync was silently and permanently dropped.

- Import `RateLimitException` and add a dedicated `except RateLimitException: log + raise` block before the generic handler in all three webhook processors (`_process_github_event`, `_process_gitlab_event`, `_process_jira_event`), so the exception escapes to the outer task's Celery retry path instead of being swallowed.
- The outer `process_webhook_event` retry now honors a provider-supplied `retry_after_seconds` on `RateLimitException` for the retry countdown, falling back to the existing exponential backoff when not present.
- Added `tests/workers/test_system_webhooks_rate_limit.py` (6 tests, all passing) covering re-raise behavior for all three handlers and the countdown logic.

## Test plan

- [x] `pytest tests/workers/test_system_webhooks_rate_limit.py -q` -- 6 passed
- [x] ruff + mypy clean on both changed files (pre-commit/pre-push hooks passed)
- [ ] Online CI (this repo's local gate runs offline/no-network; one unrelated pre-existing failure noted below is expected to be green in CI)

## Note on gate status

The local gate has exactly one unrelated pre-existing failure: `tests/test_admin_credentials.py::test_gitlab_connection_helper_sanitizes_raw_response_body` (`KeyError: 'status'`). This diff touches only `src/dev_health_ops/workers/system_webhooks.py` (plus the new test file) -- it does not touch admin credentials code. Confirmed pre-existing by running the same test against a clean `origin/main` checkout, where it also fails. Root cause: `_test_gitlab_connection`'s SSRF DNS guard (`socket.getaddrinfo`) is not mocked by that test, so it fails offline/no-network before ever reaching the mocked `httpx.AsyncClient` -- this is the known offline-gate/live-DNS pattern. Filed as CHAOS-2830 (follow-up to CHAOS-2780) and being fixed directly in a separate PR.

### DIFF
--- a/src/dev_health_ops/workers/system_webhooks.py
+++ b/src/dev_health_ops/workers/system_webhooks.py
@@ -4,6 +4,7 @@ import logging
 import os
 from datetime import datetime, timezone
 
+from dev_health_ops.exceptions import RateLimitException
 from dev_health_ops.utils.datetime import utc_today
 from dev_health_ops.workers.async_runner import run_async
 from dev_health_ops.workers.celery_app import celery_app
@@ -99,7 +100,16 @@ def process_webhook_event(
             event_type,
             exc,
         )
-        raise self.retry(exc=exc, countdown=30 * (2**self.request.retries))
+        # Honor a provider-supplied rate-limit delay when the escaping exception
+        # carries one (RateLimitException.retry_after_seconds); otherwise fall
+        # back to the generic exponential backoff.
+        retry_after = getattr(exc, "retry_after_seconds", None)
+        countdown = (
+            int(retry_after)
+            if isinstance(retry_after, (int, float)) and retry_after > 0
+            else 30 * (2**self.request.retries)
+        )
+        raise self.retry(exc=exc, countdown=countdown)
 
 
 def _is_duplicate_delivery(provider: str, delivery_id: str) -> bool:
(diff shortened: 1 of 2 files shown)


chrisgeo:
## Codex adversarial gate — 3 rounds, core converged

Lead ran `codex-companion adversarial-review` against `origin/main` three times:

1. **medium** — non-finite `retry_after_seconds` (`Retry-After: inf` → `float('inf')`) made `int(retry_after)` raise `OverflowError` out of the retry handler, dropping the webhook. **Fixed** (`math.isfinite` guard + 1h clamp + `ceil`).
2. **medium** — `math.isfinite(10**400)` itself raises `OverflowError` (huge int can't coerce to float). **Fixed** (int and float handled separately; ints clamped via pure-int `min`, never coerced).
3. **medium (deferred)** — a *distinct, deeper* swallow path: the shared `process_*_repo` backfill blocks (`processors/github.py:1848`, `processors/gitlab.py:~2040/~2395`) catch `RateLimitException` in a broad `except Exception`, so a push webhook records success while silently dropping rate-limited backfill.

Round-3 is **deferred to [CHAOS-2831](https://linear.app/fullchaos/issue/CHAOS-2831/webhook-push-backfill-still-swallows-ratelimitexception-in-shared)**, not fixed here: `process_*_repo` is caller-agnostic (webhooks, scheduled sync, CLI onboarding), so a re-raise there changes behavior for every caller and interacts with the CHAOS-2742 budget/defer machinery — it needs per-caller semantics + webhook-level integration tests, a larger blast radius than this focused worker-handler fix. This PR delivers the converged core: the webhook handlers and the outer retry countdown no longer swallow rate limits.

PR #1185, opened 2026-07-04
fix(sync): enable blame dataset availability

chrisgeo:
## Summary
- Seed the heavy `blame` dataset for planner-managed GitHub/GitLab integrations whenever the legacy `git` target is selected, without changing `sync/datasets.py` registry/flag semantics.
- Raise GitHub per-file blame fetch failures from debug to bounded warnings: first 5 failures per repo, then a per-repo summary count.
- Document the default seeding + existing-fixture enable path and fix `ci/local_validate.sh` so the documented `SCRATCH_DB` override works.

## Decision evidence
- `sync/datasets.py` already classifies `DatasetKey.BLAME` as supported for GitHub/GitLab and heavy, with processor flags `{blame_only, sync_blame}`.
- Planner runs only enabled `integration_datasets` rows, so default `git` sync configs without a `blame` row could never plan blame units.
- GitHub blame backfill is bounded by `BLAME_BACKFILL_MAX_FILES=500` and coverage-aware resumption, so omission was treated as a seed/default gap rather than an intentional permanent exclusion.

## Local dev data
- Before: local org `70d529e0-3c06-4597-8480-794fd02328b6` had 84 `integration_datasets` rows and 0 `blame` rows.
- Path used: targeted Postgres INSERT in container `dev-health-postgres-1` / db `devhealth` / user `devhealth`, after inspecting `integration_datasets` columns. Insert selected active GitHub integrations with an existing `files` dataset and copied that row's `options` into new `dataset_key='blame'`, `is_enabled=true` rows.
- After: 5 GitHub `blame` rows inserted; local org now has 89 dataset rows and 5 blame rows. This was dev-data only; production rollout is out of scope.

## Validation
- LSP diagnostics: clean on changed Python/test files.
- `SCRATCH_DB=ci_validate_chaos2862 bash ci/local_validate.sh` — PASS.
  - ruff format/check: PASS
  - mypy: PASS
  - full unit suite: 6890 passed, 44 skipped
  - argMax live-exec proof: PASS

## Review
- Explore-agent review: PASS, no blockers; forbidden files untouched.

SCREENSHOT-WAIVER: backend/docs/CI-only changes; no rendered UI changed.

### DIFF
--- a/ci/local_validate.sh
+++ b/ci/local_validate.sh
@@ -229,13 +229,14 @@ ch_argmax_proof() {
   # resolver EXPLAIN, RMT-dedup-live) needs a SEEDED ClickHouse and is NOT part of this
   # gate (CI does not run it either). To run it by hand: create a scratch db, point
   # CLICKHOUSE_URI at it, `dev-hops fixtures generate`, then `pytest -m clickhouse`.
-  CLICKHOUSE_URI="${SCRATCH_URI}" DATABASE_URI="${SCRATCH_URI}" OTEL_ENABLED=false PYTHONPATH=src \
+  CLICKHOUSE_URI="${SCRATCH_URI}" DATABASE_URI="${SCRATCH_URI}" SCRATCH_DB="${SCRATCH_DB}" OTEL_ENABLED=false PYTHONPATH=src \
     "${PROXY_OFF[@]}" "${PYBIN}" - <<'PYEOF'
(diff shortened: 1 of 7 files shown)


chrisgeo:
Note: the local validation gate was validated with the scratch-db override fix, but that ci/local_validate.sh fix is intentionally removed from this PR and now rides in #1183.

PR #1196, opened 2026-07-05
fix(sync): terminalize orphaned backfill jobs in reconciler (CHAOS-2868)

chrisgeo:
Fixes CHAOS-2868

## Zombie-banner mechanism

The admin sync page's "BACKFILL IN PROGRESS / Dispatching work... / 0% / Live — refreshing" banner renders whenever `GET /backfill-jobs` finds a `BackfillJob` whose merged status is non-terminal. `_backfill_job_run_counts` (`src/dev_health_ops/api/admin/routers/sync.py:570`) merges the linked `SyncRun` status over the stored `BackfillJob.status`, but only when `_backfill_job_sync_run_id` (`:562`) can extract and resolve a `sync_run:<uuid>` marker from `celery_task_id`. When it can't — because the marker is missing/unparseable, or it resolves to a `SyncRun` row that no longer exists — the endpoint falls back to the stored `status` (`pending`/`running`) **forever**, since nothing else ever mutates that column again.

Two repair paths already exist for the recoverable cases:
- `_terminal_runs_with_stale_observers` (`workers/sync_reconciler.py`) + `sync_observers_for_terminal_sync_run` (`workers/sync_units.py:2061`) repair pending/running `BackfillJob` rows whose marker resolves to a **terminal** `SyncRun`.
- Post-CHAOS-2867 (#1195), stranded runs get re-dispatched, hard-denied, and terminalized, after which the observer repair above fires.

Neither path helps a job whose marker can never resolve to any run. This PR adds a periodic sweep in `reconcile_sync_dispatch` that terminalizes those true orphans:

- **Missing/unparseable marker**: `_backfill_job_sync_run_id(job) is None`.
- **Marker resolves to a deleted run**: the parsed `sync_run:<uuid>` no longer has a matching `SyncRun` row.

Jobs whose marker resolves to an **existing** run (terminal or not) are left untouched — those are owned by the observer-repair pass above or the live dispatch/finalize flow.

## Change

- `_backfill_job_orphan_ttl_seconds()` — env-tunable via `SYNC_BACKFILL_JOB_ORPHAN_TTL_SECONDS` (default 3600s), following the existing `_rate_limit_observation_retention_days` int-parse pattern.
- `_backfill_job_is_orphaned(session, job)` — the two orphan predicates above.
- `_terminalize_orphaned_backfill_jobs(session, now, limit)` — scans pending/running `BackfillJob` rows older than the TTL, sets `status="failed"`, `error_message="backfill job orphaned: no linked sync run"`, `completed_at=now` for true orphans. Wired into `reconcile_sync_dispatch` right after the existing observer-repair pass, counted as `orphaned_backfill_jobs` in the task's returned dict.

## Tests

`tests/test_sync_reconciler.py` (5 new cases, following the existing fixture style):
1. Orphaned job past TTL with no marker → terminalized failed.
2. Orphaned job with marker pointing at a deleted run → terminalized failed.
3. Young orphan within TTL → left untouched.
4. Pending job with a resolvable marker to a live non-terminal run → left untouched (existing dispatch/finalize flow owns it).
5. Regression: pending job with a terminal run is still repaired by the existing observer path (unaffected by this change).

## Verification

```
.venv/bin/ruff format --check .
.venv/bin/ruff check .
.venv/bin/mypy src/dev_health_ops/workers/sync_reconciler.py
.venv/bin/python -m pytest tests/test_sync_reconciler.py tests/test_sync_units.py -q
```

All green (116 passed).

Scope is intentionally minimal — one repair loop + tests, no refactor of surrounding reconciler code. Web-side staleness-cutoff fix (`getActiveBackfillJob`, `BackfillStatus`) is tracked separately per the issue's proposed fix.

### DIFF
--- a/src/dev_health_ops/workers/sync_reconciler.py
+++ b/src/dev_health_ops/workers/sync_reconciler.py
@@ -651,7 +651,19 @@ def _backfill_job_sync_run_id(job: BackfillJob) -> str | None:
     return task_id.rsplit(marker, 1)[-1] or None
 
 
-def _backfill_job_is_orphaned(session, job: BackfillJob) -> bool:
+def _backfill_job_marker_sync_run_id(job: BackfillJob) -> uuid.UUID | None:
+    sync_run_id = _backfill_job_sync_run_id(job)
+    if sync_run_id is None:
+        return None
+    try:
+        return uuid.UUID(sync_run_id)
+    except ValueError:
+        return None
+
+
+def _backfill_job_is_orphaned(
+    sync_run_id: uuid.UUID | None, existing_run_ids: set[uuid.UUID]
+) -> bool:
     # Mirrors the admin surface's fallback in api/admin/routers/sync.py
     # (_backfill_job_sync_run_id / _backfill_job_run_counts): a job whose
     # marker is missing/unparseable, or whose marker resolves to a SyncRun
@@ -660,38 +672,44 @@ def _backfill_job_is_orphaned(session, job: BackfillJob) -> bool:
     # forever (CHAOS-2868). A marker resolving to an EXISTING run (terminal
     # or not) is owned by the observer-repair pass above or the live
     # dispatch/finalize flow and must not be touched here.
-    sync_run_id = _backfill_job_sync_run_id(job)
-    if sync_run_id is None:
-        return True
-    try:
-        run_uuid = uuid.UUID(sync_run_id)
-    except ValueError:
-        return True
-    return session.get(SyncRun, run_uuid) is None
+    return sync_run_id is None or sync_run_id not in existing_run_ids
 
 
 def _terminalize_orphaned_backfill_jobs(session, now: datetime, limit: int) -> int:
     cutoff = now - timedelta(seconds=_backfill_job_orphan_ttl_seconds())
     max_repairs = max(1, int(limit))
-    terminalized = 0
     candidates = (
         session.query(BackfillJob)
-        .filter(BackfillJob.status.in_({"pending", "running"}))
+        .filter(
+            BackfillJob.status.in_({"pending", "running"}),
+            BackfillJob.created_at <= cutoff,
+        )
         .order_by(BackfillJob.created_at.asc(), BackfillJob.id.asc())
+        .limit(max_repairs)
         .all()
     )
+    if not candidates:
+        return 0
+
+    marker_run_ids = {
+        job.id: _backfill_job_marker_sync_run_id(job) for job in candidates
+    }
+    candidate_run_ids = {
+        run_id for run_id in marker_run_ids.values() if run_id is not None
+    }
+    existing_run_ids: set[uuid.UUID] = set()
+    if candidate_run_ids:
+        rows = session.query(SyncRun.id).filter(SyncRun.id.in_(candidate_run_ids)).all()
+        existing_run_ids = {row_id for (row_id,) in rows}
+
+    terminalized = 0
     for job in candidates:
-        created_at = job.created_at
-        if created_at is None or _as_aware(created_at) > cutoff:
-            continue
-        if not _backfill_job_is_orphaned(session, job):
+        if not _backfill_job_is_orphaned(marker_run_ids[job.id], existing_run_ids):
             continue
         job.status = "failed"
         job.error_message = "backfill job orphaned: no linked sync run"
         job.completed_at = now
         terminalized += 1
-        if terminalized >= max_repairs:
-            break
     return terminalized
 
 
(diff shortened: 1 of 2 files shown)


chrisgeo:
Addressed both review findings in 55c4c66e3:

**Finding 1 (scan efficiency)** — `_terminalize_orphaned_backfill_jobs` now:
- Pushes the TTL age gate into the query (`BackfillJob.created_at <= cutoff`) plus `.limit(max_repairs)`, so young rows are never loaded and the scan is capped consistently with the file's other bounded queries.
- Replaces the per-row `session.get(SyncRun, ...)` lookup with a single bulk `SELECT SyncRun.id WHERE id IN (...)` over all parsed markers, eliminating the N+1 pattern.
- Split the orphan predicate into two pure, session-free helpers: `_backfill_job_marker_sync_run_id` (marker parsing) and `_backfill_job_is_orphaned` (set-membership decision against the bulk-fetched existing-run-id set).

Datetime comparison follows the file's existing convention (direct column `<=` comparison against an aware Python datetime, same as `SyncRunUnit.lease_expires_at <= now` elsewhere in this module) — verified against both the in-memory sqlite test suite and manual inspection of round-tripped values.

**Finding 2 (test precision)** — added 5 new tests:
- `test_backfill_job_marker_sync_run_id_parses_and_rejects_marker_variants` — pure unit test of marker parsing: resolvable UUID, missing marker text, `None` task id, and `sync_run:not-a-uuid` (unparseable → `None`).
- `test_backfill_job_is_orphaned_predicate_against_existing_run_id_set` — pure unit test of the set-membership predicate (no DB).
- `test_terminalize_orphaned_backfill_jobs_boundary_created_at_equals_cutoff` — `created_at` exactly `now - ttl` with a fixed `now` → terminalized (documents the inclusive boundary).
- `test_terminalize_orphaned_backfill_jobs_just_inside_ttl_is_untouched` — `created_at` one second younger than cutoff → left untouched.
- `test_terminalize_orphaned_backfill_jobs_unparseable_marker_is_orphaned` — `celery_task_id="worker|sync_run:not-a-uuid"` past TTL → terminalized.

The existing 5 integration-level reconciler tests (going through the full `reconcile_sync_dispatch` task) are unchanged and still pass — behavior is preserved, only the internals were made bulk/testable.

Gates, all green:
```
.venv/bin/ruff format --check .
.venv/bin/ruff check .
.venv/bin/mypy src/dev_health_ops/workers/sync_reconciler.py
.venv/bin/python -m pytest tests/test_sync_reconciler.py tests/test_sync_units.py -q
```
121 passed (30 in `test_sync_reconciler.py`, up from 25).

chrisgeo:
Fixed the MAJOR finding from the re-review in f69cca975: orphan starvation behind the candidate limit.

**Root cause**: the previous refactor applied `.limit(max_repairs)` to the candidate query *before* the orphan predicate was evaluated. If the oldest `limit` pending/running jobs all had markers resolving to existing SyncRuns (non-orphans), a true orphan sitting behind them in the `created_at` ordering was never loaded — and every subsequent reconciler run would re-select the exact same non-orphan window, starving that orphan permanently.

**Fix**: decoupled the scan budget from the repair budget.
- The candidate query now loads oldest-first up to `limit * _BACKFILL_JOB_ORPHAN_SCAN_LIMIT_MULTIPLIER` (5x) rows — still a single query, still followed by a single bulk `SyncRun.id IN (...)` existence check (no N+1 reintroduced).
- Terminalization then proceeds in `created_at` order until the repair budget (`limit`) is reached, so an orphan within the wider scan window is no longer starved by leading non-orphans.
- This keeps a pathological table (all non-orphans) bounded to one sweep of `limit * 5` rows rather than unbounded, with any remainder picked up on a later run once leading jobs terminalize or age out.
- Went with a single widened query rather than a keyset-paginated loop: paging with OFFSET while mutating rows in the same filtered set (status flips from pending/running to failed mid-loop) shifts the OFFSET math and can skip unscanned rows — a single bounded query sidesteps that correctness pitfall entirely and is simpler, matching the file's existing single-bulk-query style.
- Documented the budget semantics in the function's docstring.

**Regression test added**: `test_terminalize_orphaned_backfill_jobs_does_not_starve_orphan_behind_limit` — `limit=1`, an old non-orphan (marker resolves to a live, non-terminal `SyncRun`) ordered first, and a true orphan (no marker) immediately behind it. Asserts the orphan is terminalized in the same sweep while the non-orphan is left untouched.

Gates, all green:
```
.venv/bin/ruff format --check .
.venv/bin/ruff check .
.venv/bin/mypy src/dev_health_ops/workers/sync_reconciler.py
.venv/bin/python -m pytest tests/test_sync_reconciler.py tests/test_sync_units.py -q
```
122 passed (31 in `test_sync_reconciler.py`, up from 30).

No other files touched.

PR #1197, opened 2026-07-05
fix(coverage): make backfill requested ranges pair-aware (CHAOS-2869)

chrisgeo:
Fixes CHAOS-2869

## Problem

`_backfill_requested_ranges` in `src/dev_health_ops/api/services/sync_coverage.py` included every `BackfillJob` within the lookback window (no status filter) and expanded each job's date range across **every** `(source_id, dataset_key)` pair in scope. Covered ranges only count `SUCCESS` unit windows per exact pair. Any in-scope pair the planner never planned units for (unsupported datasets, sources added after the backfill ran, work-item family composite keys) became a permanent gap that no backfill could ever clear.

### Minimal repro (from the issue)

| Backfill job | Sources in scope | Dataset | Units actually planned | Result before fix |
| --- | --- | --- | --- | --- |
| `[Jan1 -> Jan3]` | `{A, B}` | `commits` | `SUCCESS` only for `(B, commits)` | `(A, commits)` stays gapped forever, even though the backfill "completed" |

## Fix

`_backfill_requested_ranges` now resolves each job's linked `SyncRun` via the `sync_run:<uuid>` suffix of `celery_task_id` (mirrors the parsing helper in `api/admin/routers/sync.py:_backfill_job_sync_run_id` / `workers/sync_reconciler.py`, duplicated locally to avoid a router import cycle). The job's requested interval is scoped to only the distinct `(source_id, dataset_key)` pairs that run's `SyncRunUnit` rows actually cover.

**Fallback:** if the marker is absent/unparseable, the linked run row is gone, or the run has zero units, the job falls back to the legacy all-pairs-in-scope behavior (documented in code). Such legacy rows age out of the 180-day lookback naturally.

`CoverageInterval` gains an optional `dataset_keys` field (default `()`, meaning "every dataset key in scope") so `build_coverage_summary_payload` can split backfill-requested intervals per exact pair. Public signatures of `build_coverage_summary_payload` / `build_sync_coverage_summary` are unchanged, and the payload's JSON shape (`datasets[].requested_ranges/covered_ranges/gaps/failed_ranges`) is unchanged.

## Tests

- `tests/test_sync_coverage.py`:
  - `test_backfill_interval_pair_scope_excludes_untouched_pairs` — pair-scoped backfill interval doesn't leak a gap onto an untouched pair.
  - `test_backfill_interval_without_pair_scope_applies_to_all_scope_pairs` — legacy/unresolved-marker fallback still spreads across all scope pairs.
- `tests/api/admin/test_sync_coverage_api.py` (full DB + marker resolution):
  - `test_sync_coverage_api_backfill_pair_scoped_to_planned_run_units` — core repro: a completed backfill whose run planned units only for `(source A, commits)` does not create a permanent gap on `(source B, commits)` or an unrelated in-scope dataset (`pull_requests`).
  - `test_sync_coverage_api_success_matching_backfill_range_clears_gap` — a pair-scoped backfill whose linked run's `SUCCESS` unit fully covers the requested range resolves to healthy/no-gap.
  - `test_sync_coverage_api_unresolvable_backfill_marker_falls_back_to_all_pairs` — an unparseable `sync_run:` marker falls back to the documented all-pairs-in-scope behavior.

## Verification

```
.venv/bin/ruff format --check .
.venv/bin/ruff check .
.venv/bin/mypy src/dev_health_ops/api/services/sync_coverage.py
.venv/bin/python -m pytest tests/test_sync_coverage.py tests/api/admin/test_sync_coverage_api.py -q
```
All green (28 passed). Also passed as part of the `pre-commit`/`pre-push` lefthook gates (ruff format/check + mypy across the full source tree).

## Scope notes

- Docs: grepped `ops/docs/` for coverage-semantics documentation; found no dedicated page describing the backfill-requested-range contract, so no doc update was needed.
- Out of scope (per issue): the requested/covered lookback-clock mismatch (`BackfillJob.created_at` vs `SyncRunUnit.updated_at`) is a known follow-up, not touched here.
- Does not touch `workers/sync_reconciler.py` or `web/`.

### DIFF
--- a/src/dev_health_ops/api/services/sync_coverage.py
+++ b/src/dev_health_ops/api/services/sync_coverage.py
@@ -501,22 +501,66 @@ def _backfill_job_sync_run_id(job: BackfillJob) -> str | None:
     return task_id.rsplit(marker, 1)[-1] or None
 
 
-async def _backfill_job_run_pairs(
+async def _backfill_job_run_pair_windows(
     session: AsyncSession, org_id: str, run_id: uuid.UUID
-) -> set[tuple[str, str]]:
-    """Return the distinct (source_id, dataset_key) pairs a SyncRun planned units for."""
-    stmt = (
-        select(SyncRunUnit.source_id, SyncRunUnit.dataset_key)
-        .where(
-            SyncRunUnit.org_id == org_id,
-            SyncRunUnit.sync_run_id == run_id,
-        )
-        .distinct()
+) -> dict[tuple[str, str], list[CoverageInterval]]:
+    """Return each (source_id, dataset_key) pair's merged unit-window union for a SyncRun.
+
+    An empty dict means the run has zero (valid) SyncRunUnit rows -- callers
+    MUST treat that as "this job requested nothing", not as "unresolvable"
+    (see ``_backfill_requested_ranges``): a run that legitimately planned zero
+    units must not be conflated with a marker we simply couldn't parse.
+    """
+    stmt = select(SyncRunUnit).where(
+        SyncRunUnit.org_id == org_id,
+        SyncRunUnit.sync_run_id == run_id,
+        SyncRunUnit.since_at.is_not(None),
+        SyncRunUnit.before_at.is_not(None),
     )
-    return {
-        (str(source_id), str(dataset_key))
-        for source_id, dataset_key in (await session.execute(stmt)).all()
-    }
+    raw_windows: dict[tuple[str, str], list[CoverageInterval]] = defaultdict(list)
+    for unit in (await session.execute(stmt)).scalars().all():
+        since_at = unit.since_at
+        before_at = unit.before_at
+        if since_at is None or before_at is None:
+            continue
+        pair = (str(unit.source_id), str(unit.dataset_key))
+        raw_windows[pair].append(
+            CoverageInterval(since=ensure_utc(since_at), before=ensure_utc(before_at))
+        )
+    return {pair: merge_intervals(intervals) for pair, intervals in raw_windows.items()}
+
+
+async def _resolve_backfill_job_pair_windows(
+    session: AsyncSession, org_id: str, job: BackfillJob
+) -> dict[tuple[str, str], list[CoverageInterval]] | None:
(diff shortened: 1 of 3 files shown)


chrisgeo:
Addressed both review findings in c3621609c.

**BLOCKER — phantom gap wider than the run's actual units.** Fixed: `_backfill_requested_ranges` no longer applies the job's full date range to a resolved pair. It now pulls that pair's actual `SyncRunUnit` windows (merged), then intersects (`_clip_intervals`) each with the job's declared date range. A job spanning wider than what its run actually planned (e.g. Jan1-Jan31 job, units only through Jan28) can no longer contribute a phantom `Jan28-Jan31` gap — requested is never wider than what the run actually attempted. Rewrote `test_sync_coverage_api_backfill_pair_scoped_to_planned_run_units` (previously pinning the wrong behavior) to assert the corrected semantics: a job wider than its run's actual unit window produces `gap_count == 0` / `status == "healthy"` for that pair, while the untouched pair/dataset still correctly get no permanent gap either.

**MAJOR — zero-unit runs conflated with unresolvable markers.** Fixed: replaced `_backfill_job_run_pairs` (a bare `set`) with `_resolve_backfill_job_pair_windows`, which returns `None` only when the marker is absent/unparseable (legacy all-pairs fallback), and a `dict` (possibly empty) once the marker resolves. An empty dict — a run that legitimately planned zero units — now contributes nothing for that job, instead of falling back to all-pairs-in-scope. Added `test_sync_coverage_api_backfill_marker_resolved_zero_units_contributes_nothing` covering this state explicitly.

**Gates (in worktree, all green):**
```
.venv/bin/ruff format --check .   # 1333 files already formatted
.venv/bin/ruff check .            # All checks passed
.venv/bin/mypy src/dev_health_ops/api/services/sync_coverage.py   # Success: no issues found
.venv/bin/python -m pytest tests/test_sync_coverage.py tests/api/admin/test_sync_coverage_api.py -q   # 29 passed
```
Also re-verified via the lefthook `pre-commit`/`pre-push` gates (full-tree ruff + mypy).

Payload JSON shape unchanged; no other files touched.

PR #1291, opened 2026-07-25
feat(workers): split the domain PostgreSQL role and activate the coordinator pool (CHAOS-3033, CHAOS-3113)

chrisgeo:
## Summary

- **What changed:** Splits the single `devhealth_domain` PostgreSQL runtime role into a **domain** role and a **coordinator** role ("Option B"), and lands the runtime that makes the split real: a third bounded connection pool, coordinator grants emitted from the migration, and the coordinator-attributed call sites in `dev-health-workerctl` and `dev-health-reconciler` repointed off the domain pool. Also carries the accumulated Wave 1 + Wave 2 Celery→Go cutover work already reviewed lane-by-lane (CUT-01…CUT-12).
- **Why:** The reconciled domain grants are least-privilege — coordinator-exclusive tables (`internal_service_credentials`, `worker_operator_audits`, `worker_job_routes`, `scheduled_jobs`, `scheduled_sync_occurrences`, `fixed_schedule_occurrences`, `sync_run_reference_discoveries`, `sync_run_post_dispatches`) carry no domain GRANT. Every binary still connected as the domain role, so operator and control-plane paths were denied at runtime. That is CHAOS-3113, and it is fixed here by giving those paths the role they actually need rather than by re-widening the always-on domain role.

**These two halves are co-requisite and must ship together.** The grant reconciliation alone converts the reconciler's outbox relay from working to `42501` on every tick (`buildReconcilerRelay` → `jobroute` → `SELECT … FROM public.worker_job_routes`, and `worker_job_routes` is coordinator-exclusive). Measured, not inferred — see Test Evidence.

### Role split

- `RuntimePools.Coordinator` + `CoordinatorPool()`, **opt-in per binary and fail-closed**: it returns `ErrUnavailable` rather than silently falling back to the domain pool, because a fallback would re-introduce exactly the bug this closes. Domain-only binaries (`dev-health-worker`, `dev-health-stream-runner`) are not forced to configure it.
- Validation mirrors the existing two pools: DSN present, valid identifier, DSN user matches the declared role, all three roles pairwise distinct, bounded max-conns, and a distinct error for a coordinator DSN pointed at the transaction-pooler endpoint.
- Config: `COORDINATOR_DATABASE_URI` (secret-wrapped), `RIVER_COORDINATOR_DATABASE_ROLE` (default `devhealth_coordinator`), `WORKER_COORDINATOR_DATABASE_MAX_CONNS` (default 2, bounds 1–4).
- Coordinator grants are emitted by the migration, with the table list injected from `postgres.CoordinatorPosture()` rather than hand-transcribed (`internal/storage/river` cannot import `internal/storage/postgres` — that package's tests import river, so the production import would be a cycle). One list, both sides.
- Repoints: workerctl's authenticator, auditor, `syncroute.Controller` and `jobroute.Controller`; the reconciler's route resolver and Materializer. Genuinely-domain components in the same binaries stay on the domain pool. Readiness gates on `CheckCoordinatorAuthorization`.
- Connection budget unchanged at **93/100** server connections; the coordinator is modeled as direct, server-counted connections, not PgBouncer-pooled.

### Two privilege facts a DML-verb grant analysis cannot see

Both verified empirically on PostgreSQL 16, and both hid a real requirement:

1. `LOCK TABLE … IN SHARE ROW EXCLUSIVE MODE` (and stronger) requires UPDATE/DELETE/TRUNCATE/MAINTAIN — **with no row ever written**. `jobroute.Rollback` locks `worker_job_outbox` this way while holding only SELECT+INSERT.
2. `SELECT … FOR UPDATE / FOR SHARE / FOR KEY SHARE` all require UPDATE; `FOR UPDATE OF <list>` only on the listed tables. That is why `sync_configurations` is a dual-grant table.

The manifest now documents this rule so the class is not re-lost, and a JOIN-inclusive sweep of every locking clause in `internal/` and `cmd/` found no further gaps.

## Checklist

- [x] I added/updated tests, or provided required governance evidence below.
- [x] I ran relevant local validation (minimum: `./ci/run_tests.sh unit` for `src/` changes).
- [x] I documented risk and rollback considerations.
- [x] I called out breaking changes, migrations, or config impacts.

## Test Evidence

TEST-EVIDENCE:

**Full gate — `env -i PATH=… HOME=… TMPDIR=… bash ci/local_validate.sh` → `GATE PASSED`, `GATE_EXIT=0`.** 9020 passed, 62 skipped, 0 failed (89.6s), including ruff/mypy, `go vet`, the ClickHouse scratch-DB stage and both river-compatibility suites. The scrubbed environment matters: direnv config leaks otherwise and produces false reds.

**The privilege claims are executed, not asserted.** `internal/storage/postgres/coordinator_statement_privileges_integration_test.go` runs 11 statement shapes lifted verbatim from the production call sites (`joboperator` auth + audit, `jobroute` read / `FOR UPDATE` / `UPDATE` / `LOCK`, `syncroute` pause + `FOR UPDATE`, `syncreconciler` materializer reads) against a real PostgreSQL, as the real roles:

- denied to the domain role with SQLSTATE `42501` — with the grants `ApplyPinnedMigrations` actually emits, not a copy of them;
- permitted to the coordinator role — with the grants that same migration emits, so this proves the migration issues what the statements need;
- `CheckCoordinatorAuthorization` accepts that role, plus a revoke/restore cycle on the LOCK-implied `worker_job_outbox` UPDATE.

Two vacuity traps are closed deliberately. Any SQLSTATE other than `42501` is treated as a **failure**, because `undefined_column` (42703) is raised during parse analysis *before* the permission check — a lazily-typed fixture would silently stop testing privileges. And the coordinator-exclusive tables are created *before* the migration runs, because every domain GRANT is `to_regclass`-guarded, so a table created afterwards would make the denial assertions pass even if the migration did grant it.

**Both directions verified by mutation.** Granting the domain role those privileges flips exactly the five corresponding statements to PERMITTED and leaves the other six denied; revoking coordinator `worker_job_outbox` UPDATE fails exactly the LOCK statement.

**Blast radius measured against the pre-branch base.** At `cd7aa8a8e` the same probe denied 11 of 12 shapes; the single cell this work changes is `SELECT worker_job_routes` (ALLOWED → DENIED). So the rest was already broken, and the new exposure is precisely the reconciler relay — a hot loop, not an operator path. `workerctl` was already non-functional at both commits, failing earlier at `Authenticate`.

## Risk Notes

RISK-NOTES:

**Deployment prerequisite — ordered, and this PR is not safe to deploy without it.** `dev-health-workerctl` and `dev-health-reconciler` now require a coordinator DSN and fail closed without one. Per environment: (1) create the coordinator role with a password, (2) publish `COORDINATOR_DATABASE_URI` as a secret to those services, (3) run the migration (its preflight fails closed if the role does not exist yet), (4) deploy. Role creation and the coordinator's table grants are automatic in local dev and CI; only the role's credential and DSN are genuine infra work. Full ordering in `.github/docs-legacy/architecture/chaos-3033-coordinator-pool-activation.md`.

**Rollback:** revert the merge. The migration only issues GRANT/REVOKE against the coordinator role, so reverting the code and leaving the role in place is inert. Do not revert the grant reconciliation without also reverting the repoint, or the reconciler relay breaks — the co-requisite constraint applies in both directions.

**Blast radius:** the repointed paths are operator commands (`workerctl`), the reconciler's route relay, and readiness checks. No hot data-plane path changes role. `dev-health-scheduler` is untouched and remains dormant.

**Known-open, filed, not fixed here:**

- **CHAOS-3114** — `Engine.runOccurrence` commits `fixed_schedule_occurrences` (coordinator-exclusive) together with `remaining_metric_*` and `work_graph_execution_requests` (domain-exclusive) in one transaction. No single role can hold it, so the scheduler cannot be repointed; it needs an architecture decision (outbox handoff vs. widening a role). Dormant behind its activation flags, and a **blocker on flipping them**.
- **CHAOS-3115** — `InspectEffect` reconstructs a winning row with seven independent `argMax(col, last_synced)` calls; under a genuine version tie they resolve separately and can return a per-column mix of two writers' rows. Pre-existing, latent, and no test constructs a real tie.
- **CHAOS-3116** — the testcontainers Ryuk reaper dies under concurrent container creation, failing random integration packages. A related package-scoped race *is* fixed here: `StartClickHouse` waited only on the HTTP port while every caller dials the native port.

**Two integration defects the lane merges surfaced** (textually clean, semantically broken — worth a reviewer's attention):

- `internal/domaingrants` regexed the posture out of the readiness query's SQL text; Phase 2 parameterized that query over data bound through `unnest`, so the shim stopped compiling and would have parsed zero rows if it had. Now reads `DomainPosture()` as data.
- The same checker special-cased DELETE as structurally unrepresentable — true before Phase 2 added `AllowDelete`, false after. It emitted six permanent false Criticals that would have masked a real DELETE gap.

**Pre-existing red fixed in passing:** `test_canonical_incident_feature_seed_is_in_single_head_graph` pinned `get_heads() == ["0064"]`, which fails on every legitimately-added migration; CUT-05's `0065` left it behind. It now asserts the invariant its name claims — exactly one head, and the head is the highest-numbered revision.

## Optional Context

- Linked issues: CHAOS-3033 (epic), CHAOS-3113 (fixed), CHAOS-3114 / CHAOS-3115 / CHAOS-3116 (filed, not fixed).
- Authoritative role partition: `.github/docs-legacy/architecture/chaos-3033-role-partition-manifest.md`. Activation and deploy ordering: `.github/docs-legacy/architecture/chaos-3033-coordinator-pool-activation.md`.
- Draft pending a Codex adversarial review pass over the consolidated diff.


chrisgeo:
### Code review

Five reviewers ran independently (repo-instruction compliance, shallow bug scan, git-history/blame context, prior-PR review feedback, code-comment compliance). Compliance, bug-scan, history and prior-PR lanes each returned nothing above the reporting bar. Found 5 issues, plus the two CI blockers diagnosed at the bottom.

1. `RequiresAnyWriteLock` misses `SHARE` and `ROW SHARE` lock modes, so the detector fails open in exactly the class it exists to catch. The mode test is a substring match on `"EXCLUSIVE"`, but per the PostgreSQL `LOCK` docs every form other than `ACCESS SHARE` requires `UPDATE`/`DELETE`/`TRUNCATE` — so `LOCK TABLE t IN SHARE MODE` needs a write privilege and is silently not flagged. Latent today (every in-repo `LOCK TABLE` is `SHARE ROW EXCLUSIVE`) and untested in `sql_test.go`. The sibling `FOR UPDATE` branch twelve lines above states the standard this misses: "a false negative is worse than a false positive here".

   To be precise about the other half: `ROW EXCLUSIVE` matching the substring is **not** a false positive, because explicit `LOCK TABLE ... IN ROW EXCLUSIVE MODE` also requires `INSERT`/`UPDATE`/`DELETE`/`TRUNCATE`. Only the missed modes are a defect, though the doc comment's "stricter than ROW EXCLUSIVE" framing is what makes the two easy to conflate.

   https://github.com/full-chaos/dev-health-ops/blob/e9c0fbe2f4fd24366eb0dbcc45a5caa56cc8ac77/internal/domaingrants/sql.go#L371-L377

2. Stale comment asserting the coordinator's `worker_job_outbox` grant is absent from `migrate.go` and that CHAOS-3113 is still open. Both stopped being true inside this same PR: `coordinatorGrantStatements` now emits it from `CoordinatorPosture()`, and `c1397ae31` repointed the call sites onto the coordinator pool. This matters more than ordinary comment rot — these comments are the derivation evidence for the grant surface, so a reader auditing current grant status is actively misled.

   https://github.com/full-chaos/dev-health-ops/blob/e9c0fbe2f4fd24366eb0dbcc45a5caa56cc8ac77/internal/storage/postgres/domain_authorization.go#L606-L615

3. The same staleness in the new privilege test's own prose: it says these call sites are wired to `pools.Domain` "today" and that each denial is "a live runtime failure, not a hypothetical", and four `site:` fields name `pools.Domain`. After the repoint the denials prove the domain role *would* be refused, which is still exactly the right assertion — but the surrounding text now describes the wiring as it was before the fix landed.

   https://github.com/full-chaos/dev-health-ops/blob/e9c0fbe2f4fd24366eb0dbcc45a5caa56cc8ac77/internal/storage/postgres/coordinator_statement_privileges_integration_test.go#L28-L32

4. Dual-grant count is stated as six; the postures now declare eight. `sync_configurations` (labelled "SEVENTH dual-grant" in `domain_authorization.go`) and `worker_job_outbox` both appear in `domainPosture()` and `coordinatorPosture()`. The test mechanism is data-driven and still exercises all of them correctly, so this is the comment only — but the partition is audited by these comments.

   https://github.com/full-chaos/dev-health-ops/blob/e9c0fbe2f4fd24366eb0dbcc45a5caa56cc8ac77/internal/storage/postgres/role_posture_integration_test.go#L355-L360

5. Dead documentation pointer in a new file: it directs readers to `.remember/grant-surface-derivation.md`, which does not exist and would be ephemeral scratch rather than shipped docs, and to `docs/architecture` for rationale — the real document is `.github/docs-legacy/architecture/chaos-3033-role-partition-manifest.md`, which the comment does not mention.

   https://github.com/full-chaos/dev-health-ops/blob/e9c0fbe2f4fd24366eb0dbcc45a5caa56cc8ac77/internal/storage/river/grantcheck_export.go#L10-L13

### CI blockers, both real

**`go-quality` — breaking job-contract change.** `system.retention_cleanup@1.properties.retention_policy.enum: schema constraint changed`. The enum widens from `["worker_job_terminal"]` to add `rate_limit_observations` and `external_ingest_batches`. `compareSchema` compares every non-ignored key with `reflect.DeepEqual`, so any enum change on a `@1` contract is breaking with no additive allowance — and it is right to be strict here, because `ea5e8aa59` records that the Go loader "pins the schema enum to the compiled policy list", meaning an already-deployed worker rejects the new values during a rolling deploy. Resolution is a decision (bump to `@2`, or accept as knowingly breaking with a stated deploy constraint), not a mechanical fix.

https://github.com/full-chaos/dev-health-ops/blob/e9c0fbe2f4fd24366eb0dbcc45a5caa56cc8ac77/internal/jobcontract/compatibility.go#L145-L151

**`go-container-smoke` — reproduced locally, and not infrastructure.** The worker container exits before binding `:8080` with `configuration error: profile must be one of sync, heavy, ops`, which surfaced only as the opaque `no public port '8080/tcp' published`. Cause: CUT-02 removed `DefaultProfile: "latency"`, while `ci/check_go_containers.sh` starts every target with no environment at all.

Supplying `DEV_HEALTH_PROFILE` gets the worker live with `/healthz` 200 and `/readyz` 503 as intended, but `/metrics` then returns 503 and the check requires 200. That is the deeper half: `buildQueueTelemetry` sets `queueTelemetryRequired` whenever the profile has registered descriptors and the process has queues, and with no database configured `workerMetricsSource.WritePrometheus` returns an error, which 503s the whole endpoint. The removed `latency` profile had zero descriptors, so it was the only configuration under which an unconfigured worker could serve `/metrics` — removing it removed this check's last passing path, for every remaining profile.

Worth deciding rather than patching: a `/metrics` endpoint that 503s because one source cannot reach the database also drops `dev_health_runtime_live`, `_ready` and `_uptime_seconds` — the process-level gauges are most useful precisely when the process is unready.

https://github.com/full-chaos/dev-health-ops/blob/e9c0fbe2f4fd24366eb0dbcc45a5caa56cc8ac77/cmd/dev-health-worker/dependencies.go#L440-L446

Also note `Build Image` failed on a Docker Hub timeout (`registry-1.docker.io … Client.Timeout exceeded`) and is unrelated to the change; it just needs a retry.

PR #1303, opened 2026-07-25
fix(test): repair the two integration guards the River cutover invalidated

chrisgeo:
`go-storage-integration` is red on `main` for two tests unrelated to the 20-minute deadlock in #1300. Both fail fast:

```
relay_integration_test.go:92: checked-in production route became executable:
  jobruntime.Descriptor{Kind:"investment.chunk", ..., MigrationState:"go_default", Route:"river", RollbackRoute:"celery"}

postgres_integration_test.go:122: checked-in heartbeat policy =
  jobruntime.Descriptor{Kind:"system.heartbeat", ..., MigrationState:"go_default", Route:"river", RollbackRoute:"celery"}, want non-executable
```

## Cause — stale fixture, not a cutover defect

#1292 (ee2141eca) moved 23 of 24 job kinds to `state: go_default` / `route: river`, leaving only `sync.provider_unit` on `canary`/`river_canary`. Both tests encoded "every checked-in kind except provider_unit sits on celery and is therefore non-executable" — true before the cutover, false by design after it.

Checked against the unedited `contracts/jobs/v1/migration-state.json`. **No production defect found in ee2141eca**; the guards were firing on a correct registry.

## The fix keeps the guards guarding

The failure mode to avoid here was updating a literal to the new route, which makes both tests green while quietly deleting what they protect. Two different problems needed two different treatments:

**Drift tripwires** — the top-level loop in `relay_integration_test.go` and the heartbeat assertion in `postgres_integration_test.go` exist to notice when checked-in policy changes. Their literals now name the real current policy (`river` / `go_default` / `celery` rollback / executable), so they resume tripping on the *next* unintended change. Both still run the service under test against the raw, unmodified registry.

**Behavioral subtests** — three subtests ("celery-deferred known kind…", "celery rollback never reclaims…", "relay leaves known celery rows pending…") were exercising the relay's deferred path, which `relay.go`'s `deferredRelayKinds()`/`claimDueExcept()` key purely on `Route == "celery"`. They had been relying on a production kind that *happened* to still be celery. Production now has zero such kinds, so the behavior they test would silently stop being tested.

These now override the route to `celery` explicitly via the existing `integrationRouteRegistry` helper. That decouples the behavior under test from whatever the contract currently promotes — the deferred-relay path stays covered no matter how the migration state moves next.

## Swept for the same staleness elsewhere

`cmd/dev-health-worker/reports_test.go`, `dependencies_test.go`, and `internal/joboperator/service_test.go` were already made cutover-aware in earlier work (`promotedContractRoot` / `demotedContractRoot` / `executableRegistry` helpers). No changes needed.

## Verification

Both tests pass with `-tags=integration` against real Docker, all subtests. `go build ./... && go test ./...` clean.

## Flagged, not fixed here

`go test -tags=integration ./...` reveals a **pre-existing compile break**: `cmd/dev-health-workerctl/main_integration_test.go` does not build because `newJobRouteController`'s signature drifted in #1291, before the commit this PR targets. An integration test that does not compile is not being run at all, and the packaging is such that nothing surfaced it. Untouched here; being handled separately.

## Gate

All stages pass except the two `tests/docs/*` failures caused by `ci/local_validate.sh` not installing `requirements-docs.txt` — a gate defect fixed in #1301, not a code fault.

### DIFF
--- a/internal/joboperator/postgres_integration_test.go
+++ b/internal/joboperator/postgres_integration_test.go
@@ -38,7 +38,7 @@ type allowIntegrationDomainGuard struct{}
 func (allowIntegrationDomainGuard) Check(context.Context, Action, JobSummary) error { return nil }
 
 // operatorIntegrationPolicyRegistry changes only the fixture's insertion
-// policy. The operator backend continues to use the checked-in, Celery-routed
+// policy. The operator backend continues to use the checked-in, unmodified
 // registry so this test cannot weaken or misrepresent production routing.
 type operatorIntegrationPolicyRegistry struct {
 	registry *jobruntime.Registry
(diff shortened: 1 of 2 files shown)


chrisgeo:
Superseded by #1300 — these two commits must land together. #1300 removes the 20-minute deadlock but its go-storage-integration job can never go green on its own, because the stale-fixture failures fixed here live on `main` independently of it. Cherry-picked onto #1300 as 5d5bac31b so the job goes green in one merge rather than passing through an intermediate red state.

PR #1344, opened 2026-07-31
fix(ask-dev): make organization scope executable for status/change reads (CHAOS-3255)

chrisgeo:
## Summary

Fixes CHAOS-3255: organization scope (an approved V1 direct scope) was not executable for status/change reads. Organization resolved to `outcome=exact` while contributing zero repository IDs, so `status_snapshot.v1`/`change_summary.v1` always refused with "Status reads require the complete authorized repository set; scope was not widened." even when the org had connected repositories and status data.

## Root cause

- `ScopeResolutionService.resolve_contract` (`scope_service.py`) built `authorized_repository_ids` only from `REPOSITORY`-kind entities or `entity.repository_id`; organization scope resolves to a single `AuthorizedEntity(ORGANIZATION, org_id, ...)` with no repository IDs attached, so the list was always empty.
- `ClickHouseStatusChangeSource.status_snapshot`/`change_summary` (`native_status_change.py`) read `scope.repositories`/`scope.entity_refs` directly, both empty by contract for organization scope, and refused the read.
- Adversarial review turned up a deeper layer: even after wiring the repository set through, `_PULL_REQUESTS_SQL`, `_DEPLOYMENTS_SQL`, `_TRANSITIONS_SQL`, and `_RELATIONSHIPS_SQL` had no `scope_type = 'organization'` branch in their `WHERE` disjunctions (unlike the five *_CHANGES_SQL queries, which already supported it) — so status/change reads for org scope would still silently return zero facts.

## Fix

- `ScopeResolutionService.resolve_contract` now enumerates the organization's complete authorized repository set via a new `AuthorizedEntityCatalog.organization_repository_ids` method.
  - Org at or under the 20-repository contract cap (`MAX_REPOSITORIES`, matching `DevScopeResolution.authorized_repository_ids`'s `max_length`) → full list inline.
  - Org above the cap → `authorized_repository_ids` stays empty (never truncated to a misleadingly "complete" 20-entry list) with an explanatory warning; outcome/resolved_scope are preserved as organization scope.
  - Zero authorized repos → outcome downgrades to `unresolved` (`resolved_scope=None`), an explicit insufficient-evidence result rather than an exact-but-empty scope.
  - A team-filtered organization scope is excluded from this enumeration (see below).
- `ClickHouseAuthorizedEntityCatalog.organization_repository_ids` fetches the id page and true total in a **single** query using `count() OVER ()` (verified live against ClickHouse that it reflects the pre-`LIMIT` total), so a concurrent repo insert/delete between two sequential queries can never desync a truncated page from the reported count.
- `ClickHouseStatusChangeSource._authorized_repository_ids` re-derives the authorized repository set directly from `org_id` for organization scope — an organization-native execution path that scales to any repository count without ever serializing a bounded list, bypassing the wire contract's 20-entry cap entirely for the actual query execution. **Excluded when `scope.team_ids` is non-empty** — no native query here applies a team filter, so auto-deriving the full org repo set for a team-filtered request would silently widen it; that case falls through to the pre-existing fail-closed behavior instead.
- Added `scope_type = 'organization'` to the four native SQL queries that were missing it, matching the pattern already used by the five delivery-change queries.

## Tests

- `tests/api/dev/test_scope_service.py` — org-scope repository enumeration (full set inline, zero-repos → unresolved, above-cap → non-truncated + warning, catalog failure → unresolved, request-level caching, non-org scope never queries org repos).
- `tests/api/dev/test_scope_catalog.py` — `organization_repository_ids` is tenant-scoped and reports the true total from a single window-function query.
- `tests/api/dev/test_native_status_change.py` — org-scope status/change reads surface actual PR/deployment/transition/relationship facts (not just the absence of the old warning), zero-repo org is explicit/degraded not masked, team-filtered org scope is never silently widened.
- `tests/api/dev/test_organization_scope_clickhouse_live.py` (new, `@pytest.mark.clickhouse`) — live-ClickHouse integration: a multi-repo org's full authorized set reaches `status_snapshot.v1`/`change_summary.v1`; an org above the cap stays organization-native and non-truncated; an empty org gets explicit insufficient evidence; a second tenant's repositories never enter the first tenant's authorized set. Verified locally against a running ClickHouse instance (inserted/cleaned synthetic data, zero residual rows afterward).

## Gate / review

- `bash ci/local_validate.sh` — fully green (ruff format/check, mypy, go, river, ClickHouse scratch migrate, full unit suite 9500+ passed, argMax live-exec proof), re-run after every round of fixes below.
- Two rounds of Codex adversarial review (`--effort high`), all findings fixed and re-reviewed:
  - **Round 1**: HIGH — organization `status_snapshot.v1` remained effectively empty (missing `scope_type='organization'` branches on 4 queries) — fixed. MED — team-filtered organization scope silently widened to every repo — fixed (guard on `scope.team_ids`). MED — organization `change_summary.v1` silently omitted transitions/relationships — fixed (same branch fix). MED — repository id-page/count race in `scope_catalog.py` — fixed (single window-function query, verified live against ClickHouse).
  - **Round 2**: MED — the new org-branch tests matched on table name only and would still pass with the organization branch deleted again (table-name-only mock can't detect a missing WHERE-clause branch) — fixed by asserting the literal SQL text; mutation-verified (reverted the fix, confirmed the strengthened tests fail, restored). LOW — the >20-repo warning was still emitted for a team-filtered org even though the team-filter guard prevents organization-native execution for that case, making the warning misleading — fixed by skipping repository enumeration entirely when a team filter is present.
- **Disclosed residual risk (not fixed, out of scope for this ticket)**: team filtering is not applied by any native status/change SQL query in this file — a team-filtered *direct repository* scope has the same limitation as team-filtered organization scope. This predates this change; CHAOS-3255 only ensured organization scope doesn't newly widen around it.

## Test plan

- [x] `bash ci/local_validate.sh` fully green
- [x] Live ClickHouse integration test passes against a real instance
- [x] Two rounds of Codex adversarial review, all findings resolved and re-verified
- [x] Mutation-verified the strengthened regression tests actually fail when the fix is reverted

### DIFF
--- a/src/dev_health_ops/api/dev/native_status_change.py
+++ b/src/dev_health_ops/api/dev/native_status_change.py
@@ -284,6 +284,12 @@ ORDER BY observed_at DESC, entity_id
 LIMIT {limit:UInt32}
 """
 
+_ORGANIZATION_AUTHORIZED_REPOSITORIES_SQL = """
+SELECT toString(id) AS repository_id
+FROM repos FINAL
+WHERE org_id = {org_id:String}
+"""
+
 _TRANSITIONS_SQL = """
 SELECT transition.work_item_id AS entity_id, item.title AS display_label,
        transition.from_status, transition.to_status,
@@ -578,10 +584,38 @@ class ClickHouseStatusChangeSource:
         self._policies = dict(policies or default_native_freshness_policies())
         self._now = now
 
+    async def _authorized_repository_ids(
+        self, org_id: str, scope: DevScope
+    ) -> list[str]:
+        """Bound status/change reads to the server-owned repository set.
+
+        Organization scope cannot serialize its full authorized repository
+        set onto the wire (``DevScope.repositories`` is capped at 20 entries
+        by the ask-dev/v1 contract), so it is re-derived here directly from
+        the same ``org_id`` boundary every other native query already
+        enforces (CHAOS-3255). This scales to any repository count without
+        truncation or widening. Every other direct scope keeps using the
+        scope's own bounded repository/entity refs.
+        """
+        if scope.direct_scope is not DirectScope.ORGANIZATION:
+            return self._repository_ids(scope)
+        try:
+            async with asyncio.timeout(QUERY_TIMEOUT_SECONDS):
+                rows = await query_dicts(
+                    self._client,
+                    _ORGANIZATION_AUTHORIZED_REPOSITORIES_SQL,
+                    {"org_id": org_id},
+                )
+        except Exception:
+            return []
+        return sorted(
+            {str(row["repository_id"]) for row in rows if row.get("repository_id")}
+        )
+
     async def status_snapshot(
         self, *, org_id: str, scope: DevScope, as_of: datetime, limit: int
     ) -> RawStatusSnapshot:
-        repositories = self._repository_ids(scope)
+        repositories = await self._authorized_repository_ids(org_id, scope)
         entity_id = self._entity_id(scope)
         scope_type = scope.direct_scope.value
         warnings: list[str] = []
@@ -938,7 +972,7 @@ class ClickHouseStatusChangeSource:
         limit: int,
     ) -> RawChangeSummary:
         del comparison
-        repositories = self._repository_ids(scope)
+        repositories = await self._authorized_repository_ids(org_id, scope)
         if not repositories:
             return RawChangeSummary(
                 changes=(),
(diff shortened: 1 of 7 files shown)


chrisgeo:
Resolved the CodeQL py/ineffectual-statement thread: line 243 is the `...` body of a Protocol method stub (standard typing idiom), not dead code. False positive.

PR #1345, opened 2026-07-31
fix(ask-dev): resolve named entities before executing status tools (CHAOS-3256)

chrisgeo:
## Summary

Fixes CHAOS-3256: the production `resolve_scope.v1` executor ignored the model's validated `query`/`limit` inputs entirely, calling `_resolve_exact_contract(..., requested_scope=request.scope)` — i.e. re-resolving the caller's already-authorized scope. Asking "What's the status of the Ask Dev project" from inherited organization scope therefore ran `status_snapshot.v1` against the organization, never the named project, even when the model correctly called `resolve_scope.v1` with a query.

## Fix

- **`scope_service.py`** — new `ScopeResolutionService.resolve_query_contract`: the canonical seam `resolve_scope.v1` uses when the model names an entity. Searches only the authenticated organization's authorized catalog (never free-form SQL/GraphQL, never a model-authored ID):
  - Exact match → commits a new authorized `DevScope` (project/work-unit/repository/issue/PR) for the resolved entity.
  - Multiple matches → typed `AMBIGUOUS` candidates.
  - No matches → `FORBIDDEN_OR_NOT_FOUND`; **never** falls back to organization scope.
- **`production_runtime.py`** — `resolve_scope` executor now calls `resolve_query_contract` when the model supplies a non-empty `query`, searching across every non-organization direct-scope kind. An empty/omitted query keeps the prior re-authorization behavior (defensive fallback; the tool's exposed JSON schema marks `query` required).
- **`orchestrator.py`** — `DevOrchestrator.run` now commits a successful `resolve_scope.v1` exact match's `resolved_scope` as `authorized_scope` for every subsequent tool call in the run, instead of reusing the single immutable scope captured before the model's first decision round. Guarded to only fire for `ToolID.RESOLVE_SCOPE`, only on `status == "success"`, only when `resolved_scope` is non-`None`, and only when it belongs to the authenticated `org_id` (defense in depth; the catalog is already tenant-scoped).

## Adversarial-review fix

Codex found a HIGH: `resolve_query_contract` passed the model-owned `request.limit` straight through to `search()`, so a caller-supplied `limit=1` with 2 real matches in the catalog silently truncated the search results to 1 *before* the ambiguity check — committing an arbitrary one of several matches as `EXACT` instead of reporting `AMBIGUOUS`.

Fix: `resolve_query_contract` now always probes with `limit` pinned to `MAX_CANDIDATES` (25) via `dataclasses.replace`, decides exact/ambiguous/not-found against that untruncated result, and only truncates to the caller's requested `limit` when building the typed candidate list for an ambiguous outcome. Added `test_resolve_query_contract_detects_ambiguity_despite_a_narrow_limit`.

## Tests

- `tests/api/dev/test_scope_service.py` — `resolve_query_contract` across every supported direct scope kind (repository/project/work_unit/issue/pull_request), ambiguous candidates, not-found/cross-tenant never falling back to org scope, and the narrow-limit ambiguity regression.
- `tests/api/dev/test_production_runtime.py` — `resolve_scope` executor with a query searches the authorized catalog and resolves to the named entity; without a query keeps the legacy re-authorization path.
- `tests/api/dev/test_orchestrator.py` — full orchestrator run for the literal question "What's the status of the Ask Dev project" from inherited organization scope, asserting `status_snapshot.v1` receives project scope (not organization scope) and the final answer's `resolved_scope` reflects the committed project scope; a second test proves an unresolved named-entity reference never regresses subsequent tool calls back to organization scope.

## Gate / review

- `bash ci/local_validate.sh` — fully green (ruff format/check, mypy, go, river, ClickHouse scratch migrate, full unit suite 9500+ passed, argMax live-exec proof) both before and after the adversarial-review fix.
- Two rounds of Codex adversarial review (`--effort high`):
  - **Round 1**: HIGH — `resolve_query_contract` truncated to the model's requested `limit` before checking ambiguity — fixed (see above).
  - **Round 2**: re-reviewed the fix; 16 focused tests passed, the cache/watermark interaction is correct. One residual **MED disclosed, not fixed**: `ScopeResolutionService.search()` itself asks the catalog for at most `limit` rows before `_dedupe_entities()`; a hypothetical catalog implementation that returns multiple raw rows for the same canonical entity (e.g. a future join-based search) could theoretically still under-count distinct matches within that page. The production `ClickHouseAuthorizedEntityCatalog.search()` is not affected today — every kind's query is a direct single-table `SELECT ... FROM <table> FINAL WHERE ...` with no join that could duplicate rows per entity, so one row always equals one entity. Flagging as a protocol-contract expectation (`AuthorizedEntityCatalog.search()` implementations must not return duplicate `(kind, canonical_id)` rows within their own `limit`) for whoever builds the next catalog implementation, rather than fixing a non-reachable path now.

## Test plan

- [x] `bash ci/local_validate.sh` fully green
- [x] Codex adversarial review (2 rounds), the reachable finding resolved and re-verified; one non-reachable protocol-level edge case disclosed above

## Depends on

Independent of CHAOS-3255 (different code paths); no merge-order requirement, though both touch `scope_service.py` in non-overlapping regions.

### DIFF
--- a/src/dev_health_ops/api/dev/orchestrator.py
+++ b/src/dev_health_ops/api/dev/orchestrator.py
@@ -664,6 +664,26 @@ class DevOrchestrator:
                     tool_results.append(execution.result)
                     tool_requests.append(tool_request)
                     tool_bytes_total = next_total
+                    committed_resolution = execution.result.scope_resolution
+                    committed_scope = (
+                        committed_resolution.resolved_scope
+                        if committed_resolution is not None
+                        else None
+                    )
+                    if (
+                        tool_request.tool_id is ToolID.RESOLVE_SCOPE
+                        and execution.result.status == "success"
+                        and committed_resolution is not None
+                        and committed_scope is not None
+                        and committed_scope.organization_id == org_id
+                    ):
+                        # A named-entity resolve_scope.v1 match commits a new
+                        # server-authorized scope for every subsequent tool
+                        # call in this run (CHAOS-3256) — the prior immutable
+                        # authorized_scope must not keep being reused once the
+                        # model has resolved a more specific entity.
+                        resolution = committed_resolution
+                        authorized_scope = committed_scope
                     provider_continuation += (
                         AgentMessage(
                             AgentMessageRole.ASSISTANT,
--- a/src/dev_health_ops/api/dev/production_runtime.py
+++ b/src/dev_health_ops/api/dev/production_runtime.py
@@ -76,11 +76,13 @@ from .prompts import PROMPT_VERSION
 from .runtime import BoundedDevRuntime, DevRuntimeUnavailable
 from .scope_catalog import ClickHouseAuthorizedEntityCatalog
 from .scope_service import (
+    DIRECT_SCOPE_KINDS,
     EntityKind,
     ScopeRef,
     ScopeRequestCache,
     ScopeResolutionService,
     ScopeResolveRequest,
+    ScopeSearchRequest,
     TimeRangeRequest,
 )
 from .status_change_service import (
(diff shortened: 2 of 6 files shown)


chrisgeo:
Resolved the Gitleaks thread: the flagged string was a hard-coded test fixture (not a real credential — nothing to rotate). Commit e2cbd46e7 was rewritten out of the branch; the fixture is now runtime-constructed via secrets.token_hex(32) and the current Gitleaks scan passes.

PR #1353, opened 2026-07-31
fix(ask-dev): CHAOS-3289 insufficient-evidence guard for unresolved named-entity answers

chrisgeo:
## Summary

Fixes CHAOS-3289: a status question naming a nonexistent entity ("What's the status of the Ask Dev project") was answered as if the entity existed. The run never called `resolve_scope.v1` (or called it and got ambiguous/not-found back), yet the final answer narrated organization-wide tool output under the asked-about entity's name — a fabricated-premise answer.

## Root cause / fix

`resolve_scope.v1` itself was already fixed by CHAOS-3256 (not-found returns not-found, no org fallback). This defect is the layer above: nothing enforced that a **status question naming an entity** actually produces a resolved-entity answer, a typed disambiguation, or an explicit not-found/insufficient-evidence — the model could skip `resolve_scope.v1` entirely, or call it and get not-found, and still narrate a status under the org's data.

Orchestrator-level fix, deliberately **not touching `answer_validator.py`** (owned by a separate lane working CHAOS-3288/3290):

- `DevOrchestrator.run` now tracks the run's `resolve_scope.v1` attempts (`resolve_scope_attempted`, `last_resolve_scope_outcome`).
- At `AgentFinalAnswer` time, `_unresolved_named_entity_error()` forces `RunState.INSUFFICIENT_EVIDENCE` instead of `COMPLETED` for `QuestionClass.STATUS` answers still under organization scope, in the shapes that indicate a fabricated premise:
  1. the run's last `resolve_scope.v1` attempt came back **ambiguous** or **not-found**, yet the model still tried to finalize a substantive answer; or
  2. `resolve_scope.v1` was **never attempted** and the answer carries **zero evidence-backed claims** (the literal repro shape); or
  3. `resolve_scope.v1` was **never attempted**, the answer has real claims (so (2) doesn't fire), but the answer's own text still echoes a named entity the question asked about (the *mixed variant* — genuine org-wide evidence narrated under a fabricated name).
- `prompts/composer.py` gained an explicit `named_entity_resolution` policy section as defense in depth (call `resolve_scope.v1` before other tools when a question names an entity; never describe org-wide results under an unresolved name).

Legitimate organization-wide STATUS answers with real grounded claims are unaffected — CHAOS-3255 made organization scope a fully valid, executable answer target on its own, and this guard never fires once real, non-echoing claims exist.

### Fixes applied after a Codex adversarial review of the above

- **HIGH**: an errored/rejected/timed-out `resolve_scope.v1` call (`ToolRequestRejected`/`ToolExecutionTimedOut`, so `execution.result.scope_resolution` is `None`) was flipping `resolve_scope_attempted=True` with no real outcome to judge — silently disarming *every* guard check (neither the ambiguous/not-found branch nor the never-attempted/no-evidence branch would fire), which is worse than never having called the tool. Fixed: such an attempt is now treated identically to never having called `resolve_scope.v1` at all.
- **HIGH**: the question/answer text-echo regex only recognized `status of/about/regarding/for the X project`-shaped phrasing. Broadened to also catch noun-trailing-status (`the X project status`) and noun-leading-name (`check project X status`) orderings, plus a few more connector words (`with`, `on`, `going on with`, `happening with`).

Both fixes are mutation-verified (see Tests below). Three MED findings and one narrower residual regex gap from the same reviews were **not** fixed here — deferred to CHAOS-3292 (see below) as they require larger design decisions (cross-question-class enforcement, per-attempt resolution tracking, and a live-confirmed false-positive-rate concern in a pre-existing, unrelated heuristic) that exceed this PR's minimal-blast-radius scope.

## Scoping rationale

The guard covers **`QuestionClass.STATUS` only** — this matches the literal reported repro with minimal blast radius. Extension to `REMAINING_WORK`/`OBSERVED_CHANGE`/other question classes is tracked as follow-up (CHAOS-3292), since the system prompt already asks the model to resolve named entities before *any* tool, but the runtime guard doesn't yet enforce that universally.

## Tests

8 new/extended orchestrator tests, each mutation-verified (reverted the specific guard clause in isolation, confirmed *only* the intended test(s) failed, restored):
- `test_status_answer_after_not_found_resolution_is_insufficient_evidence`
- `test_status_answer_after_ambiguous_resolution_is_insufficient_evidence`
- `test_status_answer_skipping_resolve_scope_with_no_claims_is_insufficient_evidence` (literal repro)
- `test_grounded_organization_wide_status_answer_still_completes` (no false positive on legit org-wide answers)
- `test_status_answer_narrating_unresolved_named_entity_with_real_claims_is_blocked` (mixed variant)
- `test_status_answer_about_a_different_topic_with_real_claims_still_completes` (no false positive on unrelated org-wide text)
- `test_status_answer_after_errored_resolve_scope_attempt_is_insufficient_evidence` (new — HIGH finding #1 regression)
- `test_status_answer_narrating_unresolved_named_entity_other_phrasings_is_blocked[...]` ×3 (new — HIGH finding #2 regression, parametrized over the phrasings Codex found bypassing the original regex)

Full ask-dev suite: 411/411 pass (`tests/api/dev/`, excluding the live-ClickHouse-only file). `bash ci/local_validate.sh`: fully green twice (9674 then 9678 tests — before and after the post-review fixes — plus lint/mypy/go/river/ClickHouse-scratch-migrate/argMax-live-exec).

## Live re-verification (real BYO gpt-4o-mini, seeded stack)

Verified against a disposable second API process running this worktree's patched code inside the existing container (the running `dev-health-api-1` bind-mounts the main checkout, not this worktree — see `/tmp/claude/chaos-3289-fix-verify/SUMMARY.md` for the exact methodology, torn down after verification, nothing left outside `/tmp`).

- **Negative case** (fabricated "Ask Dev project", no such project exists): **5/5 runs correctly blocked**, zero fabrication in any run. Representative: *"Scope resolution for Ask Dev project within organization ... is forbidden/not found. No authorized entities ..."*
- **Positive control** ("Fullchaos project", the org's real project): **1/10 runs completed with a genuine grounded answer**; the other 9 were blocked (5 empty-claims, 1 scope_ambiguous, 1 scope_not_found, 2 model self-refusal). Traced this is **not a regression from this PR's changes** — it's pre-existing behavior in the empty-claims heuristic, unrelated to both HIGH-finding fixes above. This confirms live one of the Codex MED findings; filed as the top item in the follow-up issue rather than fixed here given the severity asymmetry (a fabricated answer is the trust violation this ticket exists to close; an over-cautious refusal is degraded-but-safe UX) and that a principled fix is a larger design change.

Full evidence (dev_runs/dev_tool_calls rows, SSE payloads, methodology): `/tmp/claude/chaos-3289-fix-verify/`.

## Follow-up

Filed **CHAOS-3292** for: (1) extending the guard to REMAINING_WORK/OBSERVED_CHANGE and other question classes, (2) per-attempt (not just latest) resolve_scope.v1 outcome tracking, (3) remaining regex-backstop gaps (lowercase/slug names), and (4) the live-confirmed empty-claims false-positive-rate finding above.

## Test plan

- [x] `bash ci/local_validate.sh` fully green (twice — before and after the post-Codex-review fixes)
- [x] 8 new/extended tests, all mutation-verified
- [x] Live re-verification: negative case blocked 5/5, positive control demonstrated to still complete (with a documented, pre-existing, unrelated reliability caveat filed as follow-up)
- [x] Codex adversarial review (two passes — pre and post the HIGH-finding fixes)

### DIFF
--- a/src/dev_health_ops/api/dev/orchestrator.py
+++ b/src/dev_health_ops/api/dev/orchestrator.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 import asyncio
 import hashlib
 import json
+import re
 import time
 from collections import Counter
 from collections.abc import Awaitable, Callable, Mapping
@@ -59,8 +60,10 @@ from .contracts import (
     DevScopeResolution,
     DevToolRequest,
     DevToolResult,
+    DirectScope,
     FreshnessState,
     MetricID,
+    QuestionClass,
     ScopeResolutionOutcome,
     ToolID,
     dev_error_remediation,
@@ -91,6 +94,51 @@ def _dev_tool_request_limit_maximum() -> int:
         ):
             return constraint.le
     raise RuntimeError("dev_tool_request.v1 limit field must declare an upper bound")
+
+
+# CHAOS-3289: when resolve_scope.v1 was never attempted, the only remaining
+# signal that the question named a specific entity is the question text
+# itself. This pattern is deliberately narrow (an explicit, capitalized
+# entity name adjacent to one of the entity nouns Ask Dev tools actually
+# support, in one of a handful of common orderings) to keep false positives
+# on ordinary organization-wide prose vanishingly rare -- it is a backstop
+# for the exact fabricated-premise shape this issue reproduces (a genuinely
+# grounded, evidence-backed answer narrated under a name that was never
+# resolved), not a general NLU pass. Known residual gap: lowercase/slug
+# names and paraphrases outside these orderings still bypass it (tracked as
+# follow-up -- see CHAOS-3289's linked extension issue).
+_ENTITY_NAME_PATTERN = r"[A-Z][A-Za-z0-9&/'\-]*(?:\s+[A-Z][A-Za-z0-9&/'\-]*){0,3}"
+_ENTITY_NOUN_PATTERN = (
+    r"project|repository|repo|team|issue|pull request|deployment|incident|work unit"
+)
+
+# "status of/about/regarding/for/with/going on with the <Name> <noun>"
+_NAMED_ENTITY_REFERENCE = re.compile(
+    r"\b(?:status of|about|regarding|for|with|on|going on with|happening with)\s+"
+    r"(?:the\s+)?"
+    rf"({_ENTITY_NAME_PATTERN})"
+    rf"\s+(?:{_ENTITY_NOUN_PATTERN})\b"
+)
+
+# "the <Name> <noun> status" / "the <Name> <noun>'s status" -- the noun and
+# "status" trail the name instead of a leading preposition triggering it
+# (e.g. "What's the Ask Dev project status?").
+_NAMED_ENTITY_NOUN_STATUS = re.compile(
+    rf"\b({_ENTITY_NAME_PATTERN})\s+(?:{_ENTITY_NOUN_PATTERN})'?s?\s+status\b"
+)
+
+# "<noun> <Name>" -- the noun leads the name instead of trailing a
+# preposition (e.g. "Can you check project Ask Dev status?").
+_ENTITY_NOUN_LEADING = re.compile(
+    rf"\b(?:{_ENTITY_NOUN_PATTERN})\s+({_ENTITY_NAME_PATTERN})\b"
+)
+
+
+def _named_entity_phrases(text: str) -> frozenset[str]:
+    phrases = set(_NAMED_ENTITY_REFERENCE.findall(text))
+    phrases.update(_NAMED_ENTITY_NOUN_STATUS.findall(text))
+    phrases.update(_ENTITY_NOUN_LEADING.findall(text))
+    return frozenset(match.casefold() for match in phrases)
 
 
 _HARD_LIMIT_MAXIMA: dict[str, int | float] = {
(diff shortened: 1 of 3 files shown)


chrisgeo:
Wave 3.1 corrective follow-up: this PR successfully closed one fabricated-premise path, but its live result is not acceptable as the final Ask Dev contract. The recorded real-project positive control completed with a useful grounded answer in only 1/10 attempts, and the current UI exposes the late guard as an empty `refused` result with `forbidden/not_found`.

CHAOS-3289 has been reopened under CHAOS-3293. The authoritative replacement is now:

- PRD v2: https://linear.app/fullchaos/document/amendment-prd-v2-ask-dev-wave-31-evidence-backed-engineering-a65342b2bd80
- TRD v2: https://linear.app/fullchaos/document/amendment-trd-v2-ask-dev-wave-31-deterministic-investigation-health-d6e045a77554

Implementation direction: do not extend `_ENTITY_NAME_PATTERN`, `_named_entity_phrases`, `resolve_scope_attempted`, or `last_resolve_scope_outcome` into the primary architecture. Replace the late regex/answer-deletion path with server-owned intent, per-mention preflight/resolution ledger, deterministic investigation plans, relationship-closed source collection, a canonical answer frame, optional narrative, and deterministic fallback. The guard may remain temporarily as non-user-visible defense-in-depth during shadow migration, but it must not produce public copy or delete a valid frame.

The expanded Wave 3.1 scope also restores first-class portfolio/project status, project health, team health, workload/investment balance, and operational-deficiency questions before Wave 4.

PR #1363, opened 2026-08-01
fix(ask-dev): clear Semgrep OSS false positives in 0074/0075 migrations

chrisgeo:
## Summary
- Semgrep OSS flagged 5 errors/4 warnings on `main` after #1361 (CHAOS-3299) merged: f-string SQL/DDL construction in the two v2-persistence migrations, `0074_add_ask_dev_v2_persistence.py` and `0075_validate_ask_dev_v2_constraints.py`.
- All five are false positives: the interpolated values are a literal table-name tuple or module-level constraint-name constants, never external input.
- `0074`'s row-count check is genuine dynamic SQL that SQLAlchemy Core can express directly, so it's rewritten as `sa.select(sa.func.count()).select_from(sa.table(table))` instead of `sa.text(f"...")`.
- The four `DROP`/`VALIDATE CONSTRAINT` DDL statements keep their f-strings (SQLAlchemy has no bind-parameter syntax for DDL identifiers) and get a scoped `nosemgrep` naming the exact rule ids that fired, plus a one-line comment explaining why.

TEST-EVIDENCE:
- `.venv/bin/python -m pytest tests/test_ask_dev_v2_persistence_migration.py -q` — 4 passed, 2 skipped
- Live-Postgres proof against a fresh scratch DB, exercising the edited downgrade/validate paths directly: `DEV_HEALTH_POSTGRES_TEST_URI=postgresql://postgres:postgres@localhost:5555/scratch_3299_ci .venv/bin/python -m pytest tests/test_0066_celery_river_cutover_postgres.py tests/api/test_main_app_integration.py tests/api/test_work_unit_explain_hardening.py -q` — 22 passed

RISK-NOTES: No behavior change — the 0074 rewrite is semantically identical SQL (SELECT count(*) FROM <table>) expressed via SQLAlchemy Core instead of sa.text(); the four nosemgrep lines suppress only the two named rule ids on the four DDL statements that must stay f-strings (DDL identifiers have no bind-parameter form). Blast radius limited to two already-merged, not-yet-run-in-production migration files (0074/0075, part of the CHAOS-3299 stack). No test-file changes needed since this is a pure refactor/annotation of existing, already-covered migration code (see TEST-EVIDENCE — same 22+4 tests that covered this code before, still pass unchanged). No rollback needed beyond reverting this commit; no follow-up required.

### DIFF
--- a/src/dev_health_ops/alembic/versions/0074_add_ask_dev_v2_persistence.py
+++ b/src/dev_health_ops/alembic/versions/0074_add_ask_dev_v2_persistence.py
@@ -425,7 +425,7 @@ def downgrade() -> None:
         ):
             if bind.dialect.has_table(bind, table):
                 for_v2_rows = bind.execute(
-                    sa.text(f"SELECT count(*) FROM {table}")
+                    sa.select(sa.func.count()).select_from(sa.table(table))
                 ).scalar()
                 if for_v2_rows:
                     raise RuntimeError(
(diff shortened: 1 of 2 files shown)


chrisgeo:
Superseded by #1364 (same fix, correct branch name/title per updated instructions: fix/chaos-3299-semgrep-migration-literals).

PR #1427, opened 2026-08-03
fix(ci): avoid Go integration inventory deadlock

chrisgeo:
## Summary

- remove the newline-scalar/here-string transport that deadlocks Go integration inventory discovery on small local pipe capacities
- store discovered module/package pairs in parallel arrays and iterate them directly
- preserve sorted discovery, denylist validation, fail-on-zero behavior, and per-module integration test grouping
- add a regression through the public `ci/check_go.sh integration-coverage` verb

Linear: https://linear.app/fullchaos/issue/CHAOS-3347/fix-local-go-integration-coverage-hang-in-local-validate

## Target

This repository-wide local-validation fix targets `main` directly. It is independent of the Go Worker Runtime Migration stack because the deadlock blocks every local validation run.

## TEST-EVIDENCE

- baseline `9decbcb1b`: `timeout 5 bash ci/check_go.sh integration-coverage` stops after the inventory header with exit 124
- fixed focused regression passes and the public discovery verb prints 24 discovered, 0 denylisted, 24 runnable packages in about 0.9 seconds
- full escalated `PYTHONPATH=$PWD/src bash ci/check_go.sh fast` exits 0 through Go tests, build, contracts, integration vet, inventory, and advisory
- Bash syntax, ShellCheck, Ruff, mypy, diff check, and mutation-harness clean verification pass
- commit is GPG-signed and independently approved; hosted CI is required before merge

## RISK-NOTES

- the existing `fix/skip-go-local-validate` workaround branch is untouched; this PR fixes the mechanism instead of skipping Go/River stages
- no timeout increase and no integration package denylisting
- no migration 0066, deployment, cutover, scheduler ownership activation, or `goOwnsMarkers` change

SCREENSHOT-WAIVER: local backend CI shell fix; no rendered UI change

### DIFF
--- a/ci/check_go.sh
+++ b/ci/check_go.sh
@@ -313,18 +313,18 @@ check_contract() {
 # "needs a live vendor credential CI does not provision."
 declare -A INTEGRATION_DENYLIST=()
 
-# discover_integration_packages populates INTEGRATION_PACKAGES_BY_MODULE
-# (module_dir -> newline-separated list of module-relative "./pkg/dir"
-# entries) with every package that has at least one tracked or untracked,
-# non-vendor *_test.go file whose leading build-constraint comment names the
-# "integration" tag. This is a repo-wide scan for the literal constraint
+# discover_integration_packages populates parallel module/package arrays with
+# every module-relative "./pkg/dir" entry that has at least one tracked or
+# untracked, non-vendor *_test.go file whose leading build-constraint comment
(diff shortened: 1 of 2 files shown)


chrisgeo:
Temporarily closing after the base/history correction so reopening emits a normal pull_request event and runs the full hosted check set against main.

PR #1450, opened 2026-08-04
feat(ask-dev): CHAOS-3365 sync Linear projects as resolvable subjects

chrisgeo:
## What and why

`PROJECTS_QUERY` and `LinearClient.iter_projects` already existed and **had no caller**. The `projects` catalog therefore only ever held team-derived attribution rows keyed `{org_id}:linear:{team_key}` and named after the *team*. Naming a real Linear project in Ask Dev returned `NO_AUTHORIZED_MATCH`.

The Linear team auto-import now also imports Linear **projects**, keyed by the raw Linear project UUID, stores each project's own lifecycle, and retires them when Linear reports them archived or trashed. One additive ClickHouse migration; no Postgres migration, no new entity kind, no change to any read query.

## Seam verification

| hop | evidence |
|---|---|
| post-sync relay dispatches auto-import | `workers/post_sync_dispatch.py:290`, gated on `auto_import_teams` + `sync_run_id` |
| task resolves credentials, calls the runner | `workers/team_autoimport.py:310`, only for `SyncRunStatus.SUCCESS` |
| runner resolves the Linear populator | `workers/team_autoimport.py:46-58` `_IMPORTER_MODULES["linear"]` |
| new code runs there | inside `populate()`'s existing `LinearClient` block |

**Backfill:** none needed and none added. Existing orgs fill on their next post-sync run.

## Design choices

**Id space — the raw Linear project UUID.** `work_items.project_id` has always carried Linear's native project UUID (`providers/linear/normalize.py:328,355`). Using the same id means a committed project scope selects work items with no query change — the native status SQL already disjoins `project_id = {entity_id} OR project_key = {entity_id}`. A name or a composite would resolve and then match nothing.

**`project_key = NULL`.** A Linear project has no key; `scope_catalog.py:209` already guards with `ifNull(project_key,'')`. It is also what will separate native rows from team-derived ones for the follow-up reconciliation work.

**`is_active` is retirement; lifecycle is its own column.** `scope_catalog.py:214` filters `is_active = 1`, so folding "completed" into inactivity would make a project unresolvable the moment it finished — exactly when people ask what it delivered. Every lifecycle state maps to `1`; only retirement maps to `0`.

**Retirement has three distinct forms and all three are handled.**

| form | mechanism |
|---|---|
| archived | `archivedAt`, visible only because the worker passes `include_archived=True` — the connection omits archived projects by default, so an archived project otherwise just *stops arriving* and can never be retired |
| trashed (soft delete) | `trashed`, a separate flag from `archivedAt`; reading only `archivedAt` left a deleted project resolvable |
| purged (hard delete) | **out of scope for this PR** — see below |

**Lifecycle comes from `status`, not the deprecated `state`.** Linear marks `Project.state` `@deprecated(reason: Use project.status instead)`. That is not cosmetic: removing a field fails the **whole** GraphQL operation, so a deprecated selection would take native project refresh down with it rather than merely losing enrichment. The query selects `status { id name type }` and stores `status.type` — the same closed category the old string carried. `status.name` (the workspace's custom label) is selected but not stored; the ratified column set is three columns and CHAOS-3368 owns surfacing.

**One clock for the version column, stamped at the observation boundary.** `updated_at` is the ReplacingMergeTree version and carries **observation time** for every native row; provider mtime deliberately does not drive it. Two clocks in one ordering column is a defect generator — it was how a stale write could outrank a newer observation — and it is also why the version is taken when a project is actually *seen* rather than at worker start: otherwise two overlapping runs for the same org are ordered by when they **began** rather than by what they **saw**, and the run that started earlier but observed later writes the stale version. Nothing consumes `projects.updated_at` as a real modification time (`scope_catalog`'s watermark uses it only as a cache key), and the team-derived rows already stamped observation time, so this also makes the table internally consistent.

**Retirement is driven only by positive API evidence.** `archivedAt` and `trashed` are facts Linear *states* about a project it still returns. Inferring retirement from a project's **absence** is a different and much harder problem — an absent project is indistinguishable from one a truncated page, a narrowed token or a concurrent run failed to show us. Absence-based reconciliation was built here, attracted a high-severity finding in every subsequent review round, and has been **removed from this PR** and split into its own ticket, whose brief opens with a closure argument over every way a snapshot can be incomplete. What remains never infers anything from silence.

**Completeness is tracked explicitly.** `native_projects_complete` starts **false** and is set true only where the iterator demonstrably exhausted, so a truncated run is visibly degraded rather than looking identical to a healthy one. It no longer gates destruction — nothing destructive remains — but it is what the follow-up will need.

**A degraded page is never mistaken for exhaustion.** Each page is structurally validated: `projects` must be a dict carrying a `nodes` list and a `pageInfo` dict whose `hasNextPage` is an actual **boolean**, or `LinearPaginationIncomplete` is raised. Defaulting these to empty — or accepting `null`/`0`/`""` by truthiness — lets a malformed HTTP 200 or a truncated terminal page look like a clean final page, so the run silently imports part of a workspace and reports it as whole.

### Following the cross-provider precedent — and one deliberate divergence

GitHub milestones (`providers/github/provider.py:200`) and GitLab milestones (`providers/gitlab/provider.py:146`) both fetch `state="all"` rather than only open items, and both store the provider's raw state string without remapping. `include_archived=True` and the verbatim `state` column are the same moves.

Neither writes a tombstone when an item is hard-deleted, and no reconciliation pass exists anywhere in this repo for `sprints`/`work_items` — those rows go stale permanently. **This PR does not close that hole either**; hard-delete reconciliation is split out (see below). Linear projects match the existing cross-provider behaviour here rather than exceeding it.

**Divergence: Linear projects are deliberately NOT mapped to `Sprint`.** Linear *cycles* already own that mapping (`linear_cycle_to_sprint`). Adding projects to `sprints` as well would double-count iterations and corrupt sprint metrics. A Linear project is a subject, not an iteration.

## Schema

`073_linear_project_lifecycle.sql` — additive `ALTER`s adding `state LowCardinality(String) DEFAULT ''`, `target_date Nullable(Date)`, `url String DEFAULT ''`. All defaulted, so existing rows need no backfill. Verified against a **dropped and recreated** database, so it applies from clean rather than only as a delta.

Both writers were updated — `metrics/sinks/clickhouse/core.py` `write_projects` and `storage/clickhouse.py` `insert_projects`. No third writer exists.

## Blast radius

Swept `src/`, `internal/`, `cmd/`, `contracts/`, `tests/`, `docs/`, `scripts/`, `alerts/`, `deploy/`, including indirection through table-name constants and generic dimension-write paths.

- `api/dev/scope_catalog.py:207-218` is the only reader in the API; `ifNull(project_key,'')` tolerates NULL.
- Watermark (`maxOrNull(updated_at)`) is a max — extra rows only move it forward.
- No SQL join on `project_key` anywhere; `team_project_ownership.project_id` joins only to `teams`, never to `projects.id`.
- Nothing parses `projects.id` — the only colon-splitting in the repo is on `work_item_id`.
- Nothing counts rows in `projects`.

## Known limitation (deferred — CHAOS-3369)

The pre-existing team-derived rows match **no** work items, because Linear work items never carry a `project_key`. So a team/project name collision returns `AMBIGUOUS_CANDIDATES` with one dead-end candidate. I originally argued this was the correct contract for an ambiguous mention; review refuted that — one candidate is never valid, so it is a real degradation. No collision exists in the dev workspace today. Consolidating the id spaces is **CHAOS-3369**. The behaviour is pinned by a test, and the oracle reports it as a `name_ambiguous` gap rather than passing over it.

## Differential oracle

`scripts/ask_dev_project_subject_oracle.py` compares two independent real producers — the work-item sync's project references against the catalog the auto-import writes — asserting every referenced project exists, under the same name, active, and unambiguous. Read-only.

It refuses to pass when it did not measure: an empty reference side and a missing acceptance-set project both raise `OracleNotMeasured` (exit 2), and the compared field set is checked against the real `ProjectRecord`. It documents what it does *not* check (it does not run the async resolver; it has no freshness floor) so a green run is not read as more than it is.

**Live run against dev ClickHouse (read-only, `default` never written):**

```
16 gap(s) between referenced projects and the subject catalog:
  [missing] 13e65c04-40ec-4a95-8216-f7c2ce233244 'Ask Dev' — no projects row; the project cannot resolve as a subject
  … 15 more …
```

The acceptance-set entry is rediscovered as missing, and the oracle runs correctly against the **unmigrated** dev schema, confirming it stays read-only and schema-compatible. The post-fix side is proven on a scratch database; dev cannot show it until a real Linear sync runs with this code.

## Test plan

**RED first, observed.** With `src/` reverted to `origin/main`, the unit file fails 19/19 and the live file fails with `resolve_mention("Ask Dev")` returning `NO_AUTHORIZED_MATCH`.

**Fixtures generated from the producer.** The GraphQL page is not hand-authored — the node shape is parsed out of `PROJECTS_QUERY` and the builder **fails** if the query requests a field it does not model. Only `LinearClient._execute` is replaced, so `iter_projects`, its pagination loop and the query text are production code.

**26 mutations; 24 killed, 2 documented as behaviour-neutral.** Restores verified by content digest after every run — not by `git diff`, not by "it still imports". One batch harness run under-reported its results, so both high-severity fixes were re-verified individually rather than trusted. Highlights:

| mutation | died in |
|---|---|
| id = project name instead of UUID | id-space guard + live scope-selects-work-items guard |
| `is_active` keyed on lifecycle | completed/canceled guards + live completed-project guard |
| worker stops requesting archived projects | `the_worker_actually_asks_linear_for_archived_projects` |
| `trashed` ignored | trashed + delete-between-runs guards |
| revert to deprecated `Project.state` | the whole unit file (the fixture refuses to build) |
| pagination cursor guard removed | repeated-cursor and missing-cursor guards |
| completeness defaults true again | the client-setup-failure guard |
| `hasNextPage` checked for presence, not type | 4 non-boolean cases (`null`/`0`/`""`/`[]`) |
| version stamped at worker start, not observation | `rows_are_versioned_when_observed_not_when_the_worker_started` |
| page validation removed | producer-level `iter_projects_rejects_a_malformed_page_deliberately` |
| real sink drops the `state` column | live column round-trip test *only* |

**A correction to an earlier claim in this PR.** I had recorded the surviving `isinstance(projects_data, dict)` mutation as a redundant clause. Review challenged that, and re-running the mutation shows it now **dies** against three producer-level cases — my note was true only against the older test set and stopped being true once the exception-type test landed. The claim was wrong and is retracted here rather than left standing, because an inaccurate coverage claim is worse than an admitted gap.

**Six tests in this changeset would have shipped green while proving nothing.** None were visible from a passing suite:

1. The archived-project test fed a node the production connection is configured never to return — the reason `include_archived=True` exists.
2. The moved-work-item oracle test passed regardless, because ClickHouse merges the two ReplacingMergeTree parts within about a second — fixed with `SYSTEM STOP MERGES` plus a two-rows-on-disk precondition.
3. A live fixture asserted a `target_date` it never wrote — found by adding the real column round-trip test.
4. `native_projects_complete` defaulted true, so a client-construction failure reported "0 projects, complete" — which reads as *this workspace has no projects*.
5. `test_retirement_is_skipped_when_the_catalog_cannot_be_read` asserted a zero that was zero for two different reasons; the claim it protected against a wipe was false and was removed rather than justified.
6. The first live retirement test hand-built both records, so it proved RMT ordering while staying green if reconciliation were deleted outright.
7. The malformed-page tests all *omitted* `hasNextPage`, so none supplied a wrong-typed falsey value — `null`, `0` and `""` were still accepted as a clean terminal page. A false-pass surface inside the fix for the previous false-pass surface.

**Live ClickHouse coverage** proves what fake sinks structurally cannot: retirement through the **production** sink read back via `projects FINAL`, parametrized over **both arrival orders** (insert order is not delivery order), with the real resolver then returning `NO_AUTHORIZED_MATCH`; and every new column round-tripping, which is the only test that catches a column dropped from the real insert list.

**Test wiring.** `ci/run_tests.sh` already filters `-m "not benchmark and not clickhouse"` in **both** `unit_tests()` (line 161) and `ci_tests()` (line 247). No change needed; the live file is opt-in via `pytest -m clickhouse`.

### Honest gate note

`ci/local_validate.sh` was **not** run — it wedges at the live-argMax stage on this machine (known, deferred). Run instead, with the gate's literal commands:

- `ruff check src scripts tests` — clean
- `ruff format --check src scripts tests` — clean
- `mypy --install-types --non-interactive .` — **Success, 2044 source files** (the gate's exact `gate_typecheck`)
- scoped pytest across `tests/workers/`, `tests/storage`, `tests/providers`, scope-catalog, linear-provider, attribution-schema, org-id — **1447 passed, 2 skipped**
- `CLICKHOUSE_URI=…/chaos3365_scratch pytest tests/test_ask_dev_linear_project_subject_live.py -m clickhouse` — **19 passed**, against a dropped-and-recreated database

This is a scoped run, not the full unit suite. CI is the first place the whole suite executes against this branch.

## Scope split: absence-based retirement

Hard-delete reconciliation — retiring a project because Linear stopped returning it — was implemented here and has been **removed**. It drew a high-severity finding in every review round after it appeared: retirement gated on a completeness flag that could lie; a malformed page accepted as clean exhaustion; a non-boolean `hasNextPage` accepted the same way; a narrowed token returning a partial-but-well-formed snapshot; and a concurrent-run version race.

Those are all one defect class — **an incomplete view of the world authorising destruction** — and each fix closed one cell rather than the class. It moves to its own ticket, whose brief must open with a closure argument: partition every way a snapshot can be incomplete (malformed page, truncated page, narrowed token, concurrent run, provider omission) and show which mechanism covers each cell. Per-org serialisation is flagged there as the open infrastructure question.

Nothing in this PR infers anything from absence. Archival and trashing are positive statements Linear makes about projects it still returns, and they cover the common cases.

## Review

Five rounds of adversarial review, each of which found real defects rather than nits: archived projects never retiring, and the live suite able to run destructive migrations against the dev database; a dependency on a deprecated Linear field that would have failed the entire query on removal, plus the separate `trashed` deletion path; a reproduced completeness-flag bug; mixed version clocks; and a non-boolean `hasNextPage` accepted as exhaustion. Every fix carries a mutation proving its guard fails when the fix is reverted.


chrisgeo:
## Round-4 review disposition (all four findings closed)

Reviewed `0c36a7fe0`; branch is now `7eebe0089`.

**HIGH — wall-clock tombstones could make a false retirement permanent.** Correct, and the sharpest finding in the whole review cycle. Tombstones were stamped with the wall clock while active rows carried Linear's `updatedAt`. After any false absence, the project's next healthy appearance arrives with an *unchanged, older* mtime, loses to its own tombstone under `FINAL`, and stays retired forever — a transient glitch made permanent.

Fixed by putting every native project row on **one clock**: `updated_at` is now observation time for active rows and tombstones alike. Provider mtime no longer drives the ReplacingMergeTree version, so the most recent look at the world always wins. `_PROJECT_VERSION_FLOOR` and the timestamp-parsing fallback are deleted — that entire failure class disappears with the mixed clocks that created it.

Proven by `test_a_project_that_reappears_supersedes_its_own_tombstone`, which runs `populate` three times against real ClickHouse (import → absent → returns) and asserts `FINAL` flips back to `is_active = 1`. Reverting to provider-mtime versioning kills it with `(0,) != (1,)`.

**HIGH — malformed responses accepted as complete snapshots.** Correct. `iter_projects` defaulted a missing `projects`/`nodes`/`pageInfo` to empty and read absent `hasNextPage` as clean exhaustion, so a malformed HTTP 200 — or a truncated terminal page — set `native_projects_complete = True` and authorised retirement of everything it omitted. The empty-response guard did **not** cover this: a run that returned some real projects looks healthy.

Each page is now structurally validated and raises `LinearPaginationIncomplete` unless `projects` is a dict carrying a `nodes` list and a `pageInfo` dict that actually contains `hasNextPage`. Covered at the worker level across six malformed shapes, plus a producer-level test asserting the **exception type** across eight — because the worker-level assertions observe only the outcome, which any exception reaches, and so could not tell deliberate validation from an accidental `TypeError`.

**MEDIUM — tombstone destroyed last-known lifecycle.** Already fixed before this round ran (the review notes it excluded unstaged changes). Tombstones copy the full prior row and change only `is_active`/`updated_at`/`last_synced`; the live end-to-end test asserts both the label and `state` survive retirement.

**MEDIUM — the live retirement test bypassed the production path.** Correct, and the sixth false-pass surface in this changeset. The prior test hand-built both records, so it proved RMT ordering while staying green if reconciliation were deleted outright. `test_populate_retires_a_purged_project_end_to_end` now runs `populate` twice against real ClickHouse, constructs no record itself, and asserts the tombstone, the preserved lifecycle, `NO_AUTHORIZED_MATCH` from the real resolver, and that a still-present project is untouched. Deleting the reconciliation block kills it.

### Two mutations survived, and are documented rather than papered over

- Changing the unreadable-catalog return from `None` to `{}` is **behaviour-neutral** — absence is computed against that mapping, so an empty one retires nothing either. The docstring that had claimed this was a safety mechanism was removed rather than a guard invented to justify it. The real protection is the empty-response guard, which is separately tested.
- Removing the `isinstance(projects_data, dict)` clause alone survives because the sibling `nodes`/`pageInfo` checks catch every shape. Kept for clearer errors; the property is proven by clause-level mutations that do die.

### Verification

26 mutations: 24 killed, 2 documented as behaviour-neutral above. Restores verified by content digest after every run. One batch harness run under-reported results, so both HIGH-fix mutations were re-verified individually rather than trusted.

`ruff` clean · `mypy` clean (2044 files) · 1447 scoped unit tests · 19 clickhouse-marked tests against a **dropped and recreated** scratch database, so migration `073` is proven to apply from clean.

chrisgeo:
## Round-5 disposition — 2 of 3 closed, 1 open pending a decision

Branch is now `749942241`. **Not mergeable yet** — finding 3 below is a real design decision, not a fix I should make unilaterally.

### Closed

**HIGH — a falsey non-boolean `hasNextPage` still authorised truncated retirement.** Reproduced by the reviewer with `None`, `0` and `""`. The validation I added last round checked that the key *existed*, then tested it by truthiness — so a malformed response carrying a genuine prefix plus `hasNextPage: null` marked enumeration complete and authorised retiring everything it omitted. Every one of my six malformed shapes *omitted* the key, so none could catch it. This was a false-pass surface inside the fix for the previous false-pass surface.

Now requires `isinstance(has_next, bool)`, with producer-level cases for `null`, `0`, `""`, `"false"`, `"true"`, `1` and `[]`, plus a worker-level test proving a truthy prefix with a bogus flag cannot retire anything. Reverting to the presence-only check kills four of them.

**HIGH — start-time versioning could let a newer observation lose.** `now` was captured before team discovery and enumeration but became every native row's version, so two overlapping runs for the same org were ordered by when they *started* rather than by what they *saw*. A run that began earlier and observed a project later wrote a stale version and lost to a concurrent run's tombstone, keeping a live project retired.

Native rows and tombstones are now stamped at the observation boundary. A test makes enumeration take measurable time and asserts the version postdates the run's start; reverting to worker-start time kills it. Full per-org serialisation is infrastructure beyond this ticket and is **not** claimed — this is the recommended minimum, and it removes the specific interleaving described.

### Open — needs a call

**HIGH — a visibility-reduced but non-empty snapshot can still retire valid projects.** The empty-response guard only covers the all-empty case. A narrowed token that returns *some* projects passes it, and every missing active project is tombstoned after that single run. Structural pagination validity proves the cursor exhausted, not that the whole workspace was visible.

Two options, and this is a product decision rather than a bug fix:

1. **Add a multi-run grace rule** — a project must be absent across more than one healthy enumeration before retirement. Implementable off the existing `last_synced` column with no schema change.
2. **Split absence-based retirement into its own PR** and land the rest now. Retirement is the piece attracting every high-severity finding, and archival/trashing — which cover the common cases — are unaffected either way.

Consequence if shipped as-is: a transient scope reduction wrongly retires projects for one cycle. It is **self-healing** rather than permanent, because of the observation-clock fix, but it is user-visible while it lasts.

### Correction to an earlier claim of mine

I previously recorded the surviving `isinstance(projects_data, dict)` mutation as a redundant clause. The reviewer challenged that, and re-running the mutation shows it now **dies** against three producer-level cases. My original note was true only against the older test set; it stopped being true once the exception-type test landed, and I did not re-check. The PR body has been corrected — an inaccurate coverage claim is worse than an admitted gap.

### Verification

29 mutations: 28 killed, 1 documented as behaviour-neutral (the unreadable-catalog `None`-vs-`{}` return). `ruff` clean · `mypy` clean (2044 files) · 1456 scoped unit tests · 19 clickhouse-marked tests.

chrisgeo:
## Scope split applied — absence-based retirement removed. Ready to merge.

Branch is `a2c2e26c1`. The HOLD is lifted.

### What changed

**Removed:** hard-delete reconciliation — `_existing_native_projects`, `_may_retire_by_absence`, `_absence_tombstones`, the catalog read-back query, the `native_projects_retired` summary key, and every test that exercised them.

**Kept:** retirement driven by *positive API evidence*. `archivedAt` and `trashed` are facts Linear states about a project it still returns; nothing in this PR now infers anything from a project's silence. The `include_archived=True` fetch stays, because without it an archived project simply stops arriving and can never be retired at all.

**Also kept, because they harden what remains:** type-strict `hasNextPage` validation, structural page validation, observation-boundary versioning, and the explicit completeness signal. Completeness no longer gates destruction — there is none — but it makes a truncated run visibly degraded instead of indistinguishable from a healthy one, and it is the signal the follow-up will need.

### Why the split is the right call

Absence-based retirement drew a high-severity finding in *every* review round after it appeared: a completeness flag that could lie; a malformed page read as clean exhaustion; a non-boolean `hasNextPage` read the same way; a narrowed token returning a partial-but-well-formed snapshot; and a concurrent-run version race.

Those are one defect class — **an incomplete view of the world authorising destruction** — and each fix closed a single cell. A grace window would have been one more cell. The follow-up brief should open with the partition instead:

| how a snapshot can be incomplete | candidate mechanism |
|---|---|
| malformed page | structural + type validation (already landed here) |
| truncated page / stalled cursor | completeness signal (already landed here) |
| narrowed token or reduced visibility | **open** — needs authority beyond "the cursor exhausted" |
| concurrent/overlapping runs | **open** — per-org serialisation or fencing (infrastructure) |
| provider omits a record without saying so | **open** — needs a count or digest cross-check |

The first two are done. The last three are what the new ticket has to close before any absence inference is safe.

### Correction to an earlier claim of mine

I recorded a surviving mutation as a redundant clause; review challenged it and re-running shows it **dies** against three producer-level cases. My note was true only against the older test set and stopped being true once the exception-type test landed. Retracted in the PR body — an inaccurate coverage claim is worse than an admitted gap.

### Verification at `a2c2e26c1`

Re-mutated the retirement that stays, after removing its sibling, to confirm it is still guarded: dropping the `archivedAt`/`trashed` predicate kills four tests; ignoring `trashed` alone kills two; dropping `include_archived=True` kills the request-shape test; presence-only `hasNextPage` kills four non-boolean cases.

`ruff` clean · `mypy` clean (2044 files) · **1449** scoped unit tests · **17** clickhouse-marked tests against a dropped-and-recreated database.

chrisgeo:
Rebase note: main now carries the go-toolchain stderr fix (#1451); re-running your failed jobs after updating the branch picks it up.

PR #1452, opened 2026-08-04
fix(ask-dev): project status snapshot reads the committed subject's work items

chrisgeo:
## Symptom

Ask Dev, live, all merged code: *"What is the status of the Ask Dev project, and what are looking like the current blockers?"* The wave-3.1 pipeline resolved the subject correctly — `dev_runs` row `5425c82c-…` has `preflight_outcome=proceeded_committed_subject` against Linear project `13e65c04-…` "Ask Dev" — and then answered with a vague partial instead of the PRD §10 verdict/completion/blockers.

The `status_snapshot` source observation came back `data_semantics=no_data`, `observed_state=available_unknown`, every content list empty — while ClickHouse `work_items` holds **69 rows** for that project, all updated inside the run's window.

## Root cause

`native_status_change.py` — `_authorized_repository_ids` fell through to `return self._repository_ids(scope)` for a PROJECT scope, and that always returns the empty set, because **a committed project subject carries no repository dimension anywhere on the wire**:

- `scope_catalog.py:213` — the PROJECT catalog query selects `NULL AS repository_id`, so `AuthorizedEntity.repository_id` is `None` for every project (a project spans repositories; it has no repository of its own).
- `scope_service.py:777` (`committed_scope_for`) and `scope_service.py:1169` (`_resolved_scope`) populate `DevScope.repositories` only when the entity is a REPOSITORY.

Both inputs to `_repository_ids` are therefore empty, so `status_snapshot` hit its fail-closed guard (`native_status_change.py:974-981`) and returned `declared=None` with a single `authorized_repositories` UNAVAILABLE ref **before issuing a single project query**. `change_summary` shares the helper and did the same.

This is the same structural problem ORGANIZATION (CHAOS-3255) and TEAM (CHAOS-3303) already have dedicated branches for. PROJECT never got one.

`work_graph_expansion` being skipped in the same run is **not** this bug and is correct as designed: `builtin_steps.py:904-916` `_work_graph_applicable` admits only issue/pull_request refs, and the live observation is `not_applicable` / `not_measured` / `limitation=step_not_applicable`, requirement_level *conditional*. The unavailable **required** source was the mandatory `status_change` one.

## Fix

A PROJECT arm in `_authorized_repository_ids` that re-derives the repository set from canonical work-item project attribution (`_PROJECT_REPOSITORIES_SQL`), bounded by:

- the same `org_id` tenant boundary `_ORGANIZATION_AUTHORIZED_REPOSITORIES_SQL` enforces;
- the same `updated_at <= as_of` bound `_WORK_ITEMS_SQL` applies;
- the **byte-identical** `project_id`/`project_key` disjunction every project SQL arm uses, so the derived set can never admit a repository the fact queries it bounds would not themselves have drawn from (pinned by a test);
- the org-scoped `repos` catalog, **plus** an explicit arm for the zero UUID — not a repository but the sentinel the sink writes for repository-less records (`row.repo_id or uuid.UUID(int=0)`), which is every Linear work item.

A **team-filtered** project scope is deliberately excluded from the derivation and keeps failing closed, exactly as the team-filtered organization case does: no project SQL arm applies a team filter, so deriving the project's full repository set would answer "project P, team A only" with team B's work.

## Verification

**RED → GREEN.** New `tests/api/dev/test_project_scope_status_snapshot_repositories.py` builds the scope with the **real producer** (`ScopeResolutionService` over the real `ClickHouseAuthorizedEntityCatalog`), not hand-built. Every pre-existing project-scope test hand-builds a scope carrying `repository_id="repo-a"` — a shape production cannot emit — which is why they were all green while production failed closed. With the source reverted: 2 failed / 3 passed. With the fix: 11/11.

**Mutation: 14/14 killed** — drop the PROJECT branch; drop the `project_id` arm; drop the `project_key` arm; `OR` → `AND`; drop the `as_of` bound; drop the tenant bound on `work_items`; drop it from the `repos` sub-select; drop the `repos` authorization arm; drop the zero-UUID sentinel arm; drop the team-filter guard; inject a `LIMIT`; except-arm fabricates a set; drop the empty-id guard; pass the wrong id in.

**Live dev ClickHouse, read-only.** The corrected SQL `EXPLAIN`s cleanly against the production schema and returns `['00000000-0000-0000-0000-000000000000']`; feeding that into the real `_WORK_ITEMS_SQL` returns all **69 rows** (38 done / 20 backlog / 6 unknown / 4 in_progress / 1 canceled), with the blocker watermark resolving and **155** native `blocks` edges landing on the project's items. 0 linked PRs — Linear-only project, so the answer is completion + blockers, no delivery evidence.

**Suites.** `tests/api/dev` + `tests/api/graphql` + `tests/acceptance`: 2311 passed, 2 failed — both (`test_resolver_sql_explain.py::…[operating_review]`, `[analytics]`) reproduce identically on a clean `origin/main` and are untouched here. ruff + mypy clean.

## Adversarial review

Codex returned `needs-attention` with 4 HIGH + 1 MEDIUM. Two were real defects in this change and are fixed in the second commit (team-filtered widening; bypassing the `repos` authorization catalog). Three are verified pre-existing and **deliberately not fixed here** — reported rather than half-fixed:

1. **Jira project subjects still fail closed.** `team_autoimport_jira.py:106` mints the catalog id as `f"{org_id}:jira:{project_key}"`, while `providers/jira/normalize.py:517-518` writes the raw Jira id/key onto `work_items`. Neither `project_id = {entity_id}` nor `project_key = {entity_id}` matches. **`_WORK_ITEMS_SQL`, `_BLOCKERS_SQL` and `_PULL_REQUESTS_SQL` already use that identical predicate**, so Jira project subjects are broken end-to-end before and after this change — making only the derivation Jira-aware would derive repositories and then have every fact query return nothing. Needs a coordinated change across all project arms.
2. **Mandatory project data-health widens to the whole organization.** `data_health_service.py:381-388` uses `AND (empty({repository_ids}) OR …)`, so an empty list disables repository filtering, and `:195` then falls back to the full org repo set as the denominator. A committed project scope still reaches it with no repositories, so unrelated healthy repositories can make a project's evidence coverage read complete. Separate service, separate semantics (widening, not fail-closed).
3. **The unit fake is a predicate evaluator, not a SQL engine.** Acknowledged in its own docstring rather than papered over. The two properties it provably cannot see — boolean precedence and `LIMIT` — are pinned structurally instead, and `_PROJECT_REPOSITORIES_SQL` now has a real-engine `EXPLAIN` entry in the live inventory. Fully seeded multi-provider live coverage is follow-up work.

Pre-existing and unrelated, flagged in passing: `tests/api/dev/test_status_change_clickhouse_live.py` fails 10/13 on `main` because its shared fixture omits the `member_issue_ids` substitution.

## Out of scope, answered

`dev_run_resolutions` is empty for all runs, and that is **expected design, not a persistence gap**: `subject_preflight.py:239-247` documents `terminating_resolution_entry` as `None` for every termination reason other than an ambiguous mention and for every non-terminal result; `orchestrator.py:1209-1211` is the only production caller of `append_resolution` and sits inside the TERMINATE branch gated on that field; `persistence/service.py:1059` reads the table filtered to `outcome == "ambiguous_candidates"`. A `proceeded_committed_subject` run correctly writes nothing. (Same shape for `dev_run_subject_sets`: `subject_preflight.py:727-735` only mints an audit set when `len(mentions) > 1`.)

### DIFF
--- a/src/dev_health_ops/api/dev/native_status_change.py
+++ b/src/dev_health_ops/api/dev/native_status_change.py
@@ -468,6 +468,38 @@ FROM (
 WHERE g.repo_id IS NOT NULL
 """
 
+#: Re-derive one project's repository set from canonical work-item attribution.
+#:
+#: A committed PROJECT subject can never carry a repository list of its own,
+#: exactly like the TEAM subject above: ``DevScope.repositories`` is only ever
+#: populated for a REPOSITORY commit
+#: (``scope_service.ScopeResolutionService.committed_scope_for`` /
+#: ``_resolved_scope``), and the catalog row a project commit is built from
+#: selects ``NULL AS repository_id``
+#: (``scope_catalog.ClickHouseAuthorizedEntityCatalog._query_for``) because a
+#: project spans repositories and has no repository dimension. So
+#: ``_repository_ids(scope)`` returned an empty set for *every* production
+#: project subject, and ``status_snapshot``/``change_summary`` took the
+#: fail-closed "authorized repository set" branch before running a single
+#: project query -- an empty snapshot for a project whose work items match.
+#:
+#: Bounded by the same ``org_id`` tenant boundary
+#: ``_ORGANIZATION_AUTHORIZED_REPOSITORIES_SQL`` enforces, by the same
+#: ``updated_at <= as_of`` bound ``_WORK_ITEMS_SQL`` applies, and by the same
+#: ``project_id``/``project_key`` disjunction every project SQL arm uses -- so
+#: the derived set is exactly the repositories this project's own facts can be
+#: drawn from, never wider. Deliberately NOT joined against ``repos``: Linear
+#: work items carry no repository and land under a zero-UUID ``repo_id`` that
+#: has no ``repos`` row, so an existence join would re-empty the set for the
+#: provider this fix exists for.
+_PROJECT_REPOSITORIES_SQL = """
+SELECT DISTINCT toString(repo_id) AS repository_id
+FROM work_items FINAL
+WHERE org_id = {org_id:String}
+  AND updated_at <= {as_of:DateTime64(3, 'UTC')}
+  AND (project_id = {entity_id:String} OR project_key = {entity_id:String})
+"""
+
 _TRANSITIONS_SQL = (
     """
 SELECT transition.work_item_id AS entity_id, item.title AS display_label,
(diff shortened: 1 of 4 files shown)


chrisgeo:
Deferred defects tracked: CHAOS-3374 (Jira project identity end-to-end), CHAOS-3375 (project data-health widening), CHAOS-3376 (broken live EXPLAIN fixture on main).

PR #1493, opened 2026-08-05
fix(ops): CHAOS-3399 make the tests dataset selectable and repair existing integrations

chrisgeo:
CHAOS-3399: Make the `tests` dataset selectable and repair existing integrations

## What changed

1. **Repair path for existing integrations.** `PATCH /sync-configs/{id}` (the endpoint the web edit form calls) now retro-creates/re-enables the corresponding `IntegrationDataset` row(s) whenever `sync_targets` is updated. Previously this endpoint only updated the legacy `SyncConfiguration.sync_targets` Postgres column -- the planner (`_load_enabled_datasets`) reads `IntegrationDataset` rows, never `sync_targets`, so an operator who added `tests` to an *existing* integration's targets saw the checkbox flip on with zero effect (CHAOS-3398's root cause).
   - **Bidirectional** (added after a codex adversarial-review finding, see below): unchecking a dataset now disables the row too, matching what the edit form's own copy already promised (`formDiff.ts::getDatasetWarnings`: "will stop syncing").
   - **Scoped to the whole-integration config only** (added after a second codex finding, see below): `IntegrationDataset` rows are shared across every `SyncConfiguration` pointing at the same `integration_id` (a planner parent plus its per-repo children, CHAOS-2762). Only the config with no `source_id` -- the one representing the whole integration -- may mutate the shared rows, so editing one repo's child config can never silently disable/enable a dataset for its siblings.
2. **Operator-visible empty-state signal.** The sync coverage payload now distinguishes a dataset that has no enabled `IntegrationDataset` row (`status: "not_enabled"`) from one that's enabled but has produced zero rows so far (`status: "insufficient_data"`). Both used to render identically (an absent row), which is exactly what let TestOps being dark hide for months. Additive-only: the new `build_coverage_summary_payload` kwarg defaults to empty, inert for every existing caller/test.
3. **Confirmed already-correct, pinned with regression tests:** `GET /sync-targets`, `get_dataset_spec`/`_planner_dataset_keys`, and the admin batch-update endpoint (`PATCH /integrations/{id}/datasets`) already supported/worked for `tests` on github/gitlab -- that half of CHAOS-3398 was already complete, just unexercised and untested.

## Adversarial review findings (both fixed, both verified)

Two rounds of `codex:adversarial-review` on this branch surfaced real issues, not cosmetic ones:

- **[medium] Enable-only reconciliation was a one-way toggle.** The first version only created/re-enabled rows, never disabled them on removal -- an operator who unchecked `tests` would see the UI say it stopped, while the planner kept running it. Fixed by making the reconciliation bidirectional (item 1 above), verified by temporarily reverting the fix and observing the new regression test fail, then restoring it and observing it pass.
- **[high] Shared-integration blast radius.** The bidirectional version above then let ANY config with an `integration_id` mutate the shared `IntegrationDataset` rows, including a source-scoped *child* config representing just one repo. Since the planner (`_load_enabled_datasets`) always reads the shared row set for the whole integration, one repo's edit could silently disable a dataset for every sibling config sharing that integration. Fixed by scoping the reconciliation to `source_id is None` configs only (item 1 above), verified the same way (guard removed → new regression test fails; guard restored → passes) plus a dedicated regression test reproducing the exact shared-integration scenario (a parent + a source-scoped child).

A third confirmatory adversarial-review pass on the rebased final diff hit a shared-machine `codex` CLI/config issue (`model_reasoning_effort=max` unsupported by the resolved model variant) unrelated to this changeset -- the two substantive rounds above already exercised and fixed the real findings.

## Contract re-anchoring (`contracts/jobs/v1/transitional-inventory.json`)

This diff shifted line numbers for 4 already-tracked call sites in `sync.py` (twice, as the fix grew across the two review rounds): the `POST /sync-configs/{config_id}/trigger` and `.../backfill` endpoints, and their `getattr(dispatch_sync_run, "apply_async")` dispatch call sites. **Same sites, new offsets -- not new or changed surfaces.** Confirmed each time by reading the new line numbers directly and matching them against the pre-existing row's `surface`/`dispatches` fields before re-anchoring, then re-running `ci/check_transitional_inventory.py` to confirm zero UNOWNED/PHANTOM/STALE violations remain.

## Out of scope (per issue, and per a sibling finding)

- No auto-enable rule for `tests` (stays strictly opt-in).
- Did not touch `_ensure_security_dataset_for_scheduled_code_host_sync` (CHAOS-3400's territory) or CI workflow files (CHAOS-3401's territory).
- Did not raise `MAX_RUNS_PER_SYNC` / `MAX_ARTIFACTS_PER_RUN` / `MAX_REPORTS_PER_RUN`, or `SYNC_BUDGET_BUCKET_LIMITS`.
- `coverage_snapshots` / `/testops/coverage` remaining sparse is CHAOS-3401's territory, not a regression here.

## ⚠️ Known material limitation -- filed as CHAOS-3412 (priority 1)

**This fix makes `tests` selectable and the repair path structurally correct, but a scheduled (incremental) sync of a newly-enabled `tests` dataset on an org with a wide `initial_sync_depth` can defer forever and never actually run.** Reproduced live: this org's `initial_sync_depth=90` days pushes `tests`'s REST_CORE budget estimate (`4 × span_days` = 360) over the `SYNC_BUDGET_BUCKET_LIMITS` cap (250). Because there's no watermark yet (first-ever sync), every incremental attempt recomputes the same oversized window and gets deferred again -- and budget-guard deferrals never increment `attempts`, so the reconciler's retry-exhaustion predicate (`retry_count >= max_retries`) is unreachable by construction. There is no exit at either end. **This is not `tests`-specific**: `commit-stats` (4×) and `files` (3×) carry the same coefficients and would hit the identical wall if enabled fresh on an org like this one. CHAOS-3412 has the full mechanism. The live verification below only reached `success` by sidestepping the scheduler with an explicit, narrow-window backfill -- an operator who just flips the toggle and waits will not see the dataset populate until CHAOS-3412 lands.

## Verification (live, against local Docker -- not just unit tests)

Ran the actual admin FastAPI app in-process (unmodified production code, real dependency wiring) against the real dev Postgres/ClickHouse, dispatching onto the same Celery broker the running `dev-health-worker-1`/`dev-health-worker-heavy-1` containers consume from -- the write path and the actual GitHub sync executed for real.

| Step | Result |
| --- | --- |
| `GET /sync-targets` includes `tests` for github/gitlab | confirmed live |
| Baseline: the local GitHub integration has no `tests` `IntegrationDataset` row | confirmed |
| `PATCH /sync-configs/{id}` with `tests` added to `sync_targets` | 200; `tests` row created, `is_enabled=true` |
| Trigger a sync scoped to `dataset_keys=["tests"]` (incremental mode, no watermark yet) | 202 accepted, 4 units planned -- **all 4 permanently stuck in `retrying`** (`deferred by sync budget guard`, `attempts=0`), reproducing CHAOS-3412 live |
| A narrower, explicit 2-day backfill (`POST /integrations/{id}/backfill`) for the same dataset | 202 accepted, 4 units planned, **all 4 reached `success`** |
| `ci_job_runs` (cheapest positive control) | 28 → 989 for the org |
| `test_case_results` / `test_suite_results` | 0 → **195907** / **822** |
| `testops_test_metrics_daily` | ran `dev-hops metrics daily`; 0 → 8 rows, e.g. `total_cases=138944, pass_rate=0.9919` |
| `/testops/tests` (real browser, canonical `admin@devhealth.example` account) | renders **Pass Rate 99%, Failure Rate 0%, Flake Rate 0%, P95 Suite Duration 2.5m** -- matches ClickHouse exactly |
| `sync_watermarks` | correctly absent for the backfill-mode run (by design -- backfill never fast-forwards the incremental cursor past data it didn't cover incrementally; not a defect) |
| **Negative control** (dataset disabled → trigger) | `total_units: 0` -- confirmed live, both before and after re-enabling |

## Rebased onto two PRs that landed mid-flight

`origin/main` moved twice while this branch was in flight and includes both:
- **#1491** (CHAOS-3400, security opt-in) -- removes `_ensure_security_dataset_for_scheduled_code_host_sync` from `sync/planner.py`. This lane's round-2 fix reasons directly about `_load_enabled_datasets` in that same file, so rather than assume the rebase was mechanical, I read the merged `planner.py`: `_load_enabled_datasets` is byte-for-byte unchanged, just shifted from line 549→493 by the removed function above it. The mental model this PR's fix depends on (the planner reads `IntegrationDataset` rows shared by `integration_id`, filtered by `is_enabled` and optionally narrowed by `dataset_keys`) still holds exactly.
- **#1492** (CHAOS-3406, coverage-parser ENAMETOOLONG guard) -- unrelated file (`parsers/coverage.py`), no interaction.

Re-ran `ci/check_transitional_inventory.py` and the affected test files (199 passed) after each rebase to confirm.

## Gates

- `bash ci/local_validate.sh` -- **GATE PASSED** on the final rebased branch (ruff format/check, mypy, ClickHouse scratch-db migrate, full unit suite -- 11671 passed, live-ClickHouse argMax proof).
- Note: an unrelated, pre-existing environment issue (CHAOS-3402: direnv auto-loads a shared, non-worktree `ops/.env` into any shell regardless of `cd`, leaking ~64 vars including `DATABASE_URI`/`JWT_SECRET_KEY`) causes ~16 unrelated test failures (credential-resolver, auth-registration, org-deletion, cli-preflight) unless those vars are explicitly unset before running the gate. Confirmed those failures are pre-existing and unrelated by reproducing them on unmodified `main` and by seeing them disappear once the vars are scrubbed.

### DIFF
--- a/contracts/jobs/v1/transitional-inventory.json
+++ b/contracts/jobs/v1/transitional-inventory.json
@@ -2529,12 +2529,12 @@
       "notes": "grep-evading form; POST /integrations/{integration_id}/backfill"
     },
     {
-      "id": "call_site_getattr_indirection:src/dev_health_ops/api/admin/routers/sync.py:2333",
+      "id": "call_site_getattr_indirection:src/dev_health_ops/api/admin/routers/sync.py:2385",
       "surface": "getattr(dispatch_sync_run,'apply_async')",
       "class": "call_site_getattr_indirection",
       "source": {
         "file": "src/dev_health_ops/api/admin/routers/sync.py",
-        "line": 2333
+        "line": 2385
       },
       "queue_or_cadence": "sync",
       "dispatches": "dispatch_sync_run",
@@ -2553,12 +2553,12 @@
       "notes": "grep-evading form; POST /sync-configs/{config_id}/trigger"
     },
     {
-      "id": "call_site_getattr_indirection:src/dev_health_ops/api/admin/routers/sync.py:2478",
+      "id": "call_site_getattr_indirection:src/dev_health_ops/api/admin/routers/sync.py:2530",
       "surface": "getattr(dispatch_sync_run,'apply_async')",
       "class": "call_site_getattr_indirection",
       "source": {
         "file": "src/dev_health_ops/api/admin/routers/sync.py",
-        "line": 2478
+        "line": 2530
       },
       "queue_or_cadence": "sync",
       "dispatches": "dispatch_sync_run",
@@ -2652,12 +2652,12 @@
       "route_mount_prefix": ""
     },
     {
-      "id": "api_trigger_endpoint:src/dev_health_ops/api/admin/routers/sync.py:2231",
+      "id": "api_trigger_endpoint:src/dev_health_ops/api/admin/routers/sync.py:2283",
       "surface": "POST /sync-configs/{config_id}/trigger",
       "class": "api_trigger_endpoint",
       "source": {
         "file": "src/dev_health_ops/api/admin/routers/sync.py",
-        "line": 2231
+        "line": 2283
       },
       "queue_or_cadence": "n/a",
       "dispatches": "dispatch_sync_run",
@@ -2677,12 +2677,12 @@
       "route_mount_prefix": ""
     },
     {
-      "id": "api_trigger_endpoint:src/dev_health_ops/api/admin/routers/sync.py:2377",
+      "id": "api_trigger_endpoint:src/dev_health_ops/api/admin/routers/sync.py:2429",
       "surface": "POST /sync-configs/{config_id}/backfill",
       "class": "api_trigger_endpoint",
       "source": {
         "file": "src/dev_health_ops/api/admin/routers/sync.py",
-        "line": 2377
+        "line": 2429
       },
       "queue_or_cadence": "n/a",
       "dispatches": "dispatch_sync_run",
--- a/contracts/jobs/v1/transitional-inventory.json
+++ b/contracts/jobs/v1/transitional-inventory.json
@@ -2529,12 +2529,12 @@
       "notes": "grep-evading form; POST /integrations/{integration_id}/backfill"
     },
     {
-      "id": "call_site_getattr_indirection:src/dev_health_ops/api/admin/routers/sync.py:2385",
+      "id": "call_site_getattr_indirection:src/dev_health_ops/api/admin/routers/sync.py:2432",
       "surface": "getattr(dispatch_sync_run,'apply_async')",
       "class": "call_site_getattr_indirection",
       "source": {
         "file": "src/dev_health_ops/api/admin/routers/sync.py",
-        "line": 2385
+        "line": 2432
       },
       "queue_or_cadence": "sync",
       "dispatches": "dispatch_sync_run",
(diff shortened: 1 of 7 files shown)


chrisgeo:
Companion web PR: https://github.com/full-chaos/dev-health-web/pull/850

PR #1506, opened 2026-08-05
CHAOS-3430: surface watermark lag for ratcheting HEAVY datasets

chrisgeo:
## Defect

CHAOS-3412's HEAVY incremental window ratchet caps a cold-start window at
`SYNC_INCREMENTAL_HEAVY_MAX_WINDOW_DAYS` (default 7) and stamps the watermark at the
window **end**. Each capped tick therefore finalizes as an ordinary SUCCESS with
`last_sync_success=true`, while the dataset's watermark may still trail `now` by weeks.
Setup, health, and status surfaces read "complete, no failure" during a catch-up that can
take ~13 scheduled ticks from a 90-day cold start. Nothing persisted or rendered the lag.

Found by adversarial review of the ratchet changeset; graded medium and deferred at the
time because catch-up is bounded and loses no data.

## Design

Adds `SyncRunDatasetFreshness` to `SyncRunUnitSummary`, served by the existing
`GET /sync-runs/{run_id}/units` — the endpoint the admin sync-run detail page already
fetches. Deliberately a **rollup**, not extra fields on `SyncRunUnitResponse`: the ticket's
granularity is per `(source, dataset)`, units repeat that pair across chunks, and keeping
out of `_unit_to_response` made the rebase over #1503/#1504 conflict-free.

Per entry: `source_id`, `source_name`, `dataset_key`, `cost_class`, `watermark_at`,
`lag_seconds`, `catching_up`, `ticks_behind`, `window_cap_days`; plus
`catching_up_dataset_count` on the summary.

- **Read-only.** Computed from the existing `sync_watermarks` rows at response time. No
  migration, no new persisted state, and run finalization semantics are untouched — a
  capped run *is* a successful run.
- **Narrow flag.** `catching_up` requires HEAVY **and** lag strictly greater than the cap.
  A watermark exactly one cap-window back is the healthy steady state of a ratchet
  mid-flight, not arrears. `lag_seconds` is still reported honestly for every cost class;
  only the verdict is withheld.
- **Cap comes from the planner, not a second env read.** `heavy_max_window_days()`
  delegates to `planner._effective_heavy_max_window_days()`, which *widens* the cap when
  `SYNC_WATERMARK_OVERLAP >= cap`. A duplicated env read would judge lag against a cap the
  planner does not size windows with, and would report every dataset inside the widened
  window as behind — drift exactly in the configurations the planner exists to correct.
- **One batched query** per run, then three-tier watermark precedence applied in memory.
- Entries only for `WatermarkBehavior.INCREMENTAL` datasets; a dataset that never stamps a
  watermark has no lag to report.

Docs: new "Datasets still catching up" section in
`docs/admin/sync-and-coverage/status-and-freshness.md`, beside the budget section.

## Product framing

Readiness semantics are **unchanged** by this PR. The accepted position is
**disclose, don't degrade**: org readiness must not flip to partial/pending during
catch-up, because catch-up is bounded and self-healing, and degrading would fire on nearly
every org's first days. Readiness answers "can I trust this answer?", not "is ingestion
finished?" — an org catching up to April still answers last week's trend correctly.
The disclosure half is tracked separately in **CHAOS-3438** (ConnectorStatus coverage-depth
qualifier), which also carries the open product question of whether a catch-up ETA
threshold should ever escalate.

## Tests — 43 new, RED-first

- `tests/test_watermark_lag.py` (25): boundaries (at-cap, one-second-past), non-heavy never
  flagged, missing watermark, clock-skew clamp, naive-tz, cap env rejection, and two tests
  pinning the planner delegation including the overlap-widening case.
- `tests/test_watermark_lag_parity.py` (10): a **differential oracle** running the new
  in-memory `resolve_watermark` against the production `get_watermark` over real seeded
  rows — 8 acceptance cases covering every precedence tier, plus a test asserting the
  comparator itself still fails when fed a truncated row set, so it cannot pass while blind.
- `tests/api/admin/test_sync_run_units.py` (+8): flags while the run reads SUCCESS, clears
  once caught up, never flags non-heavy, never-stamped reads unknown, omits no-watermark
  datasets, honors the configured cap, dedupes pairs deterministically, cross-org isolation.

**11 mutations planted, all killed by their specific guard**, both sources verified
byte-identical after restore: always-false verdict (7 fail), drop heavy clause (6), `>`→`>=`
(at-cap only), drop skew clamp, drop `org_id` filter, drop INCREMENTAL filter, drop planner
delegation (overlap test only). One mutation that failed to apply was recorded INVALID and
redone rather than counted.

Also adds `SyncWatermark` to `_TABLES` in `tests/api/admin/test_integrations.py`, since the
units endpoint now reads that table. The endpoint deliberately does **not** swallow a
missing-table error — a missing table in production is a broken deployment and must fail
loudly rather than degrade to "no lag information".

## Validation

`ci/local_validate.sh`: ruff format, ruff check, mypy, and the FULL unit suite green.

Note for reviewers: a sourced `ops/.env` in the shell makes this gate report false RED on
clean `main` with no code change — 15 tests, traced to `.env` keys including
`LOG_LEVEL=debug` and `AUTH_AUTO_CREATE_ORG_ON_REGISTER=false`. Verified by differential
(identical FAILED lists on branch and clean `main`; all 15 pass with `.env` keys unset).
Tracked and being fixed in **CHAOS-3439**; not a property of this change.

### DIFF
--- a/docs/admin/sync-and-coverage/status-and-freshness.md
+++ b/docs/admin/sync-and-coverage/status-and-freshness.md
@@ -67,6 +67,53 @@ remedies are to synchronize a narrower window with a bounded backfill, or to
 raise that bucket's cap. Raising per-synchronization ingest caps is not a
 remedy — it changes what is fetched, not whether the unit is admitted.
 
+### Datasets still catching up
+
+A successful run does not mean a dataset has reached the current time. High-cost
+record families synchronize through capped incremental windows, one window per
+scheduled tick, and each capped tick finalizes as an ordinary successful run.
+Run status alone therefore reads "complete" for a dataset whose coverage may
+still be weeks behind.
+
+The run's unit rollup reports, for every (source, dataset) pair it planned that
+carries a watermark:
+
+- the stored watermark, and how far behind the current time it is;
+- whether that pair is **catching up** — a high-cost dataset trailing by
+  strictly more than the configured window cap; and
+- roughly how many further scheduled ticks the pair needs to reach the current
+  time, at one capped window per tick.
+
+Read these separately from run status:
+
+- **Catching up** is a healthy in-progress state, not a fault. The remedy, if
+  the pace is unacceptable, is a bounded backfill over the outstanding period —
+  catch-up itself does not accelerate.
+- A pair trailing by no more than one window is the steady state of a healthy
+  ratchet mid-flight and is not flagged.
+- Only high-cost families are capped, so only they are flagged. A lag is still
+  reported for every other dataset; interpret a large one there as a different
+  problem — a stalled schedule, a failing unit, or a provider gap — and diagnose
+  it through the run boundary above.
+- A dataset with no watermark at all has never recorded a successful read. It
+  reports no lag, because there is no coverage to measure from; that is a
+  cold-start or a never-succeeding dataset, not a dataset that is behind.
+- A collapsed record family reports one entry per child dataset, not one for the
+  family as a whole. Each child carries its own watermark, so a family that
+  looks current overall can still hold a badly stale child.
+- The estimate of remaining ticks accounts for the configured watermark overlap.
+  Each window re-reads that overlap, so one tick's forward progress is the
+  window span minus the overlap; a large overlap means many more ticks than the
+  window span alone would suggest.
+
+This report is scoped to the run you are reading, and declares that scope
+explicitly. It covers only the source-and-dataset pairs that run planned. A run
+restricted to one source, or to a subset of datasets, that reports nothing
+catching up is not telling you the workspace is current — it is telling you
+nothing was behind *among the pairs it touched*. To ask the workspace-wide
+question, read freshness per dataset across every configured source rather than
+inferring it from a single run.
+
 ## Check freshness against the product question
 
 Compare:
--- a/src/dev_health_ops/api/admin/routers/integrations.py
+++ b/src/dev_health_ops/api/admin/routers/integrations.py
@@ -625,6 +625,14 @@ async def get_sync_run_units(
         and unit.result.get("error_category") == "budget_deferred"
     )
 
+    # CHAOS-3430: watermark-vs-now lag per (source, dataset). A capped HEAVY
+    # tick finalizes as an ordinary SUCCESS, so run status alone cannot show
+    # that the dataset is still ratcheting toward the current time.
+    dataset_freshness = await svc.build_dataset_freshness(units)
+    catching_up_dataset_count = sum(
+        1 for entry in dataset_freshness if entry["catching_up"]
+    )
+
     return SyncRunUnitSummary(
         by_status=rollups["by_status"],
         by_source=rollups["by_source"],
@@ -638,6 +646,10 @@ async def get_sync_run_units(
         next_retry_at=min(retry_times) if retry_times else None,
         retry_exhausted_unit_count=retry_exhausted_unit_count,
         budget_blocked_unit_count=budget_blocked_unit_count,
+        dataset_freshness=dataset_freshness,
+        catching_up_dataset_count=catching_up_dataset_count,
+        # Declared, not implied: these describe only what THIS run planned.
+        dataset_freshness_scope="run",
         units=[
             _unit_to_response(u) for u in (units if limit is None else units[:limit])
         ],
--- a/src/dev_health_ops/api/admin/schemas/integrations.py
+++ b/src/dev_health_ops/api/admin/schemas/integrations.py
@@ -161,6 +161,34 @@ class SyncRunResponse(BaseModel):
     model_config = ConfigDict(from_attributes=True)
 
 
+class SyncRunDatasetFreshness(BaseModel):
+    """Watermark-vs-now lag for one (source, dataset) pair (CHAOS-3430).
+
+    Read from the existing ``sync_watermarks`` rows at response time — the
+    incremental window ratchet finalizes each capped HEAVY tick as an ordinary
+    successful run, so run status alone cannot say how far behind the dataset
+    still is.  ``catching_up`` is set only for a HEAVY dataset trailing by
+    strictly more than ``window_cap_days``; every other dataset still reports
+    an honest ``lag_seconds`` without the verdict.
+    """
+
+    #: IntegrationSource UUID, matching ``SyncRunUnitResponse.source_id`` so a
+    #: caller can reuse an already-resolved source label.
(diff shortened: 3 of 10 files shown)


chrisgeo:
## Codex fix round — F1–F4 applied

Each finding reproduced first (RED), then fixed, then mutation-verified. Full gate green.

### F1 (HIGH) — composite work-item units hid their child watermarks

Confirmed, not merely plausible. The planner collapses the work-item family into one
composite unit (`dataset_key="work-items"` + `family_dataset_*` flags) while the worker
advances a **separate watermark per enabled child**
(`workers/sync_units.py::_watermark_dataset_keys`). Reading the composite's raw key meant a
badly-stale child was invisible — RED: `assert 'work-item-comments' in {'work-items'}`,
with a 45-day-stale child returning no entry at all.

Fixed by expanding to effective dataset keys via `sync_coverage._effective_dataset_keys` —
**reusing** the existing expansion rather than restating it, so worker stamping, coverage
math and freshness cannot drift apart. Per-child watermark semantics are preserved: each
child resolves against its own row.

### F2 (MEDIUM) — `ticks_behind` ignored the watermark overlap

Confirmed, and the magnitude matched the finding. The planner's own docstring defines the
window as `[W - overlap, W - overlap + cap]`, so net advance per tick is `cap - overlap`.
RED showed **`assert 12 == 83`** — a 7x understatement at a 6-day overlap on a 7-day cap —
and `4 == 40` under the clamp-widened cap.

Fixed by adding `planner.heavy_ratchet_net_advance_seconds()` and delegating to it,
extending the cap delegation already in place rather than copying arithmetic into a second
place. The verdict still keys on the **cap** (a dataset within one window is a healthy
ratchet mid-flight); only the tick **estimate** moved to net advance. Those are different
questions and are now computed from different quantities on purpose.

### F3 (MEDIUM) — O(P x R) scan replaced with a pre-built index

`WatermarkIndex` keys all three precedence tiers once, so each pair resolves in O(1)
instead of re-walking every row per tier. The `SELECT` is additionally narrowed to
dataset/target values that could satisfy some tier.

The differential oracle earned its keep here: mutating the index to drop the reverse-legacy
tier failed **both** the unit test and
`test_watermark_lag_parity.py::...[reverse-legacy fallback warms a sibling from the raw git row]`.
That is precisely the silent precedence change the oracle exists to catch, so the
optimization is safe by construction rather than by inspection.

Bound, rather than benchmarking theatre: resolution is now O(R) to index plus O(1) per
pair, against O(P x R) before.

### F4 (MEDIUM) — run-scope declared, not implied (minimal fix only)

No org-scoped surface here, by decision. Instead the block declares what it is:

- `SyncRunUnitSummary.dataset_freshness_scope: "run"`;
- docs state the report covers only the pairs that run planned, and that a run reporting
  nothing catching up is not a workspace claim;
- web copy says so on the surface (dev-health-web#853).

The org-scoped requirement is recorded on **CHAOS-3438**, including the trap that an
org surface aggregating anything less than all configured pairs would inherit the same
defect under a new name.

## Verification

- 50/50 targeted (`test_watermark_lag`, `test_watermark_lag_parity`, `test_sync_run_units`).
- **4 mutations planted, 4 killed**, each by its intended guard: net advance reverted to raw
  cap (2 fail), family expansion dropped (1), index tier dropped (2 — incl. the oracle),
  scope reverted to `"workspace"` (1). Sources verified byte-identical after restore.
- `bash ci/local_validate.sh`: **GATE PASSED — 12078 passed, 0 failed**; ruff format, ruff
  check, mypy, scratch-CH create + migrate, FULL unit suite, argMax live-exec proof.

Note for reviewers: that gate is only green with a clean environment. A sourced `ops/.env`
turns 15 unrelated tests red on clean `main` with no code change — see **CHAOS-3439**.

chrisgeo:
## Verify round — per-child cost class (last finding)

Reproduced first, fixed, mutation-verified. Rebased onto `a8a8da3f0` with zero content drift.

### The finding, with its severity split honestly

`pairs.setdefault` stamped every expanded child with the **composite's** `cost_class`, so a
collapsed work-item-family unit attributed its own class to all children. Two distinct
consequences, and they are not equally live:

- **Live misreporting.** `work-item-labels` and `work-item-projects` are registered
  **light**; they were being reported as the composite's **medium**. Wrong in responses
  today. RED: `assert 'medium' == 'light'`.
- **Latent suppression.** No work-item-family child is registered heavy right now, so no
  verdict is currently being suppressed. But flagging keys on `cost_class == "heavy"`, so a
  heavy child would have its catch-up verdict silently dropped the moment a provider
  registered one. RED: `assert 'medium' == 'heavy'`.

Calling the second one latent rather than live because that is what the registry actually
says today — it is worth fixing precisely because the failure mode is silent, not because
it is currently firing.

### Fix

`_dataset_cost_class(provider, dataset_key, unit_cost_class)` resolves per child from
`get_dataset_spec` — the same provider-aware registry the rest of the sync stack uses, so a
provider-specific cost class is honoured rather than guessed. Falls back to the unit's own
class when the registry has no entry for the pair: an unknown dataset is better reported
with the run's class than dropped or crashed on.

### Tests

Two RED-first regressions on a real collapsed composite (`work-items` + three
`family_dataset_*` flags):

1. each child reports its **own** registered class (`work-items` medium,
   `work-item-labels` light, `work-item-comments` medium);
2. a **heavy child of a non-heavy composite** is still flagged when stale — asserting the
   class, `catching_up`, `ticks_behind`, and the rollup count.

The second patches the real `_REGISTRY` entry production reads (via `dataclasses.replace`)
rather than stubbing the lookup, so it exercises the actual resolution path.

**2 mutations planted, 2 killed:** reverting to `str(unit.cost_class)`, and neutering the
resolver to always fall back — each failed both tests. Source byte-identical after restore.

### Rebase + validation

Rebased onto `a8a8da3f0` (CHAOS-3219 #1505). That commit is disjoint from this changeset —
acceptance artifacts only — and I verified rather than assumed it: blob hashes captured for
all ten touched files before and after, **drift = 0**, all identical.

`bash ci/local_validate.sh` on the amended tip: **GATE PASSED — 12080 passed, 102 skipped,
1 xfailed, 0 failed**; ruff format, ruff check, mypy, scratch-CH create + migrate, FULL unit
suite, argMax live-exec proof. (12078 previously, +2 for the new regressions.)

Reviewer note unchanged: that gate is only green in a clean environment — a sourced
`ops/.env` turns 15 unrelated tests red on clean `main` with no code change (**CHAOS-3439**).

PR #1520, opened 2026-08-06
CHAOS-3403: host-wide single-flight lock for ci/local_validate.sh

chrisgeo:
## CHAOS-3403 — host-wide single-flight lock for `ci/local_validate.sh`

Split from a larger changeset: the env-neutralization half shipped separately as #1516 because it carried no unresolved review findings. This half stayed on its own branch until the lock's review findings were closed — this PR closes them.

### Why this exists

Tonight, on this machine, a lane ran two `local_validate.sh` instances from the same worktree by accident: a background launch whose tool timeout did not actually bind the process, so the first run never died while the lane believed it had, and started a second one ~20 minutes later. Same worktree means the same deterministic scratch-DB name (`ci_local_validate_<hash-of-worktree-path>`). The second run's opening `DROP DATABASE IF EXISTS` executed against the database the first run was actively using mid-suite, and the first run's argMax child was killed. This happened to an operator who *had* checked for a running gate first — it is exactly the time-of-check-to-time-of-use race this lock closes.

**Evidence, kept honest and not collapsed into one claim:**
- ClickHouse-state corruption from concurrent scratch-DB use is **ruled out** — verified via `compose.yml`: the container is a single vanilla `clickhouse-server:latest` with no ZooKeeper/Keeper service and no replication config, so there is no `ReplicatedMergeTree` coordination for two runs to corrupt; each scratch DB is independent once created.
- Resource exhaustion (two full pytest-xdist `-n4` runs plus two ClickHouse migrations contending for one host's CPU/RAM) is **well-evidenced but not reproduced live** in this PR — reproducing it live means deliberately starving this shared dev machine, which wasn't worth the cost given the collision above already demonstrates the core race.

### Design

- A symlink at `/tmp/dev-health-ops-local-validate.<CH_CONTAINER>.lock` whose **target string is the owner metadata** (`pid|start-time|cwd`). `symlink(2)` publishes the name and its content in one syscall — there is no window where the lock exists but is unpopulated, which is what made the first (mkdir + separate metadata files) design losable to an adversarial review.
- `flock(1)` is absent on macOS, hence the symlink approach rather than a kernel-level file lock.
- Bounded blocking wait (`LOCK_WAIT_SECS`, default 1800s), with `LOCK_WAIT_SECS=0` for fail-fast.
- `LOCK_DIR` is pinned to a fixed `/tmp` path rather than derived from `$TMPDIR` — keying a host-wide mutex on an inherited env var is not host-wide, which is the same bug class this PR exists to fix (two agents with different ambient `$TMPDIR` would silently acquire two different locks and both proceed).
- PID-reuse hardening: `kill -0` alone only proves *something* holds that PID now; the lock also cross-checks `ps -o lstart=` (start time) against what was recorded at acquire time, so a PID recycled after a `kill -9` isn't mistaken for the original holder.

### Second adversarial review — five findings, all reproduced and fixed

1. **HIGH — locale-dependent `ps -o lstart=`.** Neither the write side nor the read side pinned a locale, so `ps -o lstart=` renders differently under different `LC_TIME`/`LC_ALL` (`Wed Aug 5 20:46:09 2026` under `C` vs `Mi. 5 Aug. 20:46:09 2026` under `de_DE.UTF-8`). A live holder recorded under one locale was reproducibly judged dead by a checker running under another — no timing race needed. Worse, this cascaded: `release_lock()` was gated only on a local flag and `rm -f`'d unconditionally, so the wrongly-reclaimed holder's own exit deleted a legitimate new owner's live lock. **Reproduced as a genuine double-acquisition path** and fixed: every `ps -o lstart=` call now goes through one locale-pinned (`LC_ALL=C`) helper; release and reclaim both compare the lock's current on-disk content against what was actually read/computed before deleting anything.

2. **HIGH — reclaim's stale decision wasn't atomic with its delete.** `reclaim_stale_lock()` performs several syscalls (readlink, `kill -0`, `ps`) to decide "stale," then unconditionally deleted whatever was *currently* at the path — not what it read. **Reproduced as a genuine double-acquisition path**, widening the window with a `ps` wrapper that sleeps (the same technique the test suite used to prove the original mkdir-based bug): a second process reclaimed and acquired cleanly inside the window, and the first process's delayed delete then removed the second's live lock. Fixed by the same compare-and-delete mechanism as Finding 1, applied structurally to both reclaim and release.

3. **HIGH — every `ln -s` failure was treated as "lock already held."** A missing parent directory, a permissions error, and a read-only filesystem all produced the same "waiting..." message and polled out the full `LOCK_WAIT_SECS` before dying with a diagnosis naming a PID that doesn't exist. Reproduced: `LOCK_DIR=/no/such/parent/dir/lock.file LOCK_WAIT_SECS=5` hung for the full window with a wrong diagnosis. Fixed: contention (the path is already a symlink) is now distinguished from every other failure, which `die`s immediately with the real reason.

4. **MEDIUM — SIGINT to the tracked PID alone cannot interrupt a blocked foreground command.** `run_stage`, `ch_migrate`, and `docker exec` all run as foreground external commands; bash defers a signal's trap until control returns to it, and a synchronous wait for a foreground child is exactly the state where control never returns until that child does. This affects any orchestrator that signals only the tracked PID (e.g. Python's `Popen.send_signal(SIGINT)`, standard for non-terminal automation) rather than the whole process group.
   Resolved **narrower than initially proposed**. The obvious fix — `trap on_exit INT TERM EXIT` — was tested directly and found to be a regression, not a fix: installing an explicit trap for a signal makes bash defer running it until the foreground command returns, the same deferral rule that already made SIGINT undeliverable. TERM currently has **no trap**, so its default disposition already interrupts a blocked foreground command immediately (verified: untrapped TERM kills a 20s `sleep` in ~0s; the same signal trapped took the full ~19s, waiting for the sleep to finish on its own). Adding a TERM trap would have broken the one signal that already worked. The fix actually shipped is `trap 'exit 130' INT` only — TERM is deliberately left untrapped. This fixes SIGINT delivered process-group-wide (a real terminal Ctrl-C) or between commands; it does **not** fix, and no trap can fix, SIGINT to the tracked PID alone during a blocked foreground command. No regression test was written for that specific unfixable case — it cannot be made to fail deterministically without faking bash's internal signal-deferral behavior, and a test that can only ever pass is worse than no test.

5. **HIGH — the interval-overlap regression tests had silently degraded to whole-second resolution**, which made them unable to fail. `--lock-probe` used `${EPOCHREALTIME:-$(date +%s)}`; under the test suite's own fixed `PATH` (`/usr/bin:/bin:/usr/local/bin`), `bash` resolves to Apple's `/bin/bash` 3.2.57, which has no `EPOCHREALTIME` builtin, so every probe silently fell back to `date +%s` — whole seconds. The tests were deliberately reformulated to assert non-overlapping `[acquired, released)` intervals specifically so a reintroduced microsecond-scale double-acquisition bug would be falsifiable; at whole-second resolution, two racing acquisitions round to identical integer timestamps, and `a_end <= b_start` treats equal integers as "no overlap" — a pass. **The test could not fail, and now can**: built a hybrid script (the pre-fix buggy lock combined with the new sub-second probe) and re-ran Finding 2's reproduction through it — the exact same assertion the existing tests use went from silently passing (whole-second rounding masked the bug) to correctly failing (sub-second timestamps showed the real, fully-nested overlap: one interval `[.. .459919)` entirely containing the start of the other at `.504120`). Fixed: `lock_probe_time()` now tries `EPOCHREALTIME`, then `date +%s.%N`, then `python3`, then `perl`, validating each fallback's *output* by shape rather than trusting a zero exit status (a bug caught in the first draft of this fix itself — `command -v python3` succeeding doesn't mean that invocation produced valid output), and fails loudly with no silent whole-second fallback if none produce a valid sub-second timestamp.

**Incidental proof of the exit-order fix under a real kill:** during verification, a tool-imposed foreground timeout SIGTERM'd a live gate run mid-unit-suite. The lock was released and the scratch DB was dropped before the process fully exited, with no orphaned `docker exec` left behind — the `release_lock()`-before-`cleanup_scratch()` ordering in `on_exit()` held under a genuine, uncontrolled kill, not just the deliberate test for it.

### Tests

Six new regression tests in `tests/tooling/test_local_validate_lock.py`, covering Findings 1, 2, 3, and 5 (Finding 4's core defect has no deterministic test, per above). All pass against the fix; all fail against the pre-fix baseline (verified both ways).

### Gate

`ci/local_validate.sh` run end-to-end, single instance, foreground: lint, typecheck, full unit suite (12093 passed), and the live argMax proof all green. `GATE PASSED`, exit 0. (The argMax stage completing this run also confirms the known CHAOS-3362 machine wedge did not recur.)

### One thing this diff can't show

`.claude/skills/ops-local-validate/SKILL.md` (in the `dev-health` monorepo root, which is **not itself a git repository**) was updated separately to teach the single-flight lock instead of a manual pre-push orphan-process check. That change has no version history to cite — the file isn't tracked by any repo, so nothing in this PR, in CI, or in review would ever surface it. The before/after below is real, not reconstructed: the prior text was captured verbatim in an agent session's context *before* the edit happened, which is the only reason it's recoverable at all — its provenance is a session transcript, not git, and that's a real weakness, not a formality.

**Before:**
> - **Single-flight, machine-wide.** Two concurrent gates deadlock. In a multi-agent session, request the gate slot from the orchestrator first; check for orphan `local_validate`/`pytest` processes before starting.

**After:**
> - **Single-flight, machine-wide.** A second concurrent gate no longer deadlocks (CHAOS-3403): it blocks behind a host-wide lock, waiting up to `LOCK_WAIT_SECS` (default 1800s), then fails with an actionable message naming the holder's PID and cwd. `LOCK_WAIT_SECS=0` fails fast instead of waiting. In a multi-agent session, request the gate slot from the orchestrator first anyway — it's a wall-clock courtesy, not a correctness requirement.

The line being replaced was itself instructing agents to *"check for orphan `local_validate`/`pytest` processes before starting"* — the literal time-of-check-to-time-of-use practice this PR's own header diagnoses as racy, and the one that produced tonight's collision. The most-read operational document for this gate was teaching the bug.

Stated plainly because it doesn't get fixed by this PR: **`.claude/skills/` has no version control and no review path.** A doc every agent loads before running this gate can drift arbitrarily far from the script it describes, with no signal anywhere that it happened. This PR exposes that gap; it does not close it.

`SCREENSHOT-WAIVER: CI tooling change, no rendered UI surface.`


chrisgeo:
## Rebased onto main — and one unrelated commit you should push back on if you disagree

`#1508`, `#1515` and `#1516` all merged, which put this branch in conflict. Two things happened bringing it up to date.

### 1. The conflict, and why it was only in one place

`.gitignore` is now identical to main (that half shipped in #1516), so it drops out of this PR entirely. The only conflict was `ci/local_validate.sh`, and only because #1516 — the env-neutralization half that was **split out of this branch** — kept evolving after the split: its own review added a REFUTED paragraph to the `REDIS_URL` rationale that this branch never saw.

Resolved by taking main's `PROXY_OFF` block verbatim and re-appending this branch's lock section unchanged. That's safe to assert rather than eyeball, because main's block is a strict superset — diffed line-for-line, it adds 23 lines and removes none. Verified the resolved file equals the pre-merge branch tip **plus exactly that one paragraph and nothing else**, so no lock code changed in the resolution.

Net diff vs main is now three files: `AGENTS.md` (+6), the lock, and its tests.

### 2. An unrelated one-line fix, because main is red

`05006f478` adds `ASK_DEV_QUA_SHADOW_ENABLED` to `tests/_env_isolation.py`'s scrub list. **This is not part of CHAOS-3403.** It's here because `origin/main` currently fails `test_checked_in_scrub_list_matches_derivation` and this branch inherits that through the merge — there is no green gate available without it.

The cause is an ordinary merge-order race, and it is worth saying plainly that **the drift guard catching this is the guard doing its job on its first real encounter**:

- #1517 (CHAOS-3389, merged 2026-08-05 20:03) added the only read of `ASK_DEV_QUA_SHADOW_ENABLED`, at `production_runtime.py:2424`.
- #1515 (CHAOS-3402, merged 2026-08-06 04:36) branched before that landed, so the scrub literal it checked in never saw the new name.
- Both PRs were green on their own branches. Only their combination on main is red.

**Scrub, not keep** — traced before deciding, per CHAOS-3402's own rule that a name is kept only when a CI lane or test tier actually supplies it. This var has exactly one reader and no writer anywhere in the repo: not `.github/workflows/*.yml`, not `ci/*.sh`, not `.env.example`, not `ops/.env`. Nothing supplies it, so admitting it from an ambient shell buys no coverage and costs the divergence the scrub list exists to remove.

Red/green pair, not an assumption: the gate observed this test **fail** before the change (`1 failed, 12360 passed`), and `test_env_isolation_contract.py` + `test_chaos_3389_qua_shadow.py` are 41/41 after it — the shadow-mode tests set the flag themselves and are unaffected by the scrub.

If you'd rather this went to main as its own PR, say so and I'll split it — I put it here because main being red blocks everyone, not because it belongs in a lock PR.

### Gate

Full `ci/local_validate.sh` green on the merged tree, all seven stages:

```
[PASS] lint: ruff format --check
[PASS] lint: ruff check
[PASS] typecheck: mypy
[PASS] ch-scratch-create (ci_local_validate_7a3e2e4bc65b)
[PASS] ch-migrate (upgrade + status --check)
[PASS] unit suite (FULL, not subset)     12361 passed, 105 skipped, 1 xfailed, 143.53s
[PASS] argMax live-exec proof (real engine)
GATE PASSED.
```

Both gate runs took the new lock and released it cleanly — the symlink target carried `PID|start-time|worktree` as designed, and CHAOS-3362's argMax wedge did not recur on either run.

PR #1432, opened 2026-08-03
ci: gate validation workflows on local ARC

chrisgeo:
## What changed

- Gates Python Tests, Lint, Live Backend E2E, Governance Gate, and Security Scan with `SELF_HOSTED_RUNNERS`
- Routes trusted repository events to `local-arc-runners` when `SELF_HOSTED_RUNNERS=enabled`
- Preserves the existing GitHub-hosted runners when the variable is unset or disabled
- Keeps fork pull requests on GitHub-hosted runners
- Uses cancellation-safe required-check aggregators for Tests and Lint
- Includes each edited workflow in its own path filter where needed so this PR exercises the real workload
- Removes the standalone Local ARC Smoke Test now that the production validation workflows exercise ARC directly

## Why

GitHub-hosted Actions capacity has intermittently left validation jobs queued. This PR adds ARC as an explicit fallback across several workload shapes: Python matrices, service containers, linting, repository governance, dependency auditing, SAST, and secret scanning.

The local ARC runners have also completed these workloads materially faster than the corresponding GitHub-hosted runners during validation.

## Gate behavior

| Event | `SELF_HOSTED_RUNNERS=enabled` | Otherwise |
|---|---|---|
| Same-repository PR | `local-arc-runners` | Existing GitHub-hosted runner |
| Push to `main` | `local-arc-runners` | Existing GitHub-hosted runner |
| Merge queue | `local-arc-runners` | Existing GitHub-hosted runner |
| Fork PR | Existing GitHub-hosted runner | Existing GitHub-hosted runner |

## Validation

- ARC scale set registered at the `full-chaos` organization level
- Tests, Lint, and Live Backend E2E execute successfully on ARC
- Security Scan jobs execute on ARC
- Governance Gate routes to ARC
- The standalone smoke workflow was removed because the actual gated workloads now provide ongoing validation

PR #1551, opened 2026-08-07
CHAOS-3482: fail the required test gate when the run proves nothing

chrisgeo:
Fixes CHAOS-3482.

## The defect

The required `test` aggregator read only `needs.test-matrix.result` and `needs.coverage.result`, and treated `skipped` as a pass. During the 2026-08-06 Actions outage the upstream `changes` job failed — and on a rerun 35 minutes later, was cancelled — in "Set up job". Both test jobs then evaluated their `if:` to false and reported `skipped`, the `success|skipped` arm matched, and the single REQUIRED check reported **SUCCESS on a run where zero tests executed**. Reproduced 2 of 2 attempts. Only `changes: FAILURE` also appearing in the rollup made the PR mergeable-blocked; a `cancelled` upstream would not have.

`skipped` cannot simply be rejected. `coverage` is excluded from every `pull_request` by design (CHAOS-2586) and `test-matrix` is excluded from docs-only changes by the path filter, so rejecting the literal string turns every PR red. **The reason a skip is legitimate is never in the skipped job's own result** — it is in `changes`: whether the gating decision completed, and what it decided.

## The fix

The verdict moves out of the inline step into `ci/aggregate_test_results.sh`, and the `test` job now `needs: [changes, test-matrix, coverage]`, passing `github.event_name`, `needs.changes.result` and `needs.changes.outputs.code`.

The rule: **a skip is legitimate only when the job's own `if:` provably selected against it**, which requires `code` to be *literally* `false` — not merely "not `true`". `changes != success` fails immediately, because a gating decision that did not complete makes every downstream result uninformative rather than merely suspicious.

`if: always()` is kept deliberately, and pinned by a test. Under `!cancelled()` (proposed on #1432) a cancelled run leaves the required check **skipped**, and branch protection counts a skipped required check as satisfied — that relocates the silent pass one level up instead of closing it. Cancellation must be reported by a job that ran.

### Full verdict matrix

| event | changes | code | test-matrix | coverage | verdict | why |
|---|---|---|---|---|---|---|
| pull_request | failure | — | skipped | skipped | **FAIL** | the incident |
| pull_request | cancelled | — | skipped | skipped | **FAIL** | the rerun, same false green |
| pull_request | cancelled | — | cancelled | cancelled | **FAIL** | superseded push |
| pull_request | failure | false | skipped | skipped | **FAIL** | a late-failing job still publishes earlier outputs |
| push | cancelled | false | skipped | skipped | **FAIL** | same |
| pull_request | skipped / empty / garbage | — | skipped | skipped | **FAIL** | unknown state is not evidence |
| pull_request | success | true | success | skipped | pass | normal PR (coverage excluded by design) |
| pull_request | success | false | skipped | skipped | pass | docs-only PR |
| push | success | false | skipped | skipped | pass | docs-only push |
| push | success | true | success | success | pass | main |
| merge_group | success | false | success | success | pass | merge queue runs both regardless of `code` |
| workflow_dispatch | success | — | success | skipped | pass | manual run; filter not run there |
| pull_request | success | true | skipped | skipped | **FAIL** | selected, yet skipped |
| merge_group | success | false | skipped | success | **FAIL** | merge queue skip is never legitimate |
| merge_group | success | false | success | skipped | **FAIL** | coverage is selected by the merge_group clause |
| workflow_dispatch | success | — / false | skipped | skipped | **FAIL** | a manual run that tested nothing |
| push | success | true | success | skipped | **FAIL** | coverage was selected |
| pull_request | success | — | skipped | skipped | **FAIL** | undecided filter ≠ docs-only |
| push | success | — | success | skipped | **FAIL** | same, for coverage |
| pull_request | success | — | success | skipped | pass | undecided `code`, but the job ran |
| merge_group | success | — | success | success | pass | same |
| any | success | true | failure / cancelled / unknown | * | **FAIL** | unchanged behaviour |
| any | success | true | * | failure / cancelled / unknown | **FAIL** | unchanged behaviour |

## Verification

A green run is what this bug *is*, so none is offered as proof.

**The guard observed failing.** The pre-fix aggregator was materialized from `origin/main`'s inline step and driven through the identical table: **17 of 32 rows wrong**. Post-fix: **0 of 32**. The two workflow-contract tests are RED against main's `test.yml`.

**Clause-level mutation, script.** Ten mutations, each deleting exactly one clause and each `bash -n`-valid: drop the `changes`-result guard; never select test-matrix; drop the merge_group exemption; drop the workflow_dispatch exemption; revert to "not `true`" selection; drop coverage's pull_request exclusion; drop its workflow_dispatch exclusion; drop its merge_group clause; make coverage treat undecided `code` as unselected; accept an unrecognized result. **10/10 killed, each by a named row.** The sweep found three guards that no row uniquely pinned — the four rows that close those are in the table above, added because of it.

**Mutation, workflow.** Ten defects planted in `test.yml` — drop `changes` from `needs`, `always()` → `!cancelled()`, drop `EVENT_NAME` from the step env, drop each `if:` clause, re-enable the filter on dispatch, rename the `code` output, drop the checkout, remove `requirements-docs.txt`, remove `docker/**` — each observed failing its specific guard, with the file restored byte-identical afterwards.

**Not a source-digest test.** The tests execute the real `ci/aggregate_test_results.sh` that the workflow runs, and derive their expectations from production sources: the two test jobs' own `pip install -r` targets, and `scripts/acceptance`'s `RUNTIME_DEPENDENCY_PATHS`. Both guards fail loudly on an empty measurement rather than passing when they find nothing.

`shellcheck` clean; `actionlint` clean apart from the pre-existing `ubuntu-26.04-arm` label notes; 92 tests pass across `tests/tooling` and the three other files that parse `test.yml`.

## Four more zero-test greens, found by adversarial review and fixed here

1. **workflow_dispatch.** A manual `gh workflow run test.yml` computed `code` against the default branch; a ref with no gated paths reported `code=false`, skipped both jobs, and passed. Now the filter does not run on dispatch and `test-matrix` is selected there unconditionally — the construction `lint.yml` and `typecheck.yml` already use.
2. **`requirements-docs.txt`.** Both test jobs install it before running anything, but it was not in the `code` filter, so a PR touching only it skipped everything and passed. Gated; the guard reads the jobs' own install commands rather than a hand-kept list.
3. **Undecided `code`.** After `changes` succeeded, "not `true`" read as docs-only — so a renamed output or a filter that published no decision bought a green check with nothing run. Selection is now the inverse question, and this deliberately is *not* a validation of `code`: on runs where the jobs executed, an odd value changes nothing, so fail-closed does not become fail-noisy.
4. **`docker/**`.** `docker/Dockerfile` is one of the acceptance suite's `RUNTIME_DEPENDENCY_PATHS` — hashed and asserted on — but was ungated. Added, with a guard that reads that production tuple.

## Known adjacent gaps, not fixed here

- **CHAOS-3513** — `lint.yml` and `typecheck.yml` carry the identical aggregator shape (`success|skipped`, no view of their own upstream `changes`). Same defect, same reachability. Mechanical follow-up: adopt this PR's script, no new design.
- **CHAOS-3514** — `tests/docs/*.py` run in this workflow's unit tier and read `docs/**` and `mkdocs.yml`, which the `code` filter does not select on, while `docs-guards.yml` runs standalone scripts rather than those tests. A docs-only PR can therefore break a `tests/docs` assertion that no gate executes. Closing it is a policy choice (full suite on every docs PR) or an ownership move, so it is filed for a decision rather than fixed here.
- Nothing here proves the path filter **complete**. The two guards cover what is derivable from production sources; a file that matters to the suite but appears in neither can still be missing.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,6 +50,15 @@ jobs:
               - 'ci/**'
               - 'migrations/**'
               - 'requirements.txt'
+              # Both test jobs `pip install -r requirements-docs.txt` before
+              # running anything, so it is part of the test environment. It
+              # was missing here until CHAOS-3482 Codex round 2: a PR touching
+              # only this file got code=false, skipped both test jobs, and the
+              # aggregator -- correctly, by its own rule -- called that a
+              # legitimate docs-only skip and reported the required check
+              # green. test_paths_filter_covers_every_file_the_test_jobs_install
+              # derives this list from the jobs' own install commands.
+              - 'requirements-docs.txt'
               - 'pyproject.toml'
               - 'pytest.ini'
               - 'alembic.ini'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -66,6 +66,12 @@ jobs:
               - 'scripts/**'
               - 'alerts/**'
               - 'compose.yml'
+              # docker/Dockerfile is one of scripts/acceptance's
+              # RUNTIME_DEPENDENCY_PATHS: the acceptance suite hashes it and
+              # asserts on it, so a change here can turn the suite red -- but
+              # the filter did not select on it, so such a PR skipped every
+              # test job and the gate went green (CHAOS-3482 Codex round 4).
+              - 'docker/**'
               - 'deploy/**'
 
   # Fast, pytest-xdist-parallelized unit suite on the supported Python
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -315,31 +315,36 @@ jobs:
 
   # Required-status-check gate. Branch protection requires a check named
   # exactly "test"; the matrix above produces "test-matrix (3.14)"
-  # and the coverage job produces "coverage", neither of which match. This
-  # aggregator always runs, treating skipped (path filter, or coverage skipped
-  # on pull_request) as pass and failure/cancelled as fail. It is the single
-  # required check for both PRs and the merge queue. Do not remove.
+  # and the coverage job produces "coverage", neither of which match. It is the
+  # single required check for both PRs and the merge queue. Do not remove.
+  #
+  # The verdict lives in ci/aggregate_test_results.sh rather than inline here,
+  # so it can be driven through the whole result matrix by
+  # tests/tooling/test_aggregate_test_results.py. See that script's header for
+  # why `needs.changes.result` -- not the downstream results alone -- is what
+  # separates a legitimate skip from an upstream failure that skipped
+  # everything (CHAOS-3482).
   test:
     name: test
+    # Must remain `always()`. Under `!cancelled()` (or any condition that lets
+    # this job itself skip) a cancelled run leaves the REQUIRED check *skipped*,
+    # and branch protection counts a skipped required check as satisfied --
+    # which moves the silent pass one level up instead of closing it.
+    # Cancellation has to be reported by a job that actually ran, so this job
+    # runs unconditionally and the script fails the gate on `cancelled`.
     if: always()
-    needs: [test-matrix, coverage]
+    needs: [changes, test-matrix, coverage]
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
       - name: Aggregate test results
         env:
+          EVENT_NAME: ${{ github.event_name }}
+          CHANGES_RESULT: ${{ needs.changes.result }}
+          CHANGES_CODE: ${{ needs.changes.outputs.code }}
           MATRIX_RESULT: ${{ needs.test-matrix.result }}
           COVERAGE_RESULT: ${{ needs.coverage.result }}
         run: |
-          echo "test-matrix result: ${MATRIX_RESULT}"
-          echo "coverage result: ${COVERAGE_RESULT}"
-          for result in "${MATRIX_RESULT}" "${COVERAGE_RESULT}"; do
-            case "${result}" in
-              success|skipped)
-                ;;
-              failure|cancelled|*)
-                echo "test gate failed (${result})"
-                exit 1
-                ;;
-            esac
-          done
-          echo "test gate passed"
+          chmod +x ci/aggregate_test_results.sh
+          ./ci/aggregate_test_results.sh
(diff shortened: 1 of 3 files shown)


chrisgeo:
## Real-execution proof: both arms observed

Neither arm is a local simulation. Both are GitHub Actions runs of the aggregator in this PR, and the fail arm reproduces the incident's *mechanism*, not an approximation of it.

### Fail arm — the incident's exact state, now RED

[Run 31136720437](https://github.com/full-chaos/dev-health-ops/actions/runs/31136720437), on a throwaway branch off this one whose `changes` job pointed at an unresolvable action ref — so it died in "Set up job" the same way the 2026-08-06 outage killed it ("Getting action download info: Failed to resolve action download info").

```
EVENT_NAME: workflow_dispatch
CHANGES_RESULT: failure   MATRIX_RESULT: skipped   COVERAGE_RESULT: skipped

test gate failed:
  - changes reported 'failure', not 'success' -- the job that decides what must run
    did not complete, so the downstream 'skipped'/'skipped' results are not evidence
    that anything was tested
##[error]Process completed with exit code 1.
```

Job conclusions: `changes` **failure**, `test-matrix` **skipped**, `coverage` **skipped**, `test` **FAILURE**.

That result tuple — `failure` / `skipped` / `skipped` — is byte-for-byte the one that printed `test gate passed` and reported the required check SUCCESS on run 31119023762. It is now red. The throwaway branch has been deleted; the run link outlives it.

### Pass arm — a legitimate skip still passes, and says why

[Run 31136674166](https://github.com/full-chaos/dev-health-ops/actions/runs/31136674166), this PR's own required check.

```
EVENT_NAME: pull_request
CHANGES_RESULT: success (code=true)   MATRIX_RESULT: success   COVERAGE_RESULT: skipped

coverage skipped legitimately: the coverage-gated suite does not run on event
pull_request with changes code=true (CHAOS-2586)
```

Job conclusions: `changes` **success**, `test-matrix (3.14)` **success**, `coverage` **skipped**, `test` **SUCCESS**.

This is the pair that matters. Both runs contain a job reporting the literal string `skipped`, and the gate reaches opposite verdicts — because it reads `changes`, not the skip. A repair that rejected `skipped` would have turned this PR red; the one that kept accepting it would have left the fail arm green.

### Local gate

`ci/local_validate.sh`: 7/7 stages green — ruff format, ruff check, mypy, ch-scratch-create, ch-migrate, full unit suite (**12816 passed, 249 skipped, 1 xfailed, 0 failed**), argMax live-exec proof.

Rebased onto `main` after #1546 and #1550 landed. Note for CHAOS-3489's scanner: `ci/aggregate_test_results.sh` uses `printf` throughout and contains no here-documents, so it passes that budget scan clean.

PR #1555, opened 2026-08-07
CHAOS-3513: close the lint and typecheck gates, and prove they still look at something

chrisgeo:
Fixes CHAOS-3513.

## A change to the CI gates reached main with lint and typecheck never having run

On 2026-07-31, PR #1349 — *"fix(ci): governance gate fails all merge-group runs"* — entered main through the merge queue. In `merge_group` runs [30597898813](https://github.com/full-chaos/dev-health-ops/actions/runs/30597898813) and [30597912205](https://github.com/full-chaos/dev-health-ops/actions/runs/30597912205) (lint), and 30597898822 / 30597912214 (typecheck), the `changes` job logged:

```
Detected 1 changed files
##[group]Filter code = false
```

That one file, confirmed by diffing the two merge-queue commits rather than inferring it:

```
$ git diff --name-only 68dff6c16ad60fd 7c1248a8d22d674
.github/workflows/governance.yml
```

Neither `lint.yml` nor `typecheck.yml` listed `.github/workflows/**` in its `code` filter. So `code=false`, `lint-job` and `typecheck-mypy` skipped, the `success|skipped` arm matched — and both required checks reported **SUCCESS**. The gates could be edited, weakened, or disabled by a PR that they then declined to run on.

## Three coupled defects

**1. A workflow that does not gate on itself** *(the lead; observed above)*. `.github/workflows/**` added to both filters. The guard is derived rather than hand-listed — each workflow must be selected by its **own** filter — so a fourth gate added later inherits it for free.

**2. No `merge_group` selection clause.** Both gated jobs selected on `workflow_dispatch || code == 'true'`, while `test.yml`'s `test-matrix` carries a merge_group clause precisely because dorny/paths-filter has no base/head diff to compute in the queue. Added to both. **I have not observed this one firing** — the two runs above are fully explained by (1). It is belt-and-braces, and is labelled that way rather than presented as the lead.

**3. The aggregator rule**, which is what converts either skip from a silent pass into a red gate. Both gates now use the script from CHAOS-3482, generalized from "test.yml's two jobs" to any gate's job list: each job is declared `<name>|<policy>|<result>`, where the policy names the `if:` expression the script mirrors. A test asserts every job's real `if:` against the policy its workflow claims, so a policy label cannot lie. The script now also fails when handed no jobs at all, and when handed a policy it does not model — both are wiring mistakes that would otherwise yield a green check for a measurement that never happened.

They had to land together: fixing only (3) would make the merge-queue skip **fail** rather than silently pass — correct, but red on every queue run until (1) and (2) landed.

## Two more holes, found by adversarial review

**Gates that run but look at nothing.** `ruff check .` honours `.gitignore` and `.ignore` for file discovery, and over an empty file set it exits 0. Measured on this repo: appending `src/` to `.gitignore` takes ruff's file set from **1045 source files to 0**, silently. `.gitignore` was in no filter, so the PR making that change would itself skip lint, and every run afterwards would be measured under the new scope. Added `.gitignore`, `.ignore`, `ruff.toml`, `.ruff.toml`, `mypy.ini`, `.mypy.ini`, `setup.cfg` to both filters — and to `test.yml`'s, since the guards for them live in that suite.

None of those config files exists today, which is the point: each is discovered by default and outranks or overrides `pyproject.toml`, and a PR **adding** one matched no pattern, so the gate skipped on the very change that redefined it.

**The guard I wrote for that was itself the defect.** Its first version resolved ruff at `ROOT/.venv/bin/ruff` and called `pytest.skip` when absent — and ruff is installed by `lint.yml` alone (`pip install ruff`; it is in no requirements file), so in the CI job that runs this suite it would have **skipped on every run while reading as coverage**. A mutation sweep cannot catch that, because a sweep only mutates code that runs.

The measurement now lives in `ci/check_lint_scope.sh`, invoked by `lint.yml` immediately **before** the two ruff steps whose exit status is the gate — the one place ruff is guaranteed to exist. The pytest side is hermetic: a `ruff` shim on PATH prints a chosen file set, so both arms are exercised with no real binary and nothing skips. A separate contract test pins the real invocation **and its ordering** — after the ruff steps it would report on a gate already decided — and that test is what makes the shim tests coverage rather than decoration.

## Verification

**The guard observed failing.** The pre-fix lint rule, materialized from `origin/main` and driven through the new table, gets **9 of 20 rows wrong** — including the merge-queue row that matches the observed incident. The fixed script gets **0 of 20** for both gates.

**The generalization changed no existing verdict.** The 32 rows that pinned CHAOS-3482 still run, unchanged, against the generalized script: **0 wrong**. That is the differential check on the refactor, not an assertion that it looks right.

**Mutation.** 12 clause-level mutations of the script (each deleting exactly one clause, each `bash -n`-valid): all killed, each by a named row. 10 workflow mutations across the three files: each fails its specific guard, and only for the gate mutated. 5 mutations of the scope check and its wiring — including moving it *after* the ruff steps — each fail their guard.

Two findings worth keeping from writing that check:

- The first version used `printf ... | grep -q` under `set -o pipefail`. `grep -q` closes the pipe on match, `printf` takes SIGPIPE and exits 141, and pipefail reports the pipeline failed — so a **successful match read as "no match"**, and the script would have failed the lint gate on every run. Membership is now tested with bash pattern matching, no pipe. (Same pipe-lifetime class as CHAOS-3362/3489, new site; noted there.)
- With `src/` swallowed, ruff still listed **1083 files** — comfortably above the count floor. Only the `src/` membership assertion caught it. A threshold alone would have passed, which is why both assertions exist.

## Residual risk

Nothing here proves the filters **complete**. The guards cover what is derivable from production sources — each workflow's own path, its jobs' `pip install -r` targets, `RUNTIME_DEPENDENCY_PATHS`, and the documented tool-scope inputs. A file that matters to a gate but appears in none of those can still be missing. **CHAOS-3514** (docs-only PRs vs `tests/docs`) is the known adjacent instance and is filed separately with its history check: real by construction, no production occurrence found in the sampled window.


chrisgeo:
## Fail arm: the defect itself, observed RED on a real runner

Not a stand-in. A throwaway branch off this one appended `src/` to `.gitignore` — the exact change this PR argues is dangerous — and dispatched Lint: [run 31140434914](https://github.com/full-chaos/dev-health-ops/actions/runs/31140434914).

```
ruff would check 1085 files
check_lint_scope FAILED -- ruff's file set collapsed:
  - no files under src/: the lint gate would pass having inspected none of the source
Check .gitignore, .ignore, and any ruff config for an exclusion that swallowed the tree.
```

Job conclusions: `changes` **success**, `lint-job` **failure**, `lint` **FAILURE**.

Step conclusions inside `lint-job` — this is the part that matters:

```
Install dependencies        = success
Check lint scope            = failure
Run ruff (format check)     = skipped
Run ruff (linting)          = skipped
```

The gate died **before** either ruff step ran. On `main` this same commit reaches `ruff check .`, which passes over a tree with no source in it, and the required `lint` check reports green.

**The production run also settles a design question.** ruff still listed **1085 files** with the entire source tree gone — comfortably above the 500-file floor. The count assertion did not fire; only the `src/` membership assertion did. Had this guard been written as "assert a healthy-looking file count", as it nearly was, it would have passed this run. That is why both assertions exist, and the number is recorded here so a later reader doesn't simplify one away.

## Pass arm

The PR's own three required checks, all now routed through the shared aggregator:

- `lint` — [run 31140422325](https://github.com/full-chaos/dev-health-ops/actions/runs/31140422325) ✅
- `typecheck` — [run 31140422200](https://github.com/full-chaos/dev-health-ops/actions/runs/31140422200)
- `test` — [run 31140422513](https://github.com/full-chaos/dev-health-ops/actions/runs/31140422513)

## Local gate

`ci/local_validate.sh`: 7/7 green — ruff format, ruff check, mypy, ch-scratch-create, ch-migrate, full unit suite (**12882 passed, 249 skipped, 1 xfailed, 0 failed**), argMax live-exec proof.

Throwaway branch deleted; the run link outlives it.

chrisgeo:
## Pass arm: all three required checks green, through the shared aggregator

- **`lint`** — [run 31140422325](https://github.com/full-chaos/dev-health-ops/actions/runs/31140422325): `changes` success, `lint-job` success, `lint` **SUCCESS**
- **`typecheck`** — [run 31140422200](https://github.com/full-chaos/dev-health-ops/actions/runs/31140422200): `changes` success, `typecheck-mypy` success, `typecheck` **SUCCESS** (`typecheck gate passed`)
- **`test`** — [run 31140422513](https://github.com/full-chaos/dev-health-ops/actions/runs/31140422513): `changes` success, `test-matrix (3.14)` success, `coverage` skipped (by design on `pull_request`), `test` **SUCCESS**

The scope check on the same run:

```
ruff would check 2130 files
check_lint_scope OK
```

## The pair

Both runs of `lint` executed the same `ci/check_lint_scope.sh` and the same `ci/aggregate_gate_results.sh`:

| | scope check | ruff steps | `lint` |
|---|---|---|---|
| this PR | 2130 files, OK | ran, passed | **SUCCESS** |
| throwaway with `src/` gitignored | 1085 files, **no files under src/** | **skipped** | **FAILURE** |

Same code, opposite verdicts, decided by whether the tree the gate exists to lint was actually in scope. Before this PR the second row is green, having linted nothing.

`test` also demonstrates the legitimate-skip half in production: `coverage` reported the literal string `skipped` and the gate passed anyway, because that job's own condition excludes `pull_request` — the distinction CHAOS-3482 introduced and this PR extends to two more gates.

PR #1557, opened 2026-08-07
ci: boot the Ask Dev acceptance stack on a schedule (CHAOS-3219 Phase 5, CHAOS-3488)

chrisgeo:
CHAOS-3219 Phase 5 Lane 5c, and CHAOS-3488's Phase-5 clause.

## Why

CHAOS-3488's Phase 5 clause, verbatim: *"The acceptance stack has no automated boot coverage anywhere in CI, so any migration merged after a mint silently invalidates the committed artifact and the failure surfaces only when a human runs the stack by hand."*

CHAOS-3488 **is** that failure — a world snapshot minted against alembic head `0086` merged onto a main already at `0087`, and nothing noticed until a human booted the stack for a Phase 2 exit-evidence run. The currency guard added there catches the migration-drift instance at unit tier. It cannot catch "the stack no longer boots" for any of the other reasons a stack stops booting. This lane can.

It is also the route named in `gate.web-default-ci`'s `blocked_reason` clause (3), merged in #1553: the launcher already invokes the web acceptance Playwright config, and once it runs inside a workflow producing ops-resident artifacts, that row acquires checkable evidence.

## What

**`.github/workflows/ask-dev-acceptance.yml`** (new)

- `schedule` (nightly 04:23 UTC) + `workflow_dispatch` **only**. Ruling D1 keeps the PR tier at unit + static wiring guards, so the live stack run is deliberately not a `pull_request` job. No `merge_group` arm — the merge queue has been disabled repo-wide since 2026-08-01 (CHAOS-3519), and a dead trigger reads as coverage.
- `ubuntu-latest`, per D1. GitHub-hosted **larger runners are not available to this org** — `GET /orgs/full-chaos/actions/hosted-runners` returns 404 "not supported for this organization". Verified, not assumed. The job reclaims the preinstalled toolchains it provably does not use and records `df -h` before and after into the uploaded logs, so the sizing claim is checkable per run rather than asserted once.
- Boots through the canonical launcher, then runs the armed corpus. Every launcher guard still runs: `compose config`, `--wait` per service, the world-restore digest check, the per-boot principal-login proof with its wrong-password negative control, the six ops-side smokes, and the web Playwright leg.
- `report-failure` announces a red night on a single reopened issue.

[...]

PR #1559, opened 2026-08-07
ci: fix an unsatisfiable corpus floor and drain the login window (CHAOS-3219 Phase 5)

chrisgeo:
Both defects were found by **dispatch run 1** ([31146779949](https://github.com/full-chaos/dev-health-ops/actions/runs/31146779949)) — the first real execution of the lane merged in #1557. Neither was findable before a CI run existed.

## 1. The attrition floor was unsatisfiable

`ASK_DEV_CORPUS_MIN_EXECUTED` was **100**. The maximum achievable executed count is **91**.

Of the 143 `case-*.json` files, 91 are `status='active'` and **52 are `status='declared-blocked'`** — and `run_report` deliberately does not count declared-blocked cases (its own docstring says so). So the guard could never pass, on any run, however healthy.

**A guard that cannot pass is worse than no guard**, because it teaches people to ignore the lane. The bad value came from a stale "144 cases" figure in an old run log instead of from measuring the corpus — I predicted 143 in the pre-registration and was wrong for the same reason.

Now **75** — ~82% of the measured 91 — catching catastrophic attrition while surviving the corpus's re-authoring churn.

Rather than only correcting the constant, `check_corpus_has_an_attrition_floor` now **computes the active-case count from the corpus JSON** and rejects any floor above it, so this class of mistake is structurally impossible rather than a comment someone has to read. Observed failing on the literal run-1 value:

```
pins ASK_DEV_CORPUS_MIN_EXECUTED=100, but only 91 corpus cases are
status='active' and run_report counts no others -- this floor is
UNSATISFIABLE and would fail every run however healthy.
```

## 2. The corpus cannot log in inside the boot's rate-limit window

All 91 active cases errored in setup with `HTTP 429: Rate limit exceeded` — **273 login calls** in one run.

`AUTH_LOGIN_IP_LIMIT` is `"20/15minutes"` and keyed **per IP** (`src/dev_health_ops/api/middleware/rate_limit.py`). A CI job is one IP, and the boot's per-principal login proof + `prepare_ask_dev_acceptance.py` + the corpus's own logins exceed 20 in a single window when they run back to back. **Local runs separate them by minutes, so no local run could have found this.**

Added the drain step between boot and corpus. This is the **proven local recipe**, not a sleep-until-it-works: `.p2exit_run.sh` carries exactly this wait with the same stated reasoning — *"Deliberate and stated, not a sleep-until-it-works: skip it and the run fails loudly on 429 rather than silently under-measuring."* 16 minutes on a nightly is free.

Pinned by a contract test that rejects any wait under 900s — a partial drain still 429s **and** pays most of the wait — and rejects an `if:` on the step.

**It is a workaround and says so.** [CHAOS-3529](https://linear.app/fullchaos/issue/CHAOS-3529) (session/token reuse per principal) is the real fix, and its acceptance criteria require deleting this step in the same change. The step's comment names the ticket, so the sleep and its removal condition point at each other from both ends.

**Production's rate limit is deliberately NOT relaxed for CI.** It is a security control, and an acceptance-only override would make this lane stop proving production's shape — the opposite of what it exists for.

## 3. Exit 66 had acquired a second meaning

Adding the floor made exit 66 mean both *"executed nothing"* and *"executed below the floor"*. Run 1 was the second. The failure-notification message would have led a reader to diagnose it as "never reached the corpus" — false. The message now names all four shapes, and the contract check pins that vocabulary rather than a single phrase.

## Gate

`bash ci/local_validate.sh`, plain, single-flight: **GATE PASSED**, all 7 stages, 13030 passed / 249 skipped / 1 xfailed.

The first run reddened on `test_budget_guard_cooldown.py::test_budget_wall_clock_cap_exhausts_below_the_count_cap`. Diagnosed rather than re-rolled: the assertion tests the bare substring `"360"` against a message containing a random 64-char hex digest, which that run rendered as `api.github.com:360b8725...`. A substring collision, unreachable from this diff (which touches one workflow and one tooling test), passing 3/3 in isolation. Filed as CHAOS-3530.

### DIFF
--- a/.github/workflows/ask-dev-acceptance.yml
+++ b/.github/workflows/ask-dev-acceptance.yml
@@ -206,6 +206,35 @@ jobs:
           ./scripts/acceptance/run_ask_dev_compose.sh --web-root "${WEB_ROOT}" \
             2>&1 | tee "${LOG_DIR}/boot.log"
 
+      # Dispatch run 1 (31146779949): all 91 active cases errored in setup with
+      # `POST /api/v1/auth/login returned HTTP 429: Rate limit exceeded` -- 273
+      # login calls in one run. AUTH_LOGIN_IP_LIMIT is "20/15minutes" and it is
+      # keyed PER IP (src/dev_health_ops/api/middleware/rate_limit.py). A CI job
+      # is one IP, and the boot's per-principal login proof plus
+      # prepare_ask_dev_acceptance.py plus the corpus's own logins exceed 20
+      # inside a single window when they run back to back. Locally they are
+      # minutes apart, which is why this could not surface before a CI run
+      # existed.
+      #
+      # This is the PROVEN local recipe, not a sleep-until-it-works: the Phase 2
+      # exit run script (.p2exit_run.sh) carries exactly this wait, with the
+      # same reasoning -- "Deliberate and stated, not a sleep-until-it-works:
+      # skip it and the run fails loudly on 429 rather than silently
+      # under-measuring." 16 minutes on a nightly is free.
+      #
+      # A WORKAROUND with a known replacement: CHAOS-3529 (session/token reuse
+      # across corpus cases) removes the need for it. DELETE THIS STEP when that
+      # lands -- its acceptance criteria say so explicitly -- and do
+      # not leave a 16-minute sleep in a workflow whose reason has gone away.
+      # Production's limit is deliberately NOT relaxed for CI: it is a security
+      # control, and an acceptance-only override would make this lane stop
+      # proving production's shape.
+      - name: Drain the per-IP login window before the corpus
+        run: |
+          set -euo pipefail
+          echo "waiting 16m for the per-IP login window to drain (AUTH_LOGIN_IP_LIMIT=20/15minutes)"
+          sleep 960
+
       # The corpus runner's QuotaBudget refuses to start un-budgeted: without
       # ASK_DEV_PLATFORM_MONTHLY_REQUEST_MAX and
       # ASK_DEV_PLATFORM_MONTHLY_COST_MAX_MICROUSD in ITS OWN process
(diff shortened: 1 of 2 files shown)


chrisgeo:
## Pre-registered expectation for dispatch run 2 (posted BEFORE the dispatch)

Merged as `e15715337`. State re-resolved by `git merge-base --is-ancestor`, not carried over from run 1:

| branch | ancestor of main? |
|---|---|
| #1557 (the lane) | YES |
| #1559 (this fix) | YES — `e15715337` |
| CHAOS-3497 (#1556) | YES |
| corpus-authoring (`48634f2cf`) | **NO — still unmerged** |

Corpus shape on main, measured now rather than assumed: **91 `active`, 52 `declared-blocked`, 143 files.** Floor on main is 75; drain step present with `sleep 960`.

### Corrected prediction

Run 1's pre-registration said 143 executed and the run reported 91. That was my error — I equated case files with active cases, and `run_report` does not count declared-blocked ones. So:

- **91 executed corpus cases.** Same number the run-1 failure reported, now predicted deliberately instead of discovered.
- **0 skipped.**
- **Floor 75 satisfied** (91 > 75) — it should no longer be what fails.
- **0 login 429s.** The drain step is the specific claim under test.

### What run 2 can and cannot prove

**Can:** that the floor is satisfiable and the 429 is cleared — i.e. that the corpus actually *reaches the product* rather than dying in setup. That is the whole point of this arm.

**Cannot: the GREEN arm.** corpus-authoring is still unmerged, so I expect run 2 to be **red on real case failures**, and that is the lane working correctly. I am not predicting the failure count or the failing set: with 3497's fix in and the corpus re-author out, I have no sound basis for one, and inventing a number would be fitting rather than predicting. The set is a **datum** to compare against the corpus lane's prediction file once it lands.

### How the red will be reported

By discriminator, not as pass/regression:

| signal | meaning |
|---|---|
| exit 66 + `executed 0 real corpus case(s)` | never reached the product |
| exit 66 + `executed N ... at least 75 must execute` | ran, fell under the floor |
| exit 67 | nothing measurable |
| exit 1 **with** an executed-count line | ordinary case failures — the expected arm |
| no executed-count line at all | died before the corpus: unmeasured, not failing |

Run 1 was the second row. Run 2 is expected to be the fourth.

Also re-verified on run 2, independent of the corpus result: 429 count zero, drain step ran its full 960s, boot and web checkout still succeed, gate reports the verdict, artifacts upload, teardown runs, and `report-failure` updates issue #1558 rather than opening a second one.

PR #1560, opened 2026-08-07
CHAOS-3525: LLM-backed subject resolution lands — the QUA proposal commits instead of being discarded

chrisgeo:
Closes CHAOS-3525. Supersedes canceled CHAOS-3425, whose requirement it carries verbatim.

## The ticket turned out to be smaller than its framing

CHAOS-3389's design comment says the question-understanding shadow is inert in production (`provider=None`). **That is stale.** CHAOS-3452 landed an isolated shadow quota, and `production_runtime.py` now passes a separate, separately-budgeted `qua_shadow_provider`. A real provider already runs when flagged.

What was missing was only the **promotion**: the record the shadow produces was a local variable nothing read. So this is *stop discarding the answer, under a gate, with disclosure and a verifier* — not *build LLM subject resolution*.

## What it does

`"What's the status of the ACR project"` against an organization whose project is "Dev Health Agent Context Runtime (Context Fabric)" now commits that project and answers, instead of dead-ending. "ACR" is the acronym of the label's *inner* words — a candidate the deterministic layer offers and correctly refuses to pick, which is exactly why it dead-ended.

The promotion sits in the gap CHAOS-3389 deliberately left: after the deterministic decision is computed and persisted, before the branch that acts on it. `subject_preflight.py`'s decision logic is untouched. A verified proposal mints through `committed_resolution_for` — whose docstring calls itself "the one construction of an exact-match committed scope … so there is one notion of committed scope rather than two that can drift" — as a third caller, never a second notion.

Three properties hold it together:

- **it fires only where the deterministic layer declined**, so a literal catalog match is never second-guessed by a model;
- **every LLM degradation leaves the run byte-identical to the seam being off** — absent, failing, low-confidence, out-of-slice, budget-exhausted, gate off. All 27 CHAOS-3389 isolation tests pass **unchanged** in commit mode; I predicted one would invert by design and was wrong, because none of them constructs the only mutable shape (a declining run with a successful proposal);
- **an accepted proposal is never silent** — the match is disclosed in the answer's own prose and on the wire, through the channel CHAOS-3497 opened.

## Adversarial review: all six authorization surfaces hold

No path commits an entity the caller cannot see; no cardinality widening. Traced rather than asserted: `selected_entity` is only ever set from that mention's own non-overlapping slice, and every slice comes from an org- and fingerprint-scoped search. Three independent blocks stop an ORGANIZATION-kind commit. Multi-mention promotion is structurally unreachable — confirmed by tracing that `record.mentions` is always *every* mention the interpreter found, not a subset. No double-insert on `(run_id, entry_ordinal)`.

### Two findings, both fixed

**The disclosure could be silently dropped.** `DevAnswer.warnings` caps at twenty and the helper yielded once that was spent — so a QUA-committed subject could reach a reader with no disclosure while the commit stood. A silent, model-chosen subject, reachable with **no authorization failure anywhere**, and it falsified this PR's own stated invariant. Reproduced, then fixed: the disclosure takes the last slot rather than yielding it, first in the list so a truncating renderer keeps it. Displacing a producer warning is the cheaper loss — an undisclosed subject is a claim the reader cannot check; a dropped twentieth warning is not.

**Authorization depended on search ranking.** The commit-time check re-ran a fuzzy search and looked for the entity among the results, so verification depended on how it ranked against everything else matching the same text. Review could not rule out a legitimate entity sitting below the cap and being refused, and left it open. Rather than bound that, the entity is now resolved **by its own identity** — explicit ref, no organization fallback, same authorization boundary, where another tenant's entity resolves unresolved rather than exact. Strictly stronger on the singular path, where it is the only receipt.

## Authorization, stated plainly

The LLM may only **select** from an already-authorized shortlist; it never introduces, widens, or ranks across tenants. Three layers: a per-call JSON Schema that makes an out-of-range index inexpressible, a runtime verifier re-checking each index against that mention's own slice, and a commit-time re-authorization by identity. On the singular path the third is **not** defence in depth — `committed_resolution_for` mints no `subject_set_fingerprint` and the executor's cross-check covers only set batches, so nothing downstream re-verifies a singular committed entity. It is documented and tested as the sole guard, including an assertion that it was actually consulted so it cannot pass by being skipped.

## The alias contract is amended, not repealed

`alias_matching`'s never-auto-commit rule is restated as permanent — a *derived* alias or acronym match is forbidden from auto-commit, forever — and the CHAOS-3289 reasoning affirmed as still holding, which is why it still binds. What the LLM path adds is a different **class of evidence**, reached only after that layer declines, so the two never compete for the same decision. A derived alias may now be *selected*; it can never be selected *quietly*.

## Evidence

RED-first throughout, and each guard observed failing on its own with content-diff-verified restores: revert the promotion → acceptance fails; remove the declined gate → the no-override test fails; remove the confidence floor → the low-confidence test fails; remove the commit gate → the commit-off test fails; disable the disclosure → the disclosure test fails; bypass the commit-time check → three tests fail; make QUA stamp the deterministic resolver's version → the provenance test fails.

**Three of my own tests could not fail when first written** and were fixed only because I mutated the implementation and watched them stay green: one asserted a constant against itself, one restated production's flag expression instead of calling it, and one passed against both the old and new authorization checks. All three now discriminate.

A real bug was caught the same way: the first cut committed the scope but the run still died in `internal_error` (`allowed_tools` is captured from the original decision, empty for a TERMINATE). The acceptance test passed anyway because it asserted the published scope rather than an actual answer — a later, stricter test caught it. The acceptance now asserts an answer.

## Scope and known gaps

Behind `ASK_DEV_QUA_COMMIT_ENABLED`, itself gated on the shadow flag, so the ladder stays off → shadow → commit and CHAOS-3389's inertness proofs keep meaning what they say. **Ships flag-off.**

- No `tests/acceptance/corpus/` file touched, and none needs to be: provenance rides `resolver_version="qua.v1"` plus the `committed_qua_subject` diagnostic, so `derive_resolution_path` still emits only its three deterministic paths and the corpus guard stays true. Assigning `qua-committed` is the corpus lane's work.
- The confidence floor (0.85) is **provisional**, documented as such, with every commit recording the observed confidence so calibration data accumulates from day one.
- Mixed-kind cohorts never auto-commit — recorded limitation; they fall through to ranked clarification.
- **The real-provider path is unproven end to end.** Correctness is proven with scripted providers; a live probe against a real model needs the contended acceptance stack and is follow-up work, not a merge blocker while the flag ships off.

Filed separately rather than folded in: **CHAOS-3528** (a preflight diagnostic outside its closed tuple) and **CHAOS-3531** (the sibling widening disclosure has the same drop-at-the-bound hole, found while fixing this one — a correction against my own merged CHAOS-3497 work).


chrisgeo:
### Self-review note: a duplicated fixture, deferred to the follow-up

`tests/api/dev/test_chaos_3525_qua_commit.py` defines its own `ACR_PROJECT`, but the shared harness already has one for the same real subject:

`tests/_chaos_3292_preflight.py:706` — `CHAOS_3388_ACR_PROJECT`, with the **identical** display label `"Dev Health Agent Context Runtime (Context Fabric)"` and, unlike mine, the **real production catalog row's** canonical id, read live from ClickHouse during the CHAOS-3388 incident.

So there are now two fixtures for one real subject, and the one I added invents its id. That is the same "two independently-drifting approximations of the same shape" this repo's own world manifest warns about, and I would flag it in someone else's PR.

It does not weaken what the tests prove — the acronym behaviour is driven entirely by the *label*, which is byte-identical, and the canonical id is only ever compared against itself. So this is fidelity and future-drift, not correctness.

**Deliberately not fixed here.** The gate has passed on this tree and CI is nearly green; amending would invalidate both and require re-taking a contended gate slot that another lane is queued for, for a test-only improvement. It is folded into the CHAOS-3531 follow-up PR instead, which needs its own gate run regardless — reusing `CHAOS_3388_ACR_PROJECT` and deleting the local duplicate.

Recording it rather than quietly carrying it, since an unnoticed duplicate fixture is exactly how the two drift apart later.

PR #1578, opened 2026-08-07
CHAOS-3499: baseline inventory, pinned corpus with independent oracles, and shadow harness design

chrisgeo:
Discovery deliverables 1–3 for CHAOS-3499 (shadow temporal-context trial), under epic CHAOS-3498. Shadow-only: isolated store, dev/acceptance scope, no user-visible output, no Ask Dev runtime change. **Nothing has been brought up** — every environment step is gated on an orchestrator-granted slot.

All trial material lives in `trials/chaos_3499/`, deliberately outside `docs/` (customer MkDocs site; every `.md` there must be in nav or the docs gate fails).

## What is here

| Deliverable | Path |
|---|---|
| 1 — Baseline inventory appendix (§15.2) | `trials/chaos_3499/docs/baseline-inventory.md` |
| 2 — Pinned corpus + independent oracles (§15.1) | `trials/chaos_3499/corpus/`, `harness/`, `tests/` |
| 3 — Shadow harness design + gated bring-up plan | `trials/chaos_3499/docs/harness-design.md` |

## Per-class result, before any arm runs

The seven evaluation questions classify as **(a) ×1, (b) ×1, (c) ×5**.

That distribution is itself a finding: the question set is weighted toward the capability under evaluation, so an aggregate "temporal answerability" score over it would flatter any extraction-capable arm regardless of merit. This is the concrete reason §15.2's per-class reporting requirement is load-bearing rather than procedural. Q2, Q4 and Q6 straddle classes and the splits are recorded rather than collapsed.

## Corrections the code forces on the PRD/TRD

Each carries `file:line` in the inventory. Flagging these because the ADR would otherwise inherit an inaccurate baseline.

1. **`work_graph_neighbors` caps at 25, not 100** (`MAX_NEIGHBORS = 25`), and lives in ops, not acr. §10 cites it as the precedent for a 100 cap on `related_changes`.
2. **The squash-merge premise is wrong.** A deliberate two-tier handler exists (CHAOS-2435). The real gap is narrower: residual squash formats drop **with no coverage signal anywhere**. Corpus C16 is written against the corrected premise — the failure under test is *silent* emptiness.
3. **There are two native history gaps, not one.** Beside declared state, `work_graph_edges` has `discovered_at` but **no `valid_to`**, and its RMT key omits `discovered_at`. CHAOS-3563 as scoped does not close this, and Q2 stays unanswerable if only the first is closed.
4. **Go workers already expose `/metrics` on :8080.** §18's blocking prerequisite is true for Celery only. If the projector is a Go worker it is already satisfied — this materially changes the CHAOS-3500/3503 dependency.
5. **Extraction budget governance is not a drop-in reuse.** The per-call guard is provider-neutral; the monthly cap tier is Ask-Dev-namespaced in its Valkey keys and recomputes from `dev_runs`.
6. **Org deletion's blind spot is wider than stated** — Valkey allowance counters and `worker_job_outbox` (no `org_id` column at all) are already unreachable today. Belongs in CHAOS-3566's scope.
7. **New defect found:** `operational_service_repository_mappings.valid_from` is `Nullable` while every as-of filter applies `valid_from <= as_of`, and `NULL <= anything` is false in ClickHouse — null-start rows are **silently dropped from every as-of answer**. An ops defect, not a graph finding; `O7_null_valid_from` measures it separately so no graph arm can take credit for surfacing it.
8. **A production transactional outbox already exists** (`worker_job_outbox`). The projector needs no new event mechanism.

Housekeeping: **CHAOS-3068/3069 tripwire is clean** — no fixture still lists `incidents.v1` in `expected_unavailable_sources`. **`deployments.v1` staleness is UNVERIFIED** — reported with the mechanisms that would confirm it rather than asserted, since checking needs an authorized environment.

## How the oracles are made trustworthy

Authored before any arm exists; every expectation derives from how the corpus was *constructed*.

- **A measurement that did not happen FAILS.** No skip path: `Verdict.NOT_MEASURED` is a failure, a raising arm becomes a recorded `NOT_RUN` rather than an aborted sweep, and `TrialReport` marks any aggregate holding an unmeasured oracle NOT COMPARABLE.
- **Every guard observed failing, in the right place.** Each fault carries the assertion id that claims to catch it and the suite demands the oracle fail *there*. Faults see the oracle, so they corrupt a fact the oracle actually requires; an inapplicable fault reports INAPPLICABLE, and a guard fails if any fault is inert corpus-wide.
- **Oracles cannot silently define correctness.** As-of expectations are re-derived by an independent walk of ground truth and compared.
- **The axis pair cannot be passed by luck.** Valid-time and observed-time have *different* correct answers at the same instant, and a test proves no single answer satisfies both oracles.

**Verification.** Suite driven RED first (46 failures — the harness correctly reporting faults landing where nothing asserted them), then green. Three defects then planted in the oracle engine — `must_include` always passes, `NOT_MEASURED` reads as success, direction ignored in identity matching — **all three KILLED, each dying in the guard that claims to cover it**.

```
204 passed, 136 skipped
```

Skips are inapplicable fault×oracle pairs; they are reported, not hidden.

## Test isolation — stated plainly

`trials/` is outside root `pytest.ini`'s `testpaths` and outside `[tool.mypy] files`, so `ci/local_validate.sh` does not collect it. Deliberate: that gate is **host-wide single-flight** and every worktree queues behind it, and shadow-trial code must not destabilise it. The cost is real and not papered over — these tests run only when someone runs them (`trials/chaos_3499/run_oracles.sh`), and an unrun suite reports as NOT MEASURED, never as absent. Ruff *does* cover `trials/`, so lint and format are enforced.

Repo-wide `mypy` and `ruff` are green (pre-commit ran both across all 2159 source files).

## Not in this PR, by design

- **No ADR outcome.** Adopt / native / do-not-adopt comes from the corpus results.
- **No arm implementations.** They follow review of 1–3, as separately reviewable changesets.
- **No compose file, no kind pod, no bring-up.** All gated.
- **No LLM extraction spend.** Not a token.
- **No numeric thresholds.** Observed distributions first, per §16.

## Open items for the reviewer

- Arm N depends on CHAOS-3562/3563/3564/3565. If they have not landed when the trial runs, class (a)/(b) results measure a *pre-increment* native arm and the ADR must say so.
- The §16 rebuild gate has no operational definition until CHAOS-3500 selects one.
- CHAOS-3566 is named a **precondition**, not a follow-up: the trial store is invisible to org deletion the moment it holds real org data (fine for synthetic corpus, not fine at Phase 2).
- The `entitlements` cache is an untimed revocation-lag path; credential revocation is live, entitlement revocation is not.


chrisgeo:
## Update: baseline-vs-arms framing corrected (8d9ef189d)

Per the amended §14 correction — the native work is **pre-trial increments feeding a native baseline**, not a competing entrant. Fixed in code rather than only in prose, so the shape can't drift back.

**Matrix now maps 1:1 onto the amended PRD:**

```
baseline (native increments + episode readback, composed)
    vs  Graphiti arm
    vs  direct-store arm
```

Per question class, as a delta. Episode readback is registered as a baseline **component**, not a candidate — `EpisodeArtifacts` already carries the subsystem signal, so if plain readback answers Q4 the graph's margin there is zero. Scoring it as a peer would let Graphiti be compared against native alone. A flat ranking would also let a candidate 'win' a class by placing above one baseline component while losing to the baseline as a whole.

**Baseline composition** (`runner.compose_baseline`), per oracle: any component passes → baseline passes; otherwise any component unmeasured → **NOT MEASURED**; otherwise FAIL.

The middle rule is the load-bearing one. If native was never run and readback failed, we know readback can't answer and we *don't* know whether native could — so "the product already does this" is genuinely unknown, and recording FAIL would hand a candidate a win it never earned. It ships with a paired control test (a fully-measured baseline that answers nothing must still record real FAILs), because without that control the same assertion would pass against a fold that never emits FAIL at all — which would make the baseline unbeatable and every delta meaningless.

**Class (b) now refuses to render as a number** until CHAOS-3563's branch state is recorded. `DependencyState` defaults to UNRECORDED (safe direction) and rejects a mismatched issue id. State comes through the orchestrator — another lane's uncommitted worktree is not a state anyone can cite in an ADR.

**Class (a) control failure is shouted**, not left as a positive delta a reader can page past. `ComparisonReport` exposes no single headline number and a test asserts it never grows one.

**Verification:** four further defects planted — unmeasured component folds to FAIL, dependency state ignored, class (a) control always 'holds', candidate arms folded into the baseline — **all four KILLED** in the right tests. A fifth attempt was recorded **INVALID rather than counted as a kill**: ruff had rewrapped its anchor so the mutated file was never built. Re-run correctly against the real anchor. Seven kills total across both rounds.

```
217 passed, 136 skipped
```

Repo-wide mypy + ruff still green.

**One deviation to flag:** directory is `trials/chaos_3499`, not `trials/chaos-3499` — the hyphenated form is not a valid Python package name, so corpus and harness could not import each other. Placement is otherwise exactly as approved, and noted in the README.

chrisgeo:
## High-effort review: 10 CONFIRMED findings — fix before arm implementation (track paused; recorded for resume)

**Correctness (9):**
1. `corpus/ground_truth.py:597` — select() never reads `VisibilityContext.redacted_source_event_refs`: the redaction half of C08 is dead; O4_prior_attempts_after_redaction passes an arm that serves redacted content verbatim. Redaction is currently unmeasured.
2. `corpus/oracles.py:618` — O5_CONFLICTS_INJECTED asserts only the `conflicting` flag; the C14 `untrusted_content` invariant is never asserted → injection laundering passes.
3. `harness/runner.py:135` — only `arm(oracle)` is guarded; an exception inside `oracle.evaluate(response)` (e.g. tz-naive watermark → TypeError in _assert_watermark) aborts the sweep and drops all remaining rows unrun AND unreported — the exact silent-coverage-loss the docstring claims this function prevents.
4. `tests/golden.py:262` — golden truncates to max_results by GROUND_TRUTH declaration order and always stamps `truncated=False`.
5. `harness/faults.py:254` — hide_coverage_gap mutates `gaps[0]` from the response instead of a source named in `oracle.coverage`; first 2-gap scenario produces a false red accusing a healthy oracle.
6. `tests/golden.py:202` — `invalidated_by.refs` built from the fact's own opening evidence, not the invalidating record — pins the amended-PRD invalidation-provenance semantics wrong from day one.
7. `run_oracles.sh:13` — `uv run pytest` omits the dev extra; the sole documented runner fails on any freshly-synced worktree.
8. `tests/test_oracle_fault_modes.py:65` — the 136-pair fault×oracle skip set has no pinned manifest; covered pairs can silently flip to INAPPLICABLE and coverage shrinks green.
9. `corpus/cases.py:81` — C03 construction prose ('blocks from 07-18 onward') contradicts the valid_from 07-05 / observed 07-20 backfill the axis pair depends on.

**Simplification (1):**
10. `tests/golden.py:244` — golden_response never passes query subjects to gt.select (dead param); subject-scoping defects in real arms are invisible.

~20 additional cleanup-tier findings live in the review journal (see orchestrator handoff). After fixes: re-run the mutation-kill sweep (non-building/stale-anchor = INVALID, not KILLED).

chrisgeo:
## FIXED: all 10 CONFIRMED findings (track resuming)

Env: `uv sync --all-extras --dev` first (per lane instructions -- editable installs in worktrees resolve stale origin src otherwise).

| # | Finding | SHA | RED evidence |
|---|---|---|---|
| 1 | `ground_truth.py:597` -- C08 redaction dead (`redacted_source_event_refs` never read) | `de62631b6` | Reproduced: old `select()` left `sevt_ep_0001` in the returned fact's `source_event_refs` after redaction; new `forbid_source_event_refs` qualifier + `restore_redacted_source_ref` fault mode both catch it now. |
| 2 | `oracles.py:618` -- O5_CONFLICTS_INJECTED never asserted `untrusted_content` | `de62631b6` | `test_conflicts_injected_requires_untrusted_content_not_just_conflicting`: against the old oracle definition (only `require_flags={"conflicting"}`), a response with `untrusted_content` stripped evaluates to **PASS** (confirmed via direct repro). Now FAILs. |
| 3 | `runner.py:135` -- exception inside `oracle.evaluate()` aborted the sweep past the NOT_RUN guard | `da0390a8f` | Reproduced directly: old `run_oracle` body raised `TypeError: can't compare offset-naive and offset-aware datetimes` after **0/20** oracles when fed a tz-naive `indexed_through`. `test_runner.py::test_oracle_evaluation_exception_does_not_abort_the_sweep` pins that all 20 results now return, crashing ones failing loudly under `oracle_evaluation_crashed`. |
| 4 | `golden.py:262` -- truncation followed GROUND_TRUTH declaration order by accident; `truncated` always `False` | `de62631b6` | `truncated` was hardcoded `False`, trivially reproducible. `test_golden_response_declares_truncation_when_a_candidate_is_dropped` / `..._does_not_falsely_claim_truncation` pin both directions; selection is now adversarial-last (stable) before slicing so the correctness of `O4_prior_attempts_manipulated`'s golden response no longer depends on where facts happen to sit in the corpus tuple. |
| 5 | `faults.py:254` -- `hide_coverage_gap` mutated `gaps[0]` from the response, not `oracle.coverage` | `de62631b6` | Reproduced directly: on a synthetic 2-gap scenario (an unwatched gap ordered first), the old mutator hid the *unwatched* source and left the oracle's real coverage expectation (`work_graph_pr_commit`) untouched -- `oracle.evaluate()` on the "mutated" response returned **PASS**. `test_hide_coverage_gap_mutates_the_source_the_oracle_actually_checks` pins the fix. |
| 6 | `golden.py:202` -- `invalidated_by.refs` cited the opening record's own evidence, not the invalidating record's | `de62631b6` | `test_golden_response_passes[O3_supersession]` failed red before the fix: `"invalidation missing evidence refs ['ev1_adr_021'] (has ['ev1_adr_014'])"`. Fixed via `GroundTruthFact.invalidated_by_fact_key` (set on `gt_adr014_superseded` -> `gt_adr021_supersedes_adr014`) + `FactExpectation.require_invalidation_evidence_refs`. New fault mode `cite_opening_evidence_as_invalidation` covers the regression path. |
| 7 | `run_oracles.sh:13` -- `uv run pytest` omits the dev extra | `6914dda62` | Could not reproduce a hard failure in this sandbox's uv/lockfile state (pytest resolved even without the extra here), but the fix matches the repo's own documented invocation (`ci/local_validate.sh`'s `uv sync --all-extras --dev`) and removes the dependency on ambient environment state entirely -- now `uv run --extra dev pytest ...`. |
| 8 | `test_oracle_fault_modes.py:65` -- 136-pair (now 174, after two new fault modes) skip set had no pinned manifest | `de62631b6` | Not a behavioral bug to RED/GREEN -- a coverage-drift gap. Pinned the full `(fault_id, oracle_id)` INAPPLICABLE set explicitly; `test_inapplicable_pairs_match_the_pinned_manifest` fails on drift in either direction (manually verified by injecting a fake pair into the observed set on both sides). |
| 9 | `cases.py:81` -- C03 prose ("07-18 onward") contradicted its own axis-pair dates (actual `valid_from` 07-05) | `f38b07334` | Prose-only defect, not behavior-affecting (no test consumed the string), so no RED/GREEN cycle applies -- added `test_c03_prose_matches_the_backfilled_blocker_it_depends_on` as new coverage against future drift. |
| 10 | `golden.py:244` -- `golden_response` never passed `subjects` to `gt.select` (dead param) | `de62631b6` | Reproduced directly: `gt.select()` without `subjects` returned `repo_atlas_web`'s episode for `O4_prior_attempts` (query subjects = `repo_atlas_api` only) -- a leak no assertion caught. `test_golden_response_scopes_by_query_subject_not_just_predicate` pins the fix. `select()` now matches subject-OR-object; `O1`/`O3` are deliberately exempted via `Scenario.apply_subject_filter=False` (documented in code) because their query subject has no literal ground-truth edge to the required facts in this corpus -- a known, named gap rather than a silent one. |

**Mutation-kill sweep (re-run after fixes):**
- Two new fault modes added (`restore_redacted_source_ref` for finding 1, `cite_opening_evidence_as_invalidation` for finding 6), each confirmed to KILL at its claimed guard (`must_include`) on the oracle it targets, and confirmed INAPPLICABLE (not a silent no-op) everywhere else via `test_every_fault_mode_applies_to_at_least_one_oracle`.
- All planted-defect kills from the prior round (unmeasured-component fold, dependency-state-ignored, class-(a)-control-always-holds, candidate-folded-into-baseline -- `test_comparison_report.py`) re-confirmed passing against the current tree; no INVALID (non-building/stale-anchor) results in this round.
- No mutation died outside its claimed `expected_assertion_id` in this round.

**`run_oracles.sh` (green):**
```
229 passed, 174 skipped
```
(174 = the pinned INAPPLICABLE manifest, finding 8.)

ruff + mypy clean on all changed files (pre-existing, unrelated mypy noise in untouched `test_corpus_consistency.py` -- 22 errors from a `**dict` unpacking pattern, present before and after this branch, not part of any of the 10 findings).

Pushed tip: `de62631b6`.

~20 additional cleanup-tier findings from the review journal remain untouched per the orchestrator handoff -- not addressed here. Arm implementation remains held pending orchestrator review of these fixes, per the lane's MUST NOT list.

chrisgeo:
## Hardening round complete: 4 verifier findings closed, exemption tightened, process caveat addressed

Response to the independent verification of `de62631b6` (7 FIXED, 3 PARTIAL, 0 refuted). All items closed on `a4ef6eff3`.

| Item | Fix | RED evidence |
|---|---|---|
| NEW-1 (MEDIUM) — truncation regression test couldn't fail against the sort it named | Extracted `_adversarial_last` into a directly-testable primitive (`golden.py`). Two new tests: a direct unit test against a hand-reversed input (decoy first), and a `monkeypatch`-based end-to-end test that reverses `GROUND_TRUTH` itself. | Confirmed by temporarily swapping the sort for `return list(facts)`: both new tests fail; the old end-to-end-only test still passes (proving it really can't detect this class of regression, now documented as a weak companion). |
| NEW-2 (MED-LOW) — redaction checked only the `source_event_refs` channel | Broadened `forbid_source_event_refs`'s qualifier check to look at both `source_event_refs` and `evidence_refs`. New fault mode `smuggle_redacted_ref_via_evidence`. Also fixed the oracle label/rationale — the fact has exactly ONE `source_event_ref`, not "multi-source"; post-redaction it stands on its `evidence_ref` alone. | Reproduced directly: the old (narrower) check computes an empty intersection against a response with the redacted ref smuggled into `evidence_refs` — no problem detected, response would PASS. |
| NEW-3 (MEDIUM) — finding 10's leak was still invisible to real arms | Added `Oracle.require_subject_scoped` (default True) + `_assert_subject_scoped`: every returned fact's subject or object must be among `query.subjects`. New fault mode `leak_out_of_subject_fact` + a direct regression test with an explicit PASS control. | Reproduced the verifier's exact scenario directly: appending `gt_ep5_web_repo` to `O4_prior_attempts`'s golden facts evaluated to PASS before the fix, FAIL after. |
| NEW-4 (LOW) — `invalidated_by_fact_key` had no consistency guard | Added `SELF_EVIDENCING_CLOSURES` (pinned set, mirrors the existing `_DEGRADED_WATERMARK_ORACLES` idiom) + `test_closed_facts_declare_their_closure_provenance`: every closed fact must be in the pin XOR set `invalidated_by_fact_key`, never both, never neither. | Verified the guard fires: removed `gt_svc_repo_v1` from the pin, test failed with a precise message; restored it, test passed. |
| Exemption tightening | Verified empirically that 3 of the 6 previously-exempted scenarios (`O1_ci_prior_attempts_squash`, `O3_supersession_extraction_down`, `O3_supersession_deterministic_only`) are not load-bearing — flipping each individually to `apply_subject_filter=True` keeps its golden response green. Un-exempted those three; only `O1_ci_prior_attempts`, `O1_ci_prior_attempts_stale`, `O3_supersession` remain exempt (now also reflected in `Oracle.require_subject_scoped`, kept consistent between the two mechanisms). | N/A — verified via direct flip-and-rerun against all 6, not a RED/GREEN behavioral fix. |
| O3 comment fix | Corrected: ADR-014 DOES have an object edge to `proj_atlas` (`describes_deployment_design_for`); only the `supersedes` edge (ADR-021→ADR-014) has neither side as `proj_atlas`, and that's what forces the exemption for the whole oracle. | N/A — wording only. |
| Finding-8 process caveat | Re-verified the manifest drift test using genuinely-absent and genuinely-present pairs on both directions (shrink and growth) — both correctly detected. The prior round's verification used a pair already in the pinned set (an INVALID mutation reading as verification, per the caveat). | Confirmed both directions with real (not-already-pinned) pairs this time. |

**Manifest regenerated** for the two new fault modes: 174 → 197 INAPPLICABLE pairs.

**Full suite:** `run_oracles.sh` green at **250 passed, 197 skipped**.

ruff + mypy clean on all changed files — `test_corpus_consistency.py`'s 22 pre-existing, unrelated mypy errors (the `**dict`-unpack typing pattern) are unchanged in count; I initially introduced one new error there while writing the NEW-4 test (a loop-variable type-narrowing conflict) and fixed it before committing (renamed the shadowing variable) — verified back to exactly 22.

Pushed tip: `a4ef6eff3`.

chrisgeo:
## Sub-blocking findings from the arm-commits review — documented, not fixed (per nitpick policy)

The high-effort review verified 25 candidates; 10 blocking ones went to the fix ledger. These 8 lower-severity items are recorded for the record, deliberately NOT fix-committed:

- **harness/arms/native.py:100** (PLAUSIBLE) — "ground_truth.py doubles as the fixture" drops the old bring-up wording's separation between arm input and oracle answer key; `_to_fact` copies evidence_refs/source_event_refs verbatim from the answer key, which is fine for baseline components but worth restating as a boundary when candidate arms are compared.
- **tests/test_extraction_smoke.py:47** — identical live LLM completions re-issued across tests (3 reachability probes + repeated extraction.answer for the same oracles) — wasted local inference time per run.
- **tests/test_baseline_components.py:157** — leftover self-correcting draft prose in a docstring ("more recent than the decoy... no, LESS recent").
- **tests/test_extraction_smoke.py:123** — garbled assertion message ("returned {outcome}, neither of which a real arm may" with a single excluded member).
- **harness/llm/client.py:119** — `except (APIConnectionError, APIError)` is redundant (the former subclasses the latter in openai v1).
- **harness/arms/episode_readback.py:74** — `_to_fact`/`_degraded` are near-verbatim copies of native.py's; the mapping now lives in two places.
- **harness/arms/native.py:156** (PLAUSIBLE) — `_answer_q2_current_state` ignores query.subjects/allowed_relation_types/org scope/max_results, unlike the real `_BLOCKERS_SQL` it claims to reproduce — the deliberate-defect reproduction is narrower than its label.
- **harness/arms/extraction.py:130** (PLAUSIBLE) — `fact_id` built only from (subject, predicate, object) collides when two documents legitimately state the same triple with different evidence.

Promote any of these to the ledger if they become load-bearing for the measured trials (the native.py:156 and extraction.py:130 ones are the likeliest candidates).

chrisgeo:
## Arm-commits pre-merge review: 10/10 measurement-integrity findings fixed

Response to the pre-merge review of the two arm commits (25 verified candidates → 10 reported, 9 CONFIRMED + 1 PLAUSIBLE). All 10 closed on `5bbe2bbdb`.

| # | Finding | Fix | RED evidence |
|---|---|---|---|
| 1 | `claim_kind` defaulted to `observed` when missing | A row missing `claim_kind` is now MALFORMED and dropped, like a missing `subject_id` | Direct repro: old code accepted a row with no `claim_kind` and silently granted it `observed`. |
| 2 | Cloud guard was name-based; `LOCAL_LLM_BASE_URL` could point at a paid endpoint under `provider=local` | Host allowlist (`localhost`, `127.0.0.1`, `host.docker.internal`) enforced in `LLMConfig.from_env`; anything else raises `LLMUnavailable` naming the reason | `test_local_provider_rejects_a_real_endpoint` — confirmed a substring trick (`evil.localhost.attacker.example`) is also rejected, not just an obvious mismatch. |
| 3 | O3's closure requirement was only satisfiable via adapter-side `_propagate_supersession_closure` — the model's schema couldn't express it | Deleted that function entirely. Model schema gained an optional per-fact `"closes"` block (target identity + `closed_at` drawn from source text); adapter propagates ONLY what the model emitted. Added a real date (`Decided: 2026-07-09`) to ADR-021's source text so the model has a genuine chance. | Re-run smoke: O3 now FAILs because the model doesn't express `describes_deployment_design_for` or a `closes` block at all — an honest, real measurement. |
| 4 | Greedy `\[.*\]` regex could span past the real array into unrelated trailing content with another bracket | Replaced with `json.JSONDecoder.raw_decode` scanning from each `[` until the first one parses | `test_extract_json_array_stops_at_the_first_valid_array` — old regex on `"[1, 2] but also note [not json at all]"` captured the whole span; new code returns exactly `[1, 2]`. |
| 5 | Ref-passthrough test was vacuous (shape-only check, early return after first success) | Rewritten: a FAKE client response with an INVENTED evidence_ref is fed through the REAL adapter for every smoke oracle, no early return; asserts the ref survives verbatim AND the oracle FAILs | Needs no live model — `test_fabricated_evidence_ref_survives_verbatim_and_fails_the_oracle`, both smoke oracles covered. |
| 6 | `valid_to` was stamped with the trial's own indexing time | Resolved by finding 3's fix — closure now only ever comes from the model's own emitted date, or the fact stays open | Same as 3. |
| 7 | Smoke inherited ambient `LOCAL_LLM_BASE_URL`/`LOCAL_LLM_MODEL` | `_configure_local` now pins all three vars (+ clears `LOCAL_LLM_API_KEY`) to documented defaults; actual `base_url`+`model` printed in the SMOKE observation line | Confirmed live during dev: this repo's shell had `LLM_PROVIDER=openai` set for unrelated reasons; the old code would have silently used whatever ambient base_url/model was also set. |
| 8 | `max_retries` relied on SDK default | Set `max_retries=0` explicitly on client construction | `test_complete_constructs_client_with_max_retries_zero`, verified RED by temporarily removing the kwarg. |
| 9 | `response.choices[0]` would `IndexError` on empty choices | Guarded, raises the same `LLMUnavailable` as any other provider failure | `test_complete_raises_on_empty_choices`, verified RED by temporarily removing the guard (confirmed `IndexError`). |
| 10 | An arm's role was chosen entirely by the call site — nothing stopped `extraction_llm` registering as baseline, or `native`/`episode_readback` as candidate | Each arm's `answer` now declares `answer.declared_role`; `ArmRegistry.register` raises on a mismatch in either direction | `test_registry_rejects_extraction_arm_registered_as_baseline` / `..._rejects_baseline_arm_registered_as_candidate` / `..._accepts_arms_registered_under_their_declared_role` (control). |

**Bonus finding surfaced while re-testing after these fixes**, fixed in the same commit: the model occasionally emits JSON `null` for a required field, and `str(None)` was silently turning that into the literal string `"None"` instead of being rejected as malformed — a null-coerced-to-string entity id slipping through undetected is finding 1's failure mode one level lower. Added `_required_str`, used everywhere a required field is read (fact rows and closure-target rows), rejecting non-string/empty/null explicitly.

**Smoke re-run after all fixes (stable across repeated runs), honestly updated per the review's instruction:** `O3_supersession` still FAILs, but the reason has moved. Pre-fix: a direction-reversal defect (`describes_deployment_design_for` extracted reversed). Post-fix: the `supersedes` edge is now extracted correctly, in the right direction, citing the real `ev1_adr_021` ref — but the model doesn't reliably extract `describes_deployment_design_for` at all (rows come back with the now-rejected null-field pattern) and never emits a `"closes"` block despite the source stating an explicit date. **The direction-reversal defect did not reproduce under the current adapter and is superseded by this finding, not additive to it** — `harness-design.md` updated accordingly. `O5_conflicts_injected` still PASSES outright, unaffected.

**Full suite:** 285 passed, 197 skipped. One run during verification hit a transient provider timeout on `O3_supersession` mid-suite — correctly surfaced as `NOT_RUN`, not a crash, exactly the discipline requirement 6 exists to enforce. ruff + mypy clean on all changed files.

Pushed tip: `5bbe2bbdb`.

chrisgeo:
## Spot-check gap closed: finding 1 pinned

Your spot-check was right — the finding-1 fix (`5bbe2bbdb`) had no failing test; re-planting `raw.get("claim_kind") or "observed"` left the entire offline suite green.

Added `test_malformed_rows_are_dropped_not_repaired` on `bf4eb341c` — drives the real adapter through a fake completion (no live model needed): (a) missing `claim_kind` → dropped, (b) JSON `null` in a required field → dropped, never becomes the string `"None"` (pins the bonus `_required_str` fix, which was equally unpinned), (c) well-formed row → survives as the control.

RED confirmed twice, matching your exact repro: re-planted `raw.get("claim_kind") or "observed"` → new test fails on (a) while the rest of the offline suite (275 tests) stays green; separately re-planted the pre-`_required_str` `str(raw["object_kind"])` pattern → new test fails on (b). Both restored after confirming.

Full suite: 275 passed (offline) / 11 passed (live smoke) / 197 skipped. ruff + mypy clean.

Pushed tip: `bf4eb341c`.

chrisgeo:
## CodeQL alerts fixed: all 5 investigated, none dismissed

| Alert | Disposition | Reasoning |
|---|---|---|
| `py/non-iterable-in-for-loop` x4 (`runner.py:110` + 3 test-file sites) | **Analyzer-appeasement restructure** | Every site is `for klass in QuestionClass:`, the standard idiom for enumerating a `(str, Enum)` mixin's members — correct, and confirmed not a latent bug: each site references the literal `QuestionClass` name directly, never a variable that could hold something else. CodeQL can't prove `EnumMeta.__iter__` makes the mixin iterable. Added `ALL_QUESTION_CLASSES: tuple[QuestionClass, ...]` to `contracts.py` as a literal tuple — provably iterable to any analyzer — and replaced the pattern at all 9 occurrences in the trial (not just the 4 flagged, for consistency), including `runner.py`'s production code. |
| `py/unused-global-variable` (`corpus/oracles.py:30`, `_PROJECTION`) | **Real finding, chased to root** | The literal it held (`"temporal-projector.v1"`) turned out to be duplicated 4 further times: `golden.py` independently redeclared it, `faults.py` and `test_oracle_fault_modes.py` each hardcoded it in two `TemporalFact` constructions. No oracle assertion checks `projection_version` at all, so nothing would catch these 5 copies drifting apart on a version bump. Moved the canonical constant to `harness/contracts.py` as public `PROJECTION_VERSION` (the modules that need it already sit below `contracts.py`, so this doesn't invert the corpus-depends-on-harness layering) and wired all 3 consumers to import it instead of redeclaring/hardcoding. |

Full suite: 275 passed (offline) / 11 passed (live smoke) / 197 skipped — unchanged, as expected for a pure restructure with no semantic change. ruff + mypy clean.

**CI confirmed green on the new head**, `gh pr checks 1578`: CodeQL pass, Analyze (python) pass, lint/typecheck/test/governance/changes all pass, everything else skipping (path-gated, not failing).

Pushed tip: `54535ac6f`.

PR #1588, opened 2026-08-07
CHAOS-3551: render committed-cohort answers instead of refusing

chrisgeo:
## Summary

- A bounded, homogeneous, **complete** `REPOSITORY` cohort now `PROCEED`s and answers instead of terminating `feature_not_enabled` (D1). `subject_preflight`'s homogeneous-cohort branch reuses `committed_cohort_resolution_for` (CHAOS-3534) — previously wired only to disclose the correct scope outcome on an honest `TERMINATE` — to commit a real multi-repository `DevScopeResolution` and `PROCEED` with it, exactly like a singular commit. No new rendering machinery was needed: the ordinary legacy model round loop answers over the committed multi-repository scope through the ordinary `status_snapshot.v1` path (`DevScope.repositories` is already list-typed and the tool layer already handles it).
- Scope is deliberately narrow: only `EntityKind.REPOSITORY` has a v1 multi-entity `DevScope` representation (`committed_cohort_resolution_for` itself refuses every other kind — a structural guarantee, not a policy check made twice), so project/team cohorts keep terminating unsupported exactly as before. Gated on `cohort_complete` (a partial cohort keeps refusing) and `V1_SCOPE_LIST_LIMIT` (an oversized cohort keeps refusing).
- **Person-level output**: there is no `EntityKind.PERSON` anywhere in this catalog (`EntityKind` is organization/repository/project/work_unit/issue/pull_request/team), so this fix cannot open person-level output — there's no person-kind subject to leak. A `TEAM` cohort (the closest available proxy — a "who" cohort) is pinned to keep refusing.
- **Also fixes a latent bug this change made reachable**: the orchestrator's `PLURAL_COHORT`/`ORGANIZATION_WIDE` plan-eligibility gate checked only "is there a plan for this intent" and "is a subject set committed" — never the committed subject set's own *kind* against the plan's declared `supported_subject_kinds` (a field that existed but was never read). A repository cohort under a question whose vocabulary classifies as `PORTFOLIO_STATUS` (intent classification is lexical, not kind-aware) would otherwise have been handed to `status.portfolio.v1`'s step, which builds each batch entry via `_project_scope_from_ref` — documented as trusting its *caller* to restrict entries to `PROJECT` — misrepresenting a repository as a same-named project. The plan executor's own defensive fingerprint check turns that into a loud `internal_error` rather than a silent misattribution (verified — see below), but a crash is still not "render an answer". The gate now checks kind too.

## The exact seam changed

- `src/dev_health_ops/api/dev/subject_preflight.py` — new `PROCEED` branch in `SubjectPreflight.run()`'s homogeneous-cohort handling (right after the existing `committed_cohort_portfolio_v1` special case, before the `committed_cohort_v1_only` terminate), diagnostic `committed_cohort_v1_render`.
- `src/dev_health_ops/api/dev/orchestrator.py` — `plan_eligible` computation for `PLURAL_COHORT`/`ORGANIZATION_WIDE` now also checks `preflight_result.subject_set.entity_kind in plan.supported_subject_kinds`.

## RED-first proof

`tests/api/dev/test_chaos_3551_cohort_render.py` is new. Reverting just the `subject_preflight.py` hunk (`git stash` locally) reproduces RED exactly as described:
- `test_a_committed_repository_cohort_answers_instead_of_refusing`: fails — `preflight_diagnostics == [("committed_cohort_v1_only", None)]` instead of the new render diagnostic (old code still terminates).
- `test_a_committed_repository_cohort_withholds_resolve_scope`: fails — `decision is TERMINATE`, not `PROCEED`.
- The two negative-control pins (`test_a_team_cohort_still_refuses_after_the_render_fix`, `test_a_partial_repository_cohort_still_refuses`) already pass on old code — proving they're not vacuous.

Reverting just the `orchestrator.py` hunk (with the `subject_preflight.py` fix still applied) reproduces the *second*, latent-bug RED: `test_a_repository_cohort_never_reaches_the_portfolio_plan` fails with `RunState.FAILED` / `internal_error`, raised from `PlanExecutor.run()`'s own `subject_set_scopes do not match the caller's authorized subject_set_fingerprint` guard — confirming the plan executor's own defensive check prevents *silent* misattribution, but the run still crashes instead of answering, until the `orchestrator.py` gate is fixed too.

Also updated two pre-existing tests whose literal expectations this change *correctly* flips (both explicitly reference CHAOS-3551 in an updated docstring/comment):
- `test_chaos_3534_cohort_scope_outcome.py::test_a_fully_resolved_cohort_publishes_exact_not_unresolved` and `::test_a_cohort_at_exactly_the_v1_limit_still_publishes_exact` — same exact corpus question now `PROCEED`s+answers rather than terminating; the resolution-outcome assertions those tests actually exist for (`EXACT`, both repo ids) are unchanged and still pass.
- `test_chaos_3533_terminate_ledger.py`'s two DB-backed tests that drove the *exact* two-repository corpus question through a real Postgres-shaped ledger-persistence TERMINATE — retargeted to an equivalent PROJECT cohort (still terminates, since only REPOSITORY renders), preserving the persistence mechanism under test.

## Test counts

Targeted files (`test_chaos_3551_cohort_render.py`, `test_chaos_3534_cohort_scope_outcome.py`, `test_chaos_3533_terminate_ledger.py`, `test_subject_preflight.py`, `test_chaos_3393_portfolio_preflight.py`, `test_chaos_3393_portfolio_orchestrator.py`, `test_chaos_3301_controls.py`, `test_chaos_3301_review_findings.py`, `test_chaos_3366_not_found_fallback.py`, `test_chaos_3292_mutations.py`, plus the plan-registry suites): **189/189 passed, 0 skipped, 0 failed**.

Full `ci/run_tests.sh unit` (real command, `-m "not benchmark and not clickhouse"`, xdist `-n auto`): **13343 passed, 245 skipped, 1 xfailed, 12 failed**. The 12 failures are pre-existing and unrelated — confirmed via an identical run on the unmodified branch (same 12: 11× `tests/api/admin/test_org_deletion.py` needing a live Postgres, 1× `tests/docs/test_built_site_links.py` needing `mkdocs`, not installed in this venv). Passed count is baseline (13338) + exactly the 5 new tests this PR adds.

`ruff check`/`ruff format --check`/`mypy` clean on every changed file (also enforced by the repo's own pre-commit hooks, which ran clean on this commit).

## Not covered here

- The live acceptance-corpus flip for `scope.bounded-subject-set`'s `public_outcome_in` invariant happens on the phase3-corpus lane's branch (a separate, unmerged lane per the ticket), not here — this PR is the product fix the corpus case measures, not the corpus run itself.
- `subject_set_ref` on the resulting `DevAnswerFrame` stays unset for this render path (as it already is for every non-portfolio `PROCEED` today) — CHAOS-3551's own text calls this a `wrap_legacy_answer_as_frame` touch point, but wiring it up would mean threading a new free variable through the orchestrator's ~35 `finish()` call sites for a disclosure improvement, not the render/terminate defect itself. Flagging as a real, deliberately-deferred gap rather than fixing it under this ticket's narrower scope.
- The `SINGULAR` cardinality branch's `plan_eligible` gate has the same *theoretical* missing-kind-check shape the `PLURAL_COHORT` branch had — but nothing in this change adds a new way to reach it with a mismatched kind (only pre-existing singular-commit paths reach it, unchanged by this PR), so it's left alone rather than spec-creeping a second, unrelated hardening fix into this diff.
- A pre-existing, unrelated fixture-shape gap: the shared test harness's stock `answer_payload()` fixture's `metrics[0].evidence_ref_ids` is too short for `DevMetricRefV2`'s stricter v2 validation, so `wrap_legacy_answer_as_frame` silently falls back to an `internal_error` *frame* (never affecting `result.answer`/`result.error`) for **any** completed run using that fixture — reproduced on `case_a1` (a pre-existing, currently-passing singular-commit test), so it predates and is unrelated to this change. Not fixed here.

## Gate status

**The local/CI gate has NOT been run** (`ci/local_validate.sh`) — machine-wide single-flight, reserved; queued for the orchestrator's slot per instructions. No merge attempted.

### DIFF
--- a/src/dev_health_ops/api/dev/orchestrator.py
+++ b/src/dev_health_ops/api/dev/orchestrator.py
@@ -2570,8 +2570,35 @@ class DevOrchestrator:
                     # is nothing to batch over without a committed
                     # DevSubjectSet, mirroring the SINGULAR branch's
                     # has_committed_subject gate one line up.
+                    #
+                    # CHAOS-3551: and the committed subject set's own KIND
+                    # must be one `plan` actually supports. Before this
+                    # ticket every intent whose vocabulary could classify as
+                    # PORTFOLIO_STATUS only ever reached here with a PROJECT
+                    # subject_set -- subject_preflight's own homogeneous-
+                    # cohort gate committed nothing else under that intent.
+                    # CHAOS-3551 adds a second PROCEED source for the SAME
+                    # (PORTFOLIO_STATUS-classified, PLURAL_COHORT) shape: a
+                    # REPOSITORY cohort, which intent classification alone
+                    # cannot rule out (it is lexical, not kind-aware -- see
+                    # subject_preflight's own PORTFOLIO_STATUS branch
+                    # comment). Without this check, a repository cohort
+                    # phrased like a status question would reach
+                    # `_project_scope_from_ref` -- documented as relying on
+                    # ITS CALLER to restrict `ref.entity_kind` to PROJECT --
+                    # and misrepresent each repository as a same-named
+                    # "project" scope to `PortfolioStatusService.
+                    # evaluate_portfolio`, exactly the "answer about one
+                    # entity under another's name" this module's own
+                    # docstring says it never does. `plan.
+                    # supported_subject_kinds` already declares the right
+                    # answer for every plan; nothing previously read it.
                     plan_eligible = (
-                        plan_eligible and preflight_result.subject_set is not None
+                        plan_eligible
+                        and plan is not None
+                        and preflight_result.subject_set is not None
+                        and preflight_result.subject_set.entity_kind
+                        in plan.supported_subject_kinds
                     )
                 if plan_eligible:
                     assert plan is not None
--- a/src/dev_health_ops/api/dev/subject_preflight.py
+++ b/src/dev_health_ops/api/dev/subject_preflight.py
@@ -42,6 +42,7 @@ from datetime import UTC, datetime
 from enum import StrEnum
 
 from .contracts import (
+    V1_SCOPE_LIST_LIMIT,
     DevContractVersions,
     DevScope,
     DevScopeResolution,
@@ -149,6 +150,12 @@ PREFLIGHT_DIAGNOSTICS: tuple[str, ...] = (
     # PROCEED diagnostics -- see the two call sites' own comments.
     "committed_cohort_portfolio_v1",
     "committed_portfolio_org_wide",
+    # CHAOS-3551: a homogeneous, complete, v1-sized REPOSITORY cohort
+    # PROCEEDs and renders over the multi-repository scope
+    # `committed_cohort_resolution_for` (CHAOS-3534) already built --
+    # the render half of D1, mirroring `committed_cohort_portfolio_v1`
+    # for every non-portfolio intent.
+    "committed_cohort_v1_render",
     # CHAOS-3528: the CHAOS-3393 portfolio catalog-outage branch's own
     # diagnostic, emitted at a real call site but never added here -- the
     # exact CHAOS-3292 class this tuple exists to prevent.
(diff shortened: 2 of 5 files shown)


chrisgeo:
### Review record (non-blocking)

Adversarial review complete: no behavior defects found. Three documentation-level residuals noted for the record, deliberately NOT fix-round-tripped (team policy: document and move forward):

1. `subject_preflight.py` `SubjectPreflightResult.committed_resolution` field docstring still says a committed cohort never proceeds to tool execution — now false for REPOSITORY cohorts (the sibling `CommittedSubjects` docstring was updated; this one was missed).
2. `docs/contribute/architecture/ask-dev-subject-preflight.md` outcomes table still describes bounded cohorts as "Not supported yet" / `committed_cohort_v1_only` — stale for REPOSITORY cohorts. Can ride the next touch of that page.
3. `orchestrator.py` ~2196-2238: the CHAOS-3534 TERMINATE-side REPOSITORY-cohort branch is now unreachable (this PR's PROCEED branch intercepts every case matching its condition). Dead code, else-fallback remains correct.

Corpus annotation staleness for `case-scope.bounded-subject-set.json` is a disclosed scoping decision — the flip lands on the phase3-corpus branch.

PR #1593, opened 2026-08-07
CHAOS-3574: cross-tenant organization references classify as not_found

chrisgeo:
## Summary

`adv.cross-tenant.organization-id` ("What's the status of the Orbit organization?") produced `public_outcome='answered_with_gaps'` instead of the corpus case's required `not_found`, observed twice on the Phase 3 armed corpus run (degraded + measured, identical clause). Not a leak — `no_unauthorized_candidate_surfaces` passed both times — this was an outcome-classification defect.

**Why it happened:** "Orbit" is the sibling tenant's real display name, but an organization is not an `EntityKind` and never a searchable catalog entity, so `question_interpreter` could never *type* the mention (no `organization` noun in the closed `_KIND_NOUNS` table). The mention fell into `subject_preflight`'s `unresolved_untyped` branch, which by design widens to organization scope and answers instead of terminating — correct behavior for an ambiguous bare word like "Zephyr" or "DORA" (pinned by the corpus case `scope.outcome.organization-fallback`), wrong for a name that unambiguously claims to *be* an organization.

**Fix:**
- `question_interpreter.organization_mention_spans`: a small, separate recognizer for a bare name adjacent to "organization"/"org" — deliberately *not* added to `_KIND_NOUNS` (no new `EntityKind`, no new catalog search; organizations are never searchable).
- `subject_preflight`: when an unresolved untyped mention matches an organization-adjacent span, terminate on the *same* per-mention ledger entry the typed no-match path already uses (`PREFLIGHT_OUTCOME_BY_RESOLUTION`) instead of widening. A cross-tenant org name and a nonexistent one now produce byte-identical public outcome and frame shape — pinned by a dedicated indistinguishability test.

Ordinary unresolved bare names with no organization noun nearby (`Zephyr`, `DORA`, `Nightfall`) are untouched and still widen — regression-pinned.

## Test plan

- [x] RED-first: reverted the two source files, confirmed the new preflight tests failed for the traced reason (`decision=PROCEED`, `legacy_guard_required=True`, `unresolved_name_spans={'orbit'}`), then restored the fix and confirmed GREEN.
- [x] `tests/api/dev/test_subject_preflight.py` + `tests/api/dev/test_question_interpreter.py`: 94 passed, 0 skipped.
- [x] Full `tests/api/dev/` (CI-parity unit selection for this area): 2467 passed, 34 skipped (pre-existing, unrelated), 1 xfailed (pre-existing).
- [x] `ruff-format` / `ruff-check` / `mypy` (pre-commit + pre-push hooks): all passed, no issues across 2172 source files.
- No gate run, no merge (per lane scope) — corpus case `case-adv.cross-tenant.organization-id.json` already expects `not_found` and is owned by the phase3-corpus branch; it will confirm on the next armed run.

### DIFF
--- a/src/dev_health_ops/api/dev/question_interpreter.py
+++ b/src/dev_health_ops/api/dev/question_interpreter.py
@@ -408,6 +408,18 @@ _ORGANIZATION_NOUN_TRAILING = re.compile(
     rf"(?P<noun>(?i:{_ORGANIZATION_NOUN_ALTERNATION}))\b"
     rf"{_ORGANIZATION_ATTRIBUTIVE_GUARD}",
 )
+#: Same continuation list, applied to the LEADING form (CHAOS-3574 review
+#: round 3, CONFIRMED): there the idiom word is not something that FOLLOWS
+#: the noun, it IS the captured name -- "Org Chart" / "Organization
+#: Structure" satisfy ``_NAME`` (a leading capital is all it requires, not a
+#: real proper noun) exactly as a genuine org name like "Org Meridian" would.
+#: Filtered in :func:`organization_mention_spans` rather than the regex
+#: itself: a lookahead here would need to duplicate the alternation inside
+#: the existing capture group, and a plain post-match set check is simpler
+#: and exactly as precise.
+_ORGANIZATION_ATTRIBUTIVE_CONTINUATIONS_SET = frozenset(
+    _ORGANIZATION_ATTRIBUTIVE_CONTINUATIONS
+)
 
 
 def organization_mention_spans(question: str) -> frozenset[str]:
@@ -436,6 +448,12 @@ def organization_mention_spans(question: str) -> frozenset[str]:
             ]
             words = [word for word in words if word]
             if not words or all(word in _NAME_STOP_WORDS for word in words):
+                continue
+            if (
+                pattern is _ORGANIZATION_NOUN_LEADING
+                and len(words) == 1
+                and words[0] in _ORGANIZATION_ATTRIBUTIVE_CONTINUATIONS_SET
+            ):
                 continue
             found.add(_normalize(span))
     return frozenset(found)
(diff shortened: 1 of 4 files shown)


chrisgeo:
## Review round 3 fix (e87de040e)

Closed the leading-form organization idiom gap the reviewer flagged (confidence 92, structural). Repro'd first per instruction: `organization_mention_spans("What's on the Org Chart?")` did return `{"chart"}`, confirming the function-level defect. Executing the full path further showed that exact phrasing was not actually reachable through `subject_preflight`'s real mention extraction (`untyped_name_candidates` merges "Org Chart" into one two-word bare-name mention, which never equals the noun-stripped single-word span regardless of any guard — a segmentation mismatch between two independent regex passes). The genuinely reachable phrasing is a **lowercase** org-noun immediately before a **capitalized** continuation word ("the org Chart status") — RED-first confirmed against that phrasing specifically. Fixed at the `organization_mention_spans` function level regardless, so it's correct for any caller, not just the one reachable path found today.

## Two document-only notes from the reviewer (no code change)

1. **The singular-commit path's `unresolved_mention_ids`/`ambiguous_mention_ids` warning blocks (`singular_unresolved_ids`/`singular_ambiguous_ids` in `subject_preflight.py`) are currently dead code — always empty.** By the time execution reaches the singular-commit path, any blocking (typed) unresolved mention has already terminated the run earlier in `run()`, so the `blocking_ids`-scoped computation added there never actually populates. It's harmless (correct no-op, mirrors the cohort branch's own scoping for consistency) and becomes live the day a future change lets a blocking-unresolved mention reach this path alongside a resolved subject — kept for that symmetry rather than removed, but it isn't exercised today.

2. **The possessive exclusion ("the Atlas organization's status" no longer treated as unambiguously naming an organization) is a deliberate product call, not a coverage gap.** It's intentional and tested (see `test_organization_mention_spans`'s possessive case), and the pre-existing CHAOS-3289 legacy backstop still catches it if the model's answer actually narrates the unresolved name.

Full `tests/api/dev/` suite: 2483 passed, 34 skipped (pre-existing/unrelated), 1 xfailed (pre-existing). mypy/ruff clean.

PR #1595, opened 2026-08-07
CHAOS-3546: fixture-field consumer guard; wire fault_ref

chrisgeo:
## Summary

The confirmed class: a fixture/case field declared as configuration, consumed by no code path, making a scenario look configured when nothing configures it. Three instances the ticket names: `ASK_DEV_SCRIPTED_PROVIDER_SCRIPTS_DIR` (already fixed in #1567), `fault_ref`, and `provider_profile_override`.

- **Guard test** (`tests/acceptance/corpus/test_fixture_field_consumers.py`): a registry of every fixture field this repo has declared operationally meaningful, each paired with evidence a real code path consumes it — generalizes #1567's own narrow (one env var, one digest) conformance assertion into a repeatable, extensible pattern.

- **`fault_ref` — RETIRED (not wired).** Measured: its value equalled the case's own `id` in all 18 cases that declared it, with zero exceptions — never a pointer to anything, a redundant self-referential marker. Deleted from the 18 case JSONs; `CASE-SCHEMA.v1.md`'s row removed (replaced with a removal note — the key still appears as inert `"fault_ref": null` on other case files pending a bulk cleanup sweep, out of this ticket's scope, called out explicitly rather than silently left undocumented). Replaced with a DERIVED cross-check needing no field on either side: every provider-scripts registry key whose kind is `fault`/`decisions` must name a real corpus case. RED-first proven against a **synthetic** mapping — the real `provider-scripts/role-legacy_agent.json` is untouched (verified: zero hits under `provider-scripts/` in this diff) because its identity digest is pinned by the in-flight phase3 armed run for the duration of this ticket.

- **`provider_profile_override` — RETIRED (not wired).** Traced the full chain: `AgentProviderPolicy` already has the real mechanism (`denied_models`/`denied_providers`, read at `policy.py:151-158`, firing before `BoundedDevRuntime` construction), but nothing populates it — `production_runtime._provider_candidates` builds that policy from org settings only, and no acceptance-boot step writes the generic settings table a denial would live in. Deeper: the field has zero references in `fixtures/world.py` — no DB column, no migration, no per-user threading anywhere in production. Wiring it means inventing a new production feature surface, not connecting an existing seam — filed as **CHAOS-3588** (settings-seeding + org-vs-user scoping decision), out of this ticket's scope. Field deleted from `world.json` (both user entries). **Three** corpus cases it drove flip to `declared-blocked` on CHAOS-3588 — `deg.provider.unsupported`, `readiness.capabilities.unsupported-model`, and **`readiness.capabilities.degraded`** (a third case found while editing its two siblings — same `provider_profile_override: degraded-profile` / `primary.degraded-readiness-user` pattern, missed by the ticket's original two-case scope; its own authoring notes already flagged the field as unconsumed at authoring time). All three were PASSING while proving nothing about unsupported-model/degraded-readiness handling before this change — declared-blocked is the honest terminal state.

## Test plan

- [x] RED-first throughout:
  - Guard registry: positive controls (`resolution_profile_ref`/`org_alias`) pass; negative control (a fabricated field name) fails; anti-self-reference guard pinned directly.
  - `fault_ref` cross-check: proven against a planted synthetic mismatch (`test_scripted_registry_cross_check_catches_a_planted_mismatch`), never against the pinned real file.
  - `provider_profile_override` retirement: stashed the corpus/world.json edits, confirmed both new tests (`test_provider_profile_override_is_retired_from_world_json`, `test_the_provider_profile_override_cases_are_declared_blocked`) failed exactly as expected (field present / cases still `active`), restored, confirmed green.
- [x] `tests/acceptance/corpus/test_fixture_field_consumers.py`: 10 passed.
- [x] Full `tests/acceptance/corpus/`: 502 passed, no failures (count differs slightly from a pre-flip baseline because the 3 declared-blocked cases now carry fewer active invariants — expected, not a regression).
- [x] Full `tests/acceptance/`: 794 passed, 134 skipped (pre-existing, docker-gated), no regressions.
- [x] mypy / ruff clean.
- [x] `provider-scripts/role-legacy_agent.json` untouched (hard constraint — its identity digest is pinned by the in-flight armed run).
- No gate run yet, no merge — acceptance stack is owned by the phase3 armed run; corpus-side edits (declared-blocked flips + `fault_ref` deletion) are included here but merge waits until that run's grade lands, per lane scope. Code-side (guard test, derived cross-check) has no such dependency.


chrisgeo:
## Update: added a 4th instance, kill_switches (found by Phase 4d lane), and rebased

Rebased onto current origin/main (c8014a68a, #1596). Zero `provider-scripts/`/`role-legacy_agent.json` files in the diff, confirmed post-rebase.

**`kill_switches`** — RETIRED, clean deletion (no corpus dependent to flip): world.json carried a per-org `kill_switches` block (global/org/provider/role/surface/contextual_entry, hardcoded `"enabled"` everywhere) with zero production readers. CHAOS-3458 (Done) already ruled the whole granular kill-switch concept superseded by the single `emergency_disabled` mechanism (`org_policy.ASK_DEV_EMERGENCY_DISABLED_KEY`, actually read by `router.py`/`ask_dev.py`) — that ticket's own 6 named `ops.kill-switch.*` corpus ids exist precisely because nothing ever enforced this fixture. The 2 that were actually authored as case files are already declared-blocked for unrelated reasons (CHAOS-3454, CHAOS-3549) and don't reference `kill_switches` by name. Deleted from all 3 world.json org entries. RED-first proven (stashed the world.json edit, confirmed the new test fails, restored, confirmed green).

Head: c887d45f8. Full `tests/acceptance/`: 795 passed, 134 skipped (pre-existing), no regressions. mypy/ruff clean. Gate held per team-lead direction until the phase3 armed run's grade lands — one gate+CI cycle for the full PR (code+corpus together) after that.

PR #1602, opened 2026-08-08
CHAOS-3563 + CHAOS-3566: declared-state history + org-deletion derived-store registry

chrisgeo:
## Summary

- `0c0046657` CHAOS-3566: explicit derived-store registry for org deletion
- `c50cfef68` CHAOS-3563: retain declared-state history for projects (CH migration 074: `project_declared_state_history`)
- `3207b668c` CHAOS-3566: frame the registry guard as a deletion-completeness decision
- `0ae4ee8ae` CHAOS-3563: degrade honestly when as_of predates the retained floor
- `928b5978d` CHAOS-3563/3566: fix PR review findings F1-F10 (ledger below)
- `ca82312b5` CHAOS-3563/3566: fix PR review round-2 findings NEW-1..5, C1-C4 (ledger below)
- `c9ef12701` CHAOS-3563/3566: fix PR review round-3 final-verify blockers A, C1-v3, C, D (ledger below)
- `309871f72` CHAOS-3602: TTL-horizon-aware world generation and mint/restore guards (Task B ledger below)
- `0e7ee60f9` CHAOS-3602: re-mint ask-dev-world.v1 (074/075 shape) with TZ=UTC + TTL-guarded generation

## Details

**CHAOS-3563 — declared-state history (three-outcome floor contract):**
- Normal case: `as_of` within retained history resolves to the correct declared state.
- `as_of` predates the retained floor: explicit warning is surfaced and state resolves to `None` rather than silently returning a wrong/incomplete answer.
- Genuinely-no-history case is distinguished from the predates-floor case rather than collapsed into it.
- The floor fact lives in a separate `project_declared_state_floor` table (round-2 NEW-1), written only by migration 074's backfill — never merge-collapsible the way an in-history column was.
- The RMT tie-break for `project_declared_state_history` is a content-derived hash (round-3 C1-v3), not an id-generator's low bits — see the findings ledger below for why that mattered.

**CHAOS-3566 — org-deletion derived-store registry:**
- Registry guard is framed around deletion-completeness: it exists to catch derived stores that were added but never wired into the org-deletion sweep, not just to enumerate stores.
- The drift warning reports actual per-table outcomes (purged / failed / skipped-no-org_id / not-verified), for BOTH registered and unregistered tables (round-2 C3, round-3 C).

**Migration note:** CH migration 074 (`project_declared_state_history` + `project_declared_state_floor`) and migration 075 (a reconciler for 074, see round-3 finding A) both verified against `origin/main`'s current migration max (073) — no collision, both numbers free.

**Task B (orchestrator-owned, now complete):** the world snapshot has been re-minted for the 074/075 shape — see the Task B ledger below for the full investigation and fix.

## Review findings ledger, round 1 (`928b5978d`)

A high-effort review returned 10 verified findings (8 CONFIRMED, 2 PLAUSIBLE), all fixed RED-first (a failing test observed before each fix, and where practical a clause-deleted/predicate-removed mutant confirmed killed):

| # | Finding | Fix |
|---|---|---|
| F1 | `argMax` skips NULL per-column; a cleared `target_date` could resurrect a stale value from a different row than `state` | Tuple-wrapped every column into one `argMax(tuple(...), tie)` call so all fields come from the same winning row |
| F2 | Only the metrics sink wrote declared-state history; `ClickHouseStore.insert_projects` and the fixtures generator wrote `projects` alone, silently losing declared state from the new read | All three writers now mirror into `project_declared_state_history` |
| F3 | Ambiguous-provider / as-of-inactive rows leaked a watermark even though their state is suppressed | `last_synced` is now nulled in SQL unless `provider_count = 1 AND is_active = 1` |
| F4 | Floor-breach warning fired even when a project was simply created after `as_of` (no real floor breach) | Added a floor discriminator to migration 074; only a genuine floor breach warns now (redesigned in round 2/3, see NEW-1 below) |
| F5 (PLAUSIBLE) | History insert ran before `projects`; a missing history table (stale worker image) aborted the `projects` write too | Write order flipped to `projects` then history in all three writers; history insert stays fail-loud |
| F6 (PLAUSIBLE) | `argMax` tie-break used `ingested_at`, disagreeing with the table's own RMT version column (`last_synced`) — the answer could flip across a background merge | Tie-break changed to `(updated_at, last_synced)` (refined further across rounds 2-3, see C1 below) |
| F7 | org_deletion's registry-drift warning claimed "this run still purged them" before any purge occurred (false on dry-run / unresolved client) | Warning now appended after the outcome is known, worded per outcome (would-purge / purged / not verified; sharpened to per-table in rounds 2-3, see C3/C below) |
| F8 | Successful external-store deletion visits were reported through `result.warnings`, the failure channel | Added a typed `DeletionResult.external` bucket (parallel to `postgres`/`clickhouse`); warnings now reserved for genuine problems |
| F9 | A unit fake computed `in_bound` from params alone, no longer checking the real SQL still contains the `updated_at <= as_of` predicate | Restored the SQL-literal predicate gate; verified it fails when the predicate is mutated out of the real query |
| F10 | No test failed when the `provider_count == 1` or `is_active` clauses were individually deleted | Added clause-level tests, each verified against its own clause-deleted mutant |

13 additional lower-severity findings from the same review are tracked separately by the orchestrator as a documentation follow-up (not part of this PR).

## Review findings ledger, round 2 (`ca82312b5`)

The round-1 verifier re-check plus a codex cross-system review returned 9 more findings (5 verifier NEW-1..5, 4 codex C1-C4), all fixed RED-first:

| # | Finding | Fix |
|---|---|---|
| NEW-1 (HIGH, blocking) | The F4 floor marker lived as a column inside `project_declared_state_history`'s RMT merge domain — an ordinary re-sync of an unchanged project shared the same key with a fresher `last_synced` and floor=0, so the next merge silently discarded the floor marker; the reader's `argMin` also ties nondeterministically | Moved the floor fact to a separate `project_declared_state_floor` table, written ONLY by migration 074's backfill, never by any ordinary sync — a floor row can no longer be merged away. Verified with the exact repro: backfill → normal re-sync of the unchanged project → `OPTIMIZE ... FINAL` → floor breach still detected |
| NEW-2 (HIGH, blocking) | `_dt_to_clickhouse_datetime` converted to UTC then stripped tzinfo; clickhouse-connect reinterprets a naive datetime using the WRITER PROCESS's local timezone, corrupting the stored instant by that offset — this PR is the first reader to compare that column against a caller-supplied `as_of`, making it observable | Fixed to return an aware UTC datetime (never stripped); same bug found and fixed in `ClickHouseStore._normalize_datetime` (used by this PR's own F2 writer, now also has a dedicated unit test — round 3). Verified with a 3-TZ matrix (UTC / America/Los_Angeles / Asia/Tokyo) via `time.tzset()` |
| NEW-3 (MEDIUM) | `test_project_declared_state_history_clickhouse_live.py`'s differential proofs ran in no automated lane — CI excludes `-m clickhouse`, and the gate's own `ch_tests()` only ran the narrow argMax proof | Wired the file into `ci/local_validate.sh`'s `ch_tests()` (named explicitly, not a directory-wide `-m clickhouse` sweep) |
| NEW-4 (LOW) | `test_status_change_clickhouse_live.py` skipped only on `CLICKHOUSE_URI` being unset, hard-failing against the repo's own `.env` (`default` database) instead of skipping | Switched to the `_PROTECTED_DATABASES` guard pattern the sibling live-CH files already use |
| NEW-5 (LOW) | `project_declared_state_history` was missing from the fixtures world's `_CLICKHOUSE_DIGEST_TABLES` | Added it |
| (self-discovered) | `_clickhouse_tables_from_migrations()`'s regex scan used `.search()` (first match only) per semicolon-chunk — migration 074's second `CREATE TABLE` (the new floor table) shares a chunk with the first and was silently never discovered as an org-deletion target | Changed to `.finditer()`; over-discovery is already safely filtered downstream, under-discovery was the real risk |
| C1 (HIGH) — **superseded by round-3 C1-v3, see below** | `updated_at`/`last_synced` are both millisecond-precision; two different states sharing the same millisecond for both had no further tie-break | First fix used a `write_seq`-influenced `version_key`; round 3 replaced the low-bit source entirely after the verifier found it was NOT well-distributed (see C1-v3) |
| C2 (MEDIUM) | A project created exactly at its migration-074 floor instant is indistinguishable from one with real earlier state the floor couldn't recover | Documented as an accepted residual in migration 074's docstring (this is structurally inherent, not fixable without a source of truth the migration doesn't have) and pinned with a boundary test |
| C3 (HIGH) — **extended by round-3 C, see below** | org_deletion's drift warning derived its outcome only from `dry_run`, blanket-applied to every unregistered table | Track and report actual per-table outcomes (purged n / failed / skipped-no-org_id / not-verified) |
| C4 (LOW/MED) | Fixtures `ProjectRecord` had no `state`/`target_date`/`url` fields at all — the fixture world could never seed a declared project state | Added the three fields, wired into both the `projects` insert and the F2 history mirror |

## Review findings ledger, round 3 (`c9ef12701`)

A final-verify pass (team-lead + codex cross-system) returned 4 more findings, all fixed RED-first:

| # | Finding | Fix |
|---|---|---|
| A (HIGH, blocking) | Migration 074 was amended in place multiple times during development; a database that ran an OLDER shape has "074" marked applied forever (the migration runner tracks applied migrations by version NUMBER, not content) and never re-runs it — `status --check` reports 0 pending while the reader 500s on a missing `version_key` column, proven live | Added migration 075 as a RECONCILER: detects a stale shape (missing `version_key` or floor table) and rebuilds both tables from `projects FINAL`; no-op when already current. A live test builds a genuinely old-shape table by hand (never via 074/075's own scripts), confirms it breaks the reader, heals it, confirms the reader works afterward — wired into the gate |
| C1-v3 (HIGH, blocking) | The round-2 C1 fix's low-bit tie-break masked `generateSnowflakeID()`'s low 22 bits, assumed near-uniform. The verifier measured those bits as `machine_id(10) \| per-millisecond-counter(12)` — the counter RESETS every millisecond, so IDs generated ~100ms apart collided outright in 10/20 trials (the shipped docstring's "~1-in-4.19-million" claim was off by ~6 orders of magnitude), and the shipped test asserted full-`write_seq` distinctness, which could never fail for that actual defect | Replaced the low-bit source with a masked `cityHash64` of the declared-content columns: identical content → identical key (harmless collision by construction), differing content → deterministic winner, and — unlike `write_seq` — stable across a full table rebuild (what migration 075 needs). `last_synced` still governs whenever it differs, unaffected. Verified with the verifier's own stress methodology (20 repeated live runs, 0 failures) and a rebuild-stability check (`TRUNCATE` + re-insert reproduces the identical `version_key`) |
| C (MEDIUM) | The round-2 C3 per-table outcome tracking only ever warned for UNREGISTERED tables — a REGISTERED table's failed count/delete produced no warning at all, proven live on `repos` | Any failed/not-verified outcome now warns regardless of registration status; tests for both registered and unregistered failure surfacing |
| D (LOW, pre-existing) | `_CREATE_TABLE_RE` matched prose inside migration 027's own comment ("... CREATE TABLE statement (handles...)"), producing a phantom `statement` registry entry and a spurious skip warning on every production org deletion | Anchored the regex to the start of a line; removed the now-stale `statement` registry entry; regression test reproduces the exact 027 prose shape |

Non-blocking (done anyway): added a dedicated unit test for `ClickHouseStore._normalize_datetime` (the NEW-2 sibling fix, previously proven only indirectly).

## Task B ledger — WORLD_DIGEST re-mint (`309871f72`, `0e7ee60f9`)

Re-minting for the 074/075 shape surfaced a real, independent bug — not a re-mint mechanics problem. Investigated live, mechanism confirmed before fixing (three hypotheses tried and discarded before the real one):

| Hypothesis tried | Verdict | Evidence |
|---|---|---|
| Raw Native-transfer flake (client/server transport) | Ruled out | `_dump_clickhouse` issues one unpaginated `SELECT *`, confirmed by reading it — no ORDER BY/LIMIT/OFFSET to straddle a page boundary |
| ClickHouse background-merge defect (plain MergeTree silently dropping a row) | Ruled out | Reproduced the missing row on a completely fresh ClickHouse container with zero session history — but `system.part_log` showed `merge_reason: TTLDeleteMerge`, not a plain merge |
| **TTL horizon collision** | **Confirmed** | `feature_flag_event` carries `TTL toDateTime(event_ts) + INTERVAL 90 DAY DELETE`. The fixture generator's `random.randint(7, 90)` landed a row's age exactly at that horizon; ClickHouse deletes TTL'd rows at merge time, silently, no error — the row was gone before the snapshot step re-read the table. Confirmed via `system.part_log` (a part shrinking 95→94 rows, `merge_reason: TTLDeleteMerge`) and the 11 surviving sibling rows all being younger than the one that vanished |

Fix, in four parts (`309871f72`):

| # | Issue | Fix |
|---|---|---|
| 1 | No single source of truth for which ClickHouse tables carry a TTL, or how many days | New `fixtures/ttl_registry.py`: parses `TTL <col> + INTERVAL N DAY` directly from the migration source (both syntax forms used across the repo) — a migration's TTL changing can never silently drift out of sync with what generation/the mint guard treat as safe |
| 2 | `generate_feature_flags`'s hardcoded `random.randint(7, 90)` landed exactly on the TTL horizon with zero margin (the actual culprit); 3 other TTL'd generators had no protection at all | All 4 generators now clamp against the registry (`max_generated_age_days`), strictly below the mint guard's own threshold so ordinary elapsed time within one mint run can't itself trip the guard — caught live on the FIRST re-mint attempt after the initial fix (8 rows still "at or past" the threshold) and closed with a dedicated slack constant, not a bigger guess |
| 3 | `world.json`'s `pinned_now` (CHAOS-3392, frozen for WORLD_DIGEST determinism) goes stale every day it isn't refreshed, and that staleness eats directly into the TTL margin — generation is pinned-now-relative, ClickHouse's TTL enforcement is always real-time-relative | `_generate_world` computes the drift once (real time captured BEFORE any frozen-clock context), threads it into every clamp via a `ContextVar` (no generator signature changes, no second clock), and refuses outright — naming the real fix — past a 14-day staleness ceiling rather than silently squeezing the generatable-history window |
| 4 | Nothing stopped a mint from snapshotting horizon-adjacent data, and nothing told an operator a long-lived snapshot had exceeded its safe shelf life | Two new guards in `world_snapshot.py`: `_assert_no_ttl_horizon_rows` refuses to snapshot before any dump runs (not after a content-oracle mismatch already means real data loss); `assert_snapshot_not_expired` refuses to restore a snapshot past its shelf life (recorded in the manifest as `ttl_shelf_life_expiry`), naming "re-mint required" instead of a mysterious future digest mismatch |

Kept from the (ultimately not-the-root-cause, but still a real improvement) transfer-flake line of investigation: `_dump_clickhouse` now decodes and verifies its own payload before writing it to disk (retried up to 3x — a separate `count()` query was observed live to agree with a payload that was actually short, so the decoded payload is what's trusted and recorded), and the ClickHouse dump path gets the same raw-vs-dumped assertion the Postgres path already had, correctly scoped for collapsing engines (FINAL-deduped counts legitimately differ from raw pre-dedup counts, unlike a genuine drift).

**Mint evidence (`0e7ee60f9`):** re-minted on a freshly-booted stack (fresh volumes, zero prior session state). Manifest confirms migration 074 (new shape) and 075 both applied; alembic heads unchanged (`['0066', '0087']`, same as before this PR). WORLD_DIGEST `095684fb9e8781431c0d12f66fdc2e92ea21fab5ac810afeb2cac4f2710ebef2` — generated digest == restored digest (lossless round trip) proven in-mint, then **independently re-verified** after a full stack teardown/reboot via `fixtures world-restore` WITHOUT `--mint-digest` (verify mode, not mint mode): `WORLD_DIGEST ...verified`, 87 ClickHouse + 10 Postgres tables. All 10 corpus contract principals authenticate against the restored world. `scripts/acceptance/mint_ask_dev_world_snapshot.sh` also pins `TZ=UTC` (round-2 NEW-2 follow-up: hashed datetime columns made pre-fix mints TZ-dependent on the operator's own machine).

## Test plan

- [x] `bash ci/local_validate.sh` run plain from the worktree root, FIVE times total across this PR (`0ae4ee8ae`, `928b5978d`, `ca82312b5`, `c9ef12701`, `0e7ee60f9` — each earlier green was treated as void once code changed further): ruff format, ruff check, mypy, ClickHouse scratch-db migration (074 + 075), full unit suite (13371 passed / 0 failed / 245 skipped / 1 xfailed at `0e7ee60f9`), argMax live-exec proof, `test_project_declared_state_history_clickhouse_live.py`'s 13 differential live-ClickHouse proofs, migration 075's 2 live reconciliation proofs — all green, all gate-enforced.
- [x] Every finding across all rounds, including Task B, verified RED (failing test / observed wrong behavior, or a direct pre-fix-content revert for the mint/registry work) before its fix, then GREEN after.
- [x] F1/F3/F4/F5/F6/NEW-1/NEW-2/C1/C2/A/C1-v3 additionally verified against a live scratch ClickHouse with differential proofs (pre/post-`OPTIMIZE ... FINAL` agreement checks, a 3-timezone write/read matrix, exact-repro sequences, an old-shape-to-healed live migration proof).
- [x] F9/F10/the self-discovered migration-scan bug/D additionally verified by planting the exact mutant/regression each guard exists to catch and observing it fail before the fix.
- [x] C1's tie-break was stress-tested with repeated live runs after EVERY design iteration (round 2: 20-30 runs caught a first implementation bug before it shipped; round 3: 20 runs, matching the verifier's own methodology, after replacing the low-bit source entirely) — not a single pass at any point.
- [x] Task B's TTL mechanism was confirmed live before any fix was written (`system.part_log` inspection, a fresh-container reproduction ruling out session-load as a confound), and the fix itself re-verified via a genuinely independent restore (fresh stack, verify mode, no `--mint-digest`) rather than trusting the mint's own self-check alone.


chrisgeo:
## Sub-blocking findings — documented, not fixed (per nitpick policy)

The high-effort review verified 23 findings; 10 blocking ones are in the fix ledger. These 7 lower-severity cleanup findings are recorded here for the record and deliberately NOT fix-committed:

- **src/dev_health_ops/api/dev/native_status_change.py:428** — `_PROJECT_DECLARED_FACTS_SQL` scans `project_declared_state_history` twice (bounded + unbounded CTEs cross-joined) where one scan with conditional aggregates would suffice.
- **src/dev_health_ops/api/services/derived_store_registry.py:51** — `DerivedStoreKind.CLICKHOUSE_TABLE` is dead vocabulary — no `DerivedStore` is ever constructed with it; ClickHouse stores are plain strings.
- **src/dev_health_ops/migrations/clickhouse/074_project_declared_state_history.py:88** — `_table_exists` is the 8th hand-copied instance of the same probe across .py migrations (027, 042, 048, 049, 055, 061, 067, 074), and the copies disagree on fail-open vs fail-closed.
- **tests/api/admin/test_derived_store_registry.py:103** — `test_external_derived_stores_are_typed_as_external` loops over an empty tuple — it asserts nothing and cannot fail until a real external store registers.
- **tests/test_migration_074_project_declared_state_history.py:95** — `test_backfill_is_the_only_recoverable_state_documented_as_such` asserts a docstring contains the word "unrecoverable" — prose assertion with a dead second clause.
- **src/dev_health_ops/api/services/derived_store_registry.py:208** — `registered_clickhouse_tables()` is a trivial one-line wrapper called from exactly one test.
- **src/dev_health_ops/api/dev/native_status_change.py:1796** — declared-state row consumption nests four levels deep; flattening would help the next reader.

If any of these become load-bearing (e.g. the first external store registering makes the empty-loop test meaningful), promote them to a ticket then.

PR #1649, opened 2026-08-10
CHAOS-3667: cohorts require corroboration, not just comparability

chrisgeo:
## Issue and phase
CHAOS-3667 (Phase 2B). Child of CHAOS-3661 — Wave 3.2, Lane D.

## Observable outcome
`discover_cohort` included every candidate sharing ANY comparability basis with another candidate (portfolio, owning team, dependency, initiative parent, or merely measuring the same metric NAME as a peer) with zero check on whether the candidate showed any sign of the pressure the question actually asked about. "Which teams are struggling" answered "every team comparable to another team" — recall-safe but precision-poor, matching the measured defect (3/7, 0/4 in the trial).

## Architecture boundary
A second gate, applied AFTER the existing comparability gate (unchanged), never replacing it — full design writeup and measured before/after table in the Linear comment (CHAOS-3667). Summary:
- Cites `measurement_value`/`measurement_cohort_median` per candidate/metric (same "cite, never compute" discipline `drivers._measurement_candidates` already uses), gated by a 20% relative margin (`OUTLIER_MARGIN_RATIO`, same reasoning as `semantic_retrieval.DEFAULT_MARGIN_RATIO` from CHAOS-3654) and a 2-signal minimum (`MIN_CORROBORATING_METRICS`) — directly implementing the ticket's own instruction: "do not equate one noisy metric with struggle or capacity pressure."
- Three-valued disposition: zero relevant metrics measured → **preserved** (a coverage gap is not evidence of health — the recall half of the ticket's own bound); measured but not corroborated → **excluded** via the frozen contract's pre-existing, previously-unused `CohortExclusionReason.EXCLUDED_BY_QUESTION` (zero producers anywhere in the codebase before this — same shape as CHAOS-3632's `approved_documents` gap); corroborated (≥2 outlying metrics, or dependency concentration) → included, unchanged.
- No frozen-contract change needed anywhere in this PR.

## Measured effect (real `org_helio` corpus, not hand-built fixtures)
`STRUGGLING_TEAMS`/`PRESSURE_SIGNALS` now correctly separate `team_atlas` ("struggling, every axis agrees" — corpus's own comment) and `team_dorado` ("pressure that shows up outward", 2 signals) — both stay included — from `team_borealis` ("high WIP, nothing else corroborates") and `team_frost` ("healthy except one noisy metric") — both now excluded with a stated rationale. `team_ember` ("numbers exist but coverage doesn't support them") is preserved as unknown rather than falsely read as healthy. **Recall preserved, precision improved.**

## What this does NOT do (disclosed, not hidden)
`PROJECT_CAPACITY`/`PORTFOLIO_DEPENDENCY_RISK`/`DECLARED_VERSUS_ACTUAL` show **no measured change**. Traced why: this corpus's project-level demand measurements (`arrived_items` etc.) carry no cohort median at all — capacity pressure there is an *intra-entity* demand-vs-delivery-capacity ratio (the P05/P07 oracle shape I already flagged as CHAOS-3668's driver-model scope in the CHAOS-3634 PR), not an *inter-entity* cohort comparison this mechanism can reach. `PORTFOLIO_DEPENDENCY_RISK`'s empty metric set is a deliberate design boundary (pinned by its own test), not an oversight — that family's own comparability basis (a shared dependency) already is the signal.

## Base/head
Base: `origin/feature/chaos-3498-context-fabric` @ `e02d7e00b` (current tip; rebased clean past CHAOS-3650/3634/3653/3633/3652/3631/3675, no conflicts). Head: `chaos-3667-cohort-ranking`.

## Migrations / configuration / feature flags
None.

## Security / privacy / retention impact
None. Authorization filtering (applied before comparability, unchanged) still runs first; this PR only narrows which already-authorized, already-comparable candidates become members.

## RED-first evidence
`git stash` of the source change with the test file present reproduced `ImportError`, then (with imports stubbed manually during development) pre-fix member-set assertions failing against the real corpus. GREEN after.

## Verification
- New `tests/context_fabric/test_chaos_3667_cohort_ranking.py` (12 tests, differential before/after against the real corpus)
- `.venv/bin/python -m pytest tests/context_fabric/ -q` → 1430 passed, 44 skipped
- `.venv/bin/python -m pytest tests/api/dev/ -q` → 2994 passed, 74 skipped, 4 xfailed
- `.venv/bin/python -m pytest tests/context_fabric/test_chaos_3617_cohort.py tests/api/dev/test_chaos_3534_cohort_scope_outcome.py -q` → 48 passed (closest-neighbor cohort suites)
- ruff / mypy (module + full-project lefthook gate) → clean

## Known limitations and follow-up issues
- `PROJECT_CAPACITY`/`PORTFOLIO_DEPENDENCY_RISK`/`DECLARED_VERSUS_ACTUAL` precision is unchanged by this PR (see above) — closing that needs either canonical services publishing project-level cohort medians, or the CHAOS-3668 driver-model's intra-entity ratio capability, neither of which this ticket's scope covers.
- `FAMILY_PRESSURE_METRICS` is this arm's own reading of the metric vocabulary and may need revisiting as new canonical metrics are added — deliberately independent of `drivers.MEASUREMENT_CATEGORY` for the same reason `METRIC_COMPARISON_DIMENSIONS` already is (see that constant's own docstring).

## Rollout / rollback impact
Purely additive at the cohort-membership decision layer. No flag: the prior behavior (comparability-only inclusion) had no formal "correct" baseline to preserve behind one — it was the measured defect.

### DIFF
--- a/src/dev_health_ops/context_fabric/graph_arm/cohort_discovery.py
+++ b/src/dev_health_ops/context_fabric/graph_arm/cohort_discovery.py
@@ -310,29 +310,39 @@ MIN_CONCENTRATED_DEPENDENTS = 2
 #: CHAOS-3667: how far a cited value must clear its cohort median, as a
 #: fraction of the median, to count as outlying at all.
 #:
-#: A bare ``value > median`` -- the comparison ``drivers._is_outlying``
-#: already uses for a DRIVER, where a human reads one number beside another
-#: -- treats a value one unit over its median the same as one an order of
-#: magnitude over it, which is exactly right for "cite it and let the reader
-#: judge" but wrong for "let this alone put a whole entity into a ranked
-#: answer": a margin this small is measurement noise around the median, not
-#: a signal, and 20% is a defensible floor for "worth citing as elevated"
-#: without being tuned to any specific metric's units or any specific
-#: entity's numbers -- the same relative-margin reasoning
-#: ``semantic_retrieval.DEFAULT_MARGIN_RATIO`` uses for a different
-#: comparison entirely.
+#: **Derivation, stated so it can be checked rather than trusted.** A bare
+#: ``value > median`` -- the comparison ``drivers._is_outlying`` already
+#: uses for a DRIVER, where a human reads one number beside another -- is
+#: right for "cite it and let the reader judge" and wrong for "let this
+#: alone put a whole entity into a ranked answer": a margin this small is
+#: measurement noise around the median, not a signal. 20% is a round,
+#: order-of-magnitude relative floor for "materially outside the norm" --
+#: the same style and reasoning as ``semantic_retrieval.
+#: DEFAULT_MARGIN_RATIO`` (CHAOS-3654), chosen the same way: from the shape
+#: a real signal has to have, not from sweeping or fitting any dataset.
+#: **It was chosen before, and is unchanged by, what any entity in
+#: ``investigation_corpus/world.py`` happens to measure** -- proven in
+#: ``test_chaos_3667_cohort_ranking.py::TestTheGateGeneralizesToHeldOutData``,
+#: which exercises this exact constant against synthetic entity ids and
+#: values that appear nowhere in the corpus, including the boundary itself
+#: (a value exactly 20% over its median does not count; strict inequality).
 OUTLIER_MARGIN_RATIO = 0.20
 
 #: CHAOS-3667: how many independent outlying metrics a candidate needs
 #: before a measurement corroborates its inclusion.
 #:
-#: One is not enough, by the ticket's own instruction: "do not equate one
-#: noisy metric with struggle or capacity pressure." A single elevated
-#: number is a correlate a canonical service happened to report; two
-#: independent ones sitting outside their own cohort's norm, at the same
-#: time, are a pattern. Dependency concentration (:func:`_dependency_
-#: concentration`) is a relational signal, never a metric, and always
-#: satisfies this on its own -- see :func:`_corroboration`.
+#: **Derivation.** One is not enough, by the ticket's own instruction: "do
+#: not equate one noisy metric with struggle or capacity pressure." Two is
+#: not a fitted value -- it is the smallest integer that can express that
+#: instruction at all: any value <= 1 collapses back to "one noisy metric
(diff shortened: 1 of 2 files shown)


chrisgeo:
Pushed the held-out-controls commit per the merge condition (afabe7ff9).

**Derivation, independent of org_helio's cases (now stated explicitly in the code comments, not just here):**
- `MIN_CORROBORATING_METRICS = 2` is not fit data — it's the smallest integer that can express the ticket's own instruction ("do not equate one noisy metric with struggle or capacity pressure") at all. Any value ≤ 1 collapses back to exactly the rule the ticket forbids.
- `OUTLIER_MARGIN_RATIO = 0.20` is a round, order-of-magnitude relative floor for "materially outside the norm" — same style and reasoning as `semantic_retrieval.DEFAULT_MARGIN_RATIO` (CHAOS-3654), chosen from the shape a real signal has to have, not swept or fit against any dataset.

**New held-out synthetic controls** (`TestTheGateGeneralizesToHeldOutData`, 6 tests): calls `_corroboration`/`_PressureReading` directly against entity ids and values (`entity_zeta_never_in_any_corpus`, `entity_omega_never_in_any_corpus`) that share no number with anything in `investigation_corpus/world.py`:
- two independent >20%-margin signals correctly corroborate (should-pass)
- one wide-margin signal alone does not, at a metric/margin combination distinct from both `team_borealis`'s and `team_frost`'s real numbers (should-fail)
- the margin boundary itself: exactly +20% does not count, strictly over it does (strict inequality, not an off-by-one that happens to work on real data)
- a metric outside the family's declared vocabulary is never checked, however extreme its value (family-scoping is real)
- dependency concentration alone corroborates with zero readings at all (the relational path is independent of measurement entirely)

Only the metric *names* (incidents, cycle_time_median_days, ...) are shared with the corpus test file — `FAMILY_PRESSURE_METRICS` is this module's own declared axis vocabulary, not a corpus fact.

Verification: `pytest tests/context_fabric/ -q` → 1436 passed, 44 skipped; `pytest tests/api/dev/ -q` → 2994 passed, 74 skipped, 4 xfailed; ruff/mypy clean.

PR #1696, opened 2026-08-10
CHAOS-3730: mirror GitLab files extension case sensitivity

chrisgeo:
## Scope

Correct GitLab files extension matching to preserve Python's case-sensitive behavior. Mixed-case paths such as `src/Main.GO` are no longer fetched when the Python scanner excludes them.

## Evidence

- Baseline divergence reproduced against the real Python producer
- Typed `requested_content_paths` traversal oracle
- Shared mutation harness: 36/36 killed; the new clause is killed by the live oracle
- Full providersync, vet, Ruff, and mypy green

Linear: CHAOS-3730

SCREENSHOT-WAIVER: Backend-only provider behavior; no rendered UI changes.


chrisgeo:
Closing as superseded. Every commit in this PR is patch-equivalent to feat/go-worker-runtime-migration; the case-sensitivity fix is present there as f292228fa. The PR is stale/dirty only because its stacked base moved.

PR #1779, opened 2026-08-18
Remove non-gating review scaffolding and collapse route-switch duplication (CHAOS-3875)

chrisgeo:
## Summary

Two separable things, in that order in the history:

**A. CHAOS-3875 cleanup** (commits 1–3): removed non-gating review scaffolding and collapsed the route-switch duplication. +308 / −45,275.
  - **Mutation harness (~24k lines).** `scripts/mutation_harness*.py` (5 files), `tests/tooling/test_mutation_harness*.py` (5 files), `internal/providersync/mutation_plan_integrity_test.go`, `tests/tooling/README.md`, and **98** plan JSONs across four `mutation-plans/` trees. `ci/local_validate.sh`'s `mutation_clean` stage — the only thing anywhere that invoked the harness — is gone with it, and the declared stage set drops 9→8 (5→4 under `SKIP_CLICKHOUSE=1`).
  - **`internal/domaingrants` + `cmd/dev-health-grantcheck` (~8k lines)** and the `grant-advisory` verb, plus `internal/storage/river/grantcheck_export.go`, which existed only to feed it.
  - **`cmd/dev-health-sync-parity` + `scripts/worker/observe_sync_dispatch_parity.py`**, with the `sync-parity` Dockerfile stage, the parity binary's build step, and the contract test that pinned their layout.
  - **`DEV_HEALTH_QUEUES`**: queue topology is flag-only now, and the conflict-detection branch is gone.
  - **Profile plumbing**: resolution moved out of `internal/platform/config` into `internal/platform/shell`, where the flag is already parsed.
  - **Route switches**: `tests/test_compose_config.py`'s hand-maintained census is now derived from `contracts/provider-matrix/v1/matrix.json`.

**B. Three test fixes found while validating A** (commits 4–6). These are unrelated to the cleanup and are **easy to split into their own PR — say the word and I will.** They are here only because chasing "why is the gate red" turned them up. Two are real product bugs.

### Scope notes — please read before approving

**1. The mutation-harness removal is wider than the issue's "39 plan JSONs."** That count covers `tests/tooling/mutation-plans` only. Three more trees existed — `internal/providersync/testdata/mutation-plans` (57), `internal/platform/config/testdata/mutation-plans` (1), `internal/providerfamilycontract/testdata/mutation-plans` (1) — for 98 total. All go for the same reason: `ci/local_validate.sh` only ever ran `mutation_harness.py verify`, so the plans themselves were never executed by any automation, and `mutation_plan_integrity_test.go` existed solely to keep them honest.

**2. `domaingrants` — delete vs. gate. This PR takes the default (delete) and asks you to confirm.** CHAOS-3875 says "either make it gate or delete it." Gating it would mean accepting its current findings as the baseline, and its analysis is regex over Go source with a package doc that points at a path on a developer's laptop. Say the word and I will restore it behind a gating verb instead.

**3. Item 3 (evidence/meta-test weight, ~4.5k lines) is NOT done — deliberately. Every part of it is load-bearing.** Evidence, not assumption:
   - `ci/evidence/go-worker-migration/` is read by `ci/check_river_compat_static.sh` (lines 10, 49–51), which `go.yml` runs. The issue's premise — "go.yml's path filter for it is already commented out" — is about the *path filter*, not the *consumer*: the script `jq`-asserts specific keys in `v0-celery-baseline/capture.json`, `v0-celery-baseline/local-resource-snapshot.json`, `v1-river-spike/compatibility-matrix.json`, and `v1-river-spike/local-harness-results.json`. Deleting the tree breaks that check.
   - `scripts/user_guide_evidence_contract.py` and `scripts/user_guide_evidence_validation.py` are not "referenced only from a checked-in PR-body template" — they are libraries imported by `scripts/validate_user_guide_evidence.py` and by three test modules under `tests/docs/`.
   - `tests/tooling/test_local_validate_heredocs.py` (399 lines) guards a real pipe-wedge failure class across *every* `ci/*.sh`, not just `local_validate.sh`; `test_local_validate_lock.py` (1063 lines) guards a real single-flight lock with adversarial-review history behind it.

   I did remove the one genuinely dead piece in that area: `tests/tooling/README.md`, which was entirely the mutation-harness runbook.

**4. `ci/check_go_containers.sh` was still using `DEV_HEALTH_QUEUES`.** The issue's "all deploy artifacts use flags since `1001ded9`" is true of deploy artifacts, but the container smoke check was a live consumer. It now appends `--queues=sync,sync_provider` after the image, which the `worker` target accepts directly (ENTRYPOINT, no CMD). `DEV_HEALTH_QUEUE_CONCURRENCY` and `DEV_HEALTH_WORKER_GROUP` keep their env paths — the issue scoped the removal to `DEV_HEALTH_QUEUES`.

**5. Route switches are derived and enforced, not generated.** The four copies (Go typed config table, `compose.yml`, `.env.example`, the Go-workers overlay) were already bound together by `_PROVIDER_ROUTE_SWITCH_NAMES` in `tests/test_compose_config.py` — a hand-maintained frozenset that was itself a fifth copy. Deriving *that* from `matrix.json` closes the loop with one change and no codegen, and preserves the hand-written per-field documentation on `config.Config` that a generator would have flattened. Adding a pair to `matrix.json` now fails three tests by name until every copy carries the switch.

### The three fixes in part B

**`fix(jira)` — a real bug, and my own regression.** Jira Cloud returns the reporter's local offset, not always `+0000`. Pydantic preserved whichever it parsed, so `operational_ordering_codec.canonical_datetime` — which requires a zero-offset datetime — raised `OperationalOrderingEncodingError` and aborted conflict-key construction for the **entire batch** on a single `+0200` incident. The Go route already normalized (`parseJiraIncidentTime` → `parsed.UTC().Truncate(microsecond)`), so the two runtimes also emitted different rows for one payload. This came in with the CHAOS-3869 fix, which added the `jira_cloud_non_utc_offset` oracle case and fixed the Go side without matching Python. Naive timestamps are now rejected rather than assumed UTC, matching Go, whose layouts all require an explicit offset.

**`fix(synccoverage)` — a real bug, pre-existing.** `canonicalBackfillWindows` diverged structurally from the authoritative `_canonical_backfill_windows`; both feed the same focused-backfill selector. Two defects: (a) the selector only accepts UTC-midnight boundaries and Python *skips* anything else, but Go *coerced* partial days onto day boundaries — manufacturing windows the API would refuse, including a degenerate `since == before` window covering nothing; (b) a window is only actionable against its own source and dataset, but Go iterated datasets alone, dropped source identity, and merged adjacent windows across both, so one source's gap could widen another source's window. `TestPayloadBackfillScheduleAndStatusSemantics` asserted the degenerate window, so it encoded the defect; it now asserts the skip (with the gap still counted in `overall`), and a new test covers the positive path.

**`test(helm)` — an over-pinned assertion, no product change.** `test_local_post_install_requires_bundled_postgres` asserted helm's schema-error path as a JSON Pointer (`/postgresql/enabled`); current helm renders it dotted (`postgresql.enabled`), and the repo pins no helm version anywhere. Both renderings are now accepted; the chart must still refuse the combination and name the field, matching how the sibling test above already asserts.

**Nothing was activated.** No `WORKER_*_ENABLED` switch flipped, migration 0066 untouched, `deployment_state` and replica defaults unchanged, PR #1738 not merged. `internal/jobs/metrics/*/compatibility_http.go` was not touched.

## Checklist

- [x] I added/updated tests, or provided required governance evidence below.
- [x] I ran relevant local validation (minimum: `./ci/run_tests.sh unit` for `src/` changes).
- [x] I documented risk and rollback considerations.
- [x] I called out breaking changes, migrations, or config impacts (or wrote `none`).

## Test Evidence

TEST-EVIDENCE:

**`bash ci/check_go.sh fast` — green end to end.**

```
$ PATH="$PWD/.venv/bin:$PATH" bash ci/check_go.sh fast; echo "exit=$?"
exit=0
```

That includes the full `live-python-oracles` verb, now clean across every package:

```
ok  github.com/full-chaos/dev-health-ops/internal/providersync        102.031s
ok  github.com/full-chaos/dev-health-ops/internal/providerfoundation    1.568s
ok  github.com/full-chaos/dev-health-ops/internal/scheduler/sync        1.538s
ok  github.com/full-chaos/dev-health-ops/internal/jobs/metrics/daily    2.731s
ok  github.com/full-chaos/dev-health-ops/internal/synccoverage          1.463s
```

**`bash ci/check_go.sh multi-replica-workers`:**

```
ok  github.com/full-chaos/dev-health-ops/cmd/dev-health-worker  10.120s
multi-replica worker gate: measured 4 terminal jobs
```

**`bash ci/local_validate.sh`** — the stage manifest confirms `mutation_clean` is gone and every remaining stage is declared and executed:

```
   ✔  lint: ruff format --check          2370 files already formatted
   ✔  lint: ruff check                   All checks passed!
   ✔  typecheck: mypy                    Success: no issues found in 2403 source files
   ✔  clickhouse: docker probe (container 'dev-health-clickhouse-1' reachable, dev-hops present)
   ✔  ch-scratch-create (ci_local_validate_440b4cdd6d46)
   ✔  ch-migrate (upgrade + status --check)
   ✗  unit suite (FULL, not subset)      3 failed, 13079 passed, 645 skipped, 1 xfailed
GATE_STAGE_MANIFEST result=FAILED declared=8 executed=7
  declared_ids=lint_format,lint_check,typecheck,ch_probe,ch_scratch_create,ch_migrate,unit_suite,ch_argmax_proof
```

**The 3 remaining failures are a container artifact, diagnosed rather than assumed.** They are the process-group reaping tests (`test_worker_metrics_process.py` ×2, `test_worker_workgraph.py` ×1). `os.killpg` *does* kill the grandchild — but this container's PID 1 is `process_api`, which does not reap orphans, so the dead process sits as an unreaped zombie and the tests' `os.kill(pid, 0)` liveness probe keeps succeeding. Proof:

```
parent 23851: gone
child  23852: state=Z ppid=1     <- Z = killed, reparented to PID 1, never reaped
PID 1 is: process_api
```

The production code (`_terminate_compatibility_process`: `start_new_session=True` + `os.killpg`) and the tests are both correct; a normal reaping init makes them pass. **Deliberately not "fixed"** — teaching `_assert_process_reaped` to accept a zombie would blunt a real leak assertion to suit one sandbox.

**Focused suites** (all green):

```
tests/tooling/ + tests/workers/test_go_worker_deployment_contract.py   204 passed, 7 skipped
tests/test_compose_config.py                                            46 passed, 7 skipped
tests/workers/ + tests/test_sync_units.py + tests/_env_isolation.py    746 passed, 9 skipped
tests/test_helm_migration_hook.py + test_helm_api_probes.py
  + tests/workers/test_helm_go_pgbouncer.py                              9 passed
tests/providers/jira/test_jsm_incident_normalize.py                     11 passed
go test ./internal/platform/... ./cmd/...                              all ok
```

`test_helm_go_pgbouncer.py` now actually executes rather than skipping, which also validates PR #1777's pooler-oracle fix against real `helm template` output for the first time.

**The Jira regression tests fail without the fix.** Reverting only `jsm_models.py`:

```
FAILED tests/providers/jira/test_jsm_incident_normalize.py::test_normalize_pins_non_utc_jira_offsets_to_utc
FAILED tests/providers/jira/test_jsm_incident_normalize.py::test_normalize_rejects_a_naive_jira_timestamp
2 failed, 9 passed
```

**Drift detection proven by injection, not by inspection.** Injecting a synthetic `(github, packages)` pair into `matrix.json`:

```
FAILED tests/test_compose_config.py::test_provider_route_switch_inventory_matches_go_config
FAILED tests/test_compose_config.py::test_route_switches_default_off_for_producer_gate
FAILED tests/test_compose_config.py::test_provider_route_env_example_is_unset_and_default_off
3 failed, 43 passed, 7 skipped
```

All three pass again once the pair is removed, and `git diff` on `matrix.json` is empty in the committed tree. The derived set is the same 40 names the frozenset listed, so this adds enforcement without changing the census.

**Counts that moved, and why:**
- `tests/tooling/test_go_integration_sharding.py`: pinned `internal/providersync` top-level test count 922 → **921**, because `mutation_plan_integrity_test.go` contributed exactly one.
- `tests/tooling/test_local_validate_stage_manifest.py`: `declared=9 executed=9` → **8/8**, and `5/5` → **4/4** under `SKIP_CLICKHOUSE=1`, matching the removed `mutation_clean` stage.
- `tests/tooling/test_check_go_public_verbs.py`: dropped `grant-advisory` from `PUBLIC_VERBS`. Its sibling assertion `source.count("    check_multi_replica_workers\n") == 3` is unchanged and still passes.

## Risk Notes

RISK-NOTES:

- **Blast radius — the cleanup.** Two behavioural changes reach a running process, both narrow:
  1. A worker configured *only* via `DEV_HEALTH_QUEUES` now fails at startup with "at least one worker queue is required" instead of booting. No deploy artifact is in that state — Compose, Swarm, raw Kubernetes, and the Helm chart all pass `--queues`. The one consumer that was, `ci/check_go_containers.sh`, is converted here.
  2. Profile resolution moved packages. Order is unchanged and now pinned by a table in `shell_test.go` covering flag-beats-env, env-beats-default, blank-env-falls-through, both undeclared-value rejections, and both profile-free-service cases. `config_test.go` gained the complementary assertion that `Load` carries the value verbatim and never reads `DEV_HEALTH_PROFILE` itself. Only `dev-health-stream-runner` declares profiles.
- **Blast radius — the two product fixes.** Both change emitted data, so both are worth a reviewer's eye even though both move Go/Python toward each other rather than apart. The Jira fix shifts stored `source_version_at`/`source_event_at` for non-UTC-offset incidents to the correct instant (same moment, UTC-expressed) and makes previously-crashing batches succeed. The synccoverage fix changes `backfill_windows` shape: full RFC3339 stamps instead of date-only, `source_ids`/`dataset_keys` now present, and fewer windows (ones the selector would have refused are no longer offered). Any consumer reading the old date-only form should be checked — the Python API has always emitted the new shape, so a correct consumer already handles it.
- **Everything else is deletion.** No production code path lost a caller: `internal/syncdispatchcontract` and `internal/syncreconciler` stay (imported by the reconciler, workerctl, jobrescue, syncroute, and syncdispatchruntime), and `contracts/sync-dispatch/v1` is still shipped into the reconciler and operator runtime images.
- **Rollback.** `git revert` of any commit restores its files; nothing here is stateful, and no migration, schema, or switch default changed. The frozen parity evidence records its `source_revision` (`a1d86de`) in `capture.json`, so the removed oracle is recoverable from history if the comparison is ever needed again — its README now says so instead of printing a `go run` line for a binary that no longer exists.
- **What you lose.** The ability to run a checked-in mutation plan with one command, and the grant-surface report in the CI log. `AGENTS.md` and `deploy/go-workers/provider-sync-porting-recipe.md` keep the durable lessons (mutate compound predicates clause by clause; never verify a restore with a build or a git check; a `SURVIVED` result is not automatically a coverage gap) as hand-run discipline. Records of past passes in `contracts/provider-matrix/v1/README.md` are reworded to past tense and say plainly that the plan is recoverable only from history.
- **Monitoring.** None required; nothing activates.
- **Follow-up worth filing.** `.github/workflows/go.yml`'s `push`/`pull_request` triggers are commented out. That is how the stale multi-replica gate name, a red `internal/storage/river` integration test, and the Helm pooler-oracle drift all reached this review unnoticed — and it means this PR's Go verbs will not run on CI either. It is also why both product bugs above sat undetected: the Jira one is only visible through the `live-python-oracles` verb. Independent of this lane, but it is the reason "no GitHub workflow runs it" was true of so much of what CHAOS-3875 lists.
- **Correction to an earlier version of this description.** It claimed ~145 live-Python-oracle failures attributable to a CPython 3.14-vs-3.13.14 mismatch. That was wrong on both counts: I had omitted `DEV_HEALTH_LIVE_PYTHON_ORACLE_PROOF_DIR`, so most of those "failures" were the harness's own opt-in guard firing, and `pyproject.toml` sets `requires-python = ">=3.14"` with CI on 3.14 — 3.13.14 is the River compatibility harness's pin, a different thing. There was exactly **one** real oracle failure (the Jira one), now fixed.

## Optional Context

- **Linked issues:** CHAOS-3875 — https://linear.app/fullchaos/issue/CHAOS-3875/remove-non-gating-review-scaffolding-and-collapse-route-switch
- **Base branch:** `feat/go-worker-runtime-migration` (PR #1738's head), not `main`.
- SCREENSHOT-WAIVER: backend-only change — tooling removal, Go/Python configuration plumbing, provider normalization, and CI scripts. No user-facing surface is affected.


chrisgeo:
**Open question resolved: `domaingrants` delete is confirmed** by the repo owner. No change needed — this PR already deletes `internal/domaingrants`, `cmd/dev-health-grantcheck`, `internal/storage/river/grantcheck_export.go`, and the `grant-advisory` verb. Scope note 2 in the description is now settled, not pending.

The other two notes still stand as written: **item 3 stays rejected** on the evidence in note 3, and **the two product fixes in part B are separable** into their own PR on request.

---

### One thing this PR does *not* carry, and why

The description's follow-up note says `.github/workflows/go.yml`'s triggers are commented out. The owner confirmed that was deliberate — constant runs plus fast commits were exhausting the org's Actions concurrency — and okayed re-enabling them.

The history refines that story: the triggers were never *disabled*. `d814ca0` **created** `go.yml` with `push` and `pull_request` both commented from the start, leaving `merge_group` and `workflow_dispatch` as the only automatic triggers. `merge_group` fires only for the queue into main, so a long-lived feature branch like `feat/go-worker-runtime-migration` has had **zero** Go CI for its entire life — which is exactly why the stale multi-replica gate name, the red `internal/storage/river` integration test, and the Helm pooler-oracle drift all reached this review by hand rather than by CI.

I prepared the change and then **could not push it**: this session's credentials lack the GitHub `workflow` scope. I did not route around that through the API — modifying CI workflow definitions is precisely what that scope gates, and an authorization to make a change is not an authorization to bypass the mechanism restricting it. The patch went to the owner to apply directly.

For the record, what it does:

- **Re-enables `pull_request` only.** PR runs already carry `cancel-in-progress: true`, bounding the workflow to one live fan-out per PR (8 declared jobs, ~13 once the integration shard matrix expands, several at a 25-minute cap). That bound is what makes this trigger affordable under the concurrency limit that caused the original problem.
- **Leaves `push` commented, deliberately.** `push` is per-commit *and* intentionally not cancellable — `cancel-in-progress` is false for non-PR events so each main commit gets a real verdict rather than "cancelled" — so a burst of merges holds several full fan-outs at once. That is the shape that exhausted concurrency. `merge_group` still gates the merge itself, so leaving `push` off costs post-merge re-runs on main, not gate coverage.
- Corrects the `concurrency` comment, which described `on.push` as live when it never has been.

Until that patch lands, **this PR's own Go verbs will not run on CI** — the `ci/check_go.sh fast` and `multi-replica-workers` results in the description are local runs.

---
_Generated by [Claude Code](https://claude.ai/code)_

chrisgeo:
## First CI run: 15/18 green. All 3 failures predate this PR — reproduced at `1001ded9`.

The `pull_request` trigger is live, so this workflow has now run against a PR for the first time. **`go-quality` passed**, which is the job that actually covers this PR's changes — it runs `ci/check_go.sh all` (fmt, vet, test, live-python-oracles, build, contract, integration-vet, shard plans, multi-replica gate). So the cleanup itself is validated by CI, not just locally.

The three red jobs are all pre-existing breakage on `feat/go-worker-runtime-migration`, surfacing for the first time because the branch has never had CI. I verified each by rebuilding at **`1001ded9`** — PR #1738's own head, before any of the six lane branches — in a clean worktree.

### 1. `go-container-smoke` — scheduler exits instead of staying live

```
container scheduler exited before publishing :8080; its output was:
{"level":"ERROR","msg":"configure runtime dependencies","error_category":"dependency_configuration_failed"}
```

The smoke contract is that a foundation binary stays live and fails *readiness* until its dependencies are configured. The worker does exactly that; the scheduler crashes instead. Side by side locally, with no database configured:

| binary | `/healthz` | `/readyz` | process |
|---|---|---|---|
| `dev-health-worker` | 200 | 503 | stays live ✅ |
| `dev-health-scheduler` | — | — | exits, `dependency_configuration_failed` ❌ |

The worker behaves correctly because CHAOS-3864 (Lane A, PR #1774) added `databaseConfigurationRejected` to separate a *declared configuration rejection* (stay live, report unready) from an *operational* failure (crash-loop). The scheduler's `openSchedulerDatabase` → `postgres.NewRuntimePools` failure never got that treatment, so it propagates as a fatal dependency error. `reconciler` and `stream-runner` are the other two `RUNTIME_TARGETS` and are likely the same shape.

Reproduces identically at `1001ded9`. My only change to `cmd/dev-health-scheduler/` across all six lanes is a corrected comment (Lane E).

### 2. `go-storage-integration-shard-packages-3` — 12 grant/authorization failures

All in `internal/storage/postgres`, all `PostgreSQL readiness check failed`, spanning `domain_grant_reconciliation_integration_test.go`, `coordinator_statement_privileges_integration_test.go`, `domain_authorization_integration_test.go`, and `runtime_authorization_integration_test.go`.

These are grant-posture tests and this PR deletes `internal/domaingrants`, so that connection is worth ruling out explicitly — it does not hold. `git diff 69562a9..HEAD -- internal/storage/postgres/` is **two comment edits and nothing else**, and the deleted package had no importers (`go build ./...` and `go vet ./...` are clean). Confirmed by running the tests directly:

- on this branch → fail
- on base `69562a9` → fail identically
- at `1001ded9` → fail identically

### 3. `river-compatibility` — session crash-rescue probe

```
river compatibility harness: session-crash-rescue failed
river compatibility harness: exited during session_crash_rescue_worker with status 1
```

`tests/compatibility/river/` is untouched by all six lanes (`git diff --stat 1001ded9..HEAD` reports no files there).

---

### What I propose

These are three genuine defects in PR #1738's own scope, not CHAOS-3875's, and #1 in particular looks like a real production behaviour bug: a scheduler that crash-loops on an unconfigured dependency instead of reporting unready is exactly what the stay-live contract exists to prevent, and it would fail a Kubernetes rollout differently than intended.

I have **not** folded fixes into this PR. Adding process-boot changes for three more binaries, plus a grant-posture investigation, on top of a 45k-line deletion would make it unreviewable and mix two unrelated concerns.

The usual "the base is red, wait for it to recover" resolution does not apply here — this branch has never had CI, so nothing will fix it on its own. Happy to take all three on as a separate PR against the same base; say the word and I'll start with the scheduler one, which looks smallest and highest-value.

---
_Generated by [Claude Code](https://claude.ai/code)_

chrisgeo:
## Go workflow is fully green. One Python failure left, and it was mine — fixed in `7d3fea1`.

### The Go side is done

Every job in the Go workflow passed on `566f8b5`, including the three that were red:

| job | before | now |
|---|---|---|
| `go-quality` | ❌ | ✅ |
| `go-container-smoke` | ❌ | ✅ |
| `river-compatibility` | ❌ | ✅ |
| `go-storage-integration` + all 6 shards | ❌ | ✅ |
| container-security ×7, reproducibility | ✅ | ✅ |

**Correction to my earlier comment on this PR.** I reported "15 of 18 jobs passed, including `go-quality`". That was wrong — `go-quality` was red in that first run too, on the same `ModuleNotFoundError: No module named 'clickhouse_connect'` it had until `0f6f01c`. The `failed_only` log query returned 3 of the 5 failing jobs and I treated it as complete, so the claim that this PR was CI-validated rested on a bad reading. It is validated now; it wasn't then.

Two fixes closed the Go side:

- **`0f6f01c`** — the daily-metrics oracle reaches `job_daily` → benchmarking → metrics sinks → ClickHouse sink → connectors → Celery, a chain no other oracle package touches, so 13 packages were missing from `ci/requirements-live-python-oracles.txt`. That file is installed `--no-deps` precisely so this fails loudly. It passed locally because the dev venv has the full extras; I resolved it against a throwaway CPython 3.13.14 venv holding only that file.
- **`566f8b5`** — `TestMultiReplicaFleetSurvivesDatabaseOutage` waited 30s for a completion River's own completer documents as taking up to ~37s when attempts time out. Below the ceiling of the mechanism the test exists to exercise. Not a timeout bump: locally both jobs complete in 0.3s, which doesn't bound the CI case at all, because whether the outage window covers the sibling job's completion is timing-dependent.

### The remaining `test` failure was my regression

The Tests workflow was **green** on `f27c821`, `ee50616`, and `093c93bc`, and went red exactly at the rebase — which is when the four CI fixes landed. So it was mine, not pre-existing. Two tests:

```
test_local_postgres_bootstraps_distinct_go_runtime_roles
test_least_privilege_domain_grants_reject_swapped_privilege_blocks[script_path1]
```

Both fail on an exact-set assertion over which tables `scripts/worker/provision_river_roles.sql` grants DELETE, and the comment above it states the convention I broke:

> The domain ROLE holds DELETE on more tables through `domainPosture()`; they are simply not granted by these scripts. Stating this unscoped would be false.

The bootstrap script grants a deliberate **subset** of the posture — the pinned River migration owns the rest, through `runtimeGrantStatements` in `internal/storage/river/migrate.go`, which already grants both chunked-persistence tables correctly and is what actually runs. I had justified the addition as keeping two parallel lists in agreement; that premise was simply wrong, and this test exists to stop the bootstrap script drifting toward the full posture. `7d3fea1` removes it.

Nothing is lost: the half of that commit that actually fixed the twelve `internal/storage/postgres` failures — creating and granting the two tables in the integration fixtures that were missing them — is untouched.

**One thing for whoever owns the base branch:** `feat/go-worker-runtime-migration` carries those same four lines, merged from the CI-fix branch, so **its own Tests workflow is red for this reason** until this PR lands. Nothing to do on the base separately — merging this fixes it — but worth knowing if the base looks red in the meantime.

Local verification before pushing: `tests/test_compose_config.py`, `tests/providers/jira/`, `tests/workers/`, `tests/test_sync_units.py`, and `tests/docs/` — 1047 passed, 8 skipped, 0 failed.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1801, opened 2026-08-19
fix(scheduler): declare the scheduled-report tables in the coordinator posture (CHAOS-3935)

chrisgeo:
## Production impact

Scheduled reports have not run since the Go scheduler was activated. Every
`scheduled_reports_dispatch` cycle fails:

```
permission denied for table saved_reports (SQLSTATE 42501)
```

## Cause

`coordinatorPosture()` omits the three tables `internal/scheduler/fixed/reports.go` uses on the
coordinator pool. The posture is the single declaration behind both the derived GRANTs
(`coordinatorGrantStatements` — *"the table list is never written here"*) and the readiness
assertion, so with the tables absent **both sides agreed they were not needed**: readiness passed,
the process started, and every cycle failed at runtime.

| Table | Added | Why |
|---|---|---|
| `saved_reports` | SELECT + **UPDATE** | `dueScheduledReportsSQL` ends `FOR UPDATE OF job, report SKIP LOCKED`; `report` is the `saved_reports` alias, and a row lock requires UPDATE |
| `scheduled_report_occurrences` | SELECT + INSERT + UPDATE | insert, link-run update, and an unqualified `FOR UPDATE` |
| `report_runs` | SELECT + INSERT | insert plus the replay read |

A DML-verb reading of the file finds no `UPDATE` against `saved_reports` and concludes SELECT is
enough — the same trap `scheduled_jobs` already reflects for the `job` alias in that same clause.

## Why the existing guard missed it

`fixed_engine_statement_privileges_integration_test.go` exists for exactly this, and its header
even names the `FOR UPDATE` trap. But it works from a **hand-maintained statement list** and had
zero report entries; the producer post-dates CHAOS-3114.

Its reasoning stops one step short. It argues the posture and the grant list "share an author and
a construction method, so they cannot check each other" — correct — but the **statement list also
shares an author with the code it audits**. A producer added without an entry is invisible: the
suite proves the statements it knows about can run, not that it knows all the statements.

## Changes

1. **Instance fix** — the three tables in `coordinatorPosture()`. GRANTs and readiness both follow.
2. **The five report statements** added to the integration suite, with
   `TestScheduledReportStatementsAreDeniedWithoutTheReportTableGrants` as their mutation control.
   It asserts the statements **pass before** the revocation, so the 42501s after it are
   attributable rather than incidental.
3. **Class fix** — `TestCoordinatorPostureCoversSchedulerSQLTables` derives every `public.<table>`
   named in `internal/scheduler` from the **source** and fails if any is absent from the posture.
   A producer added tomorrow either lands in the posture or fails the build. It proves coverage
   only; sufficiency stays with the integration suite, which runs the statements as the real
   restricted role.

## Verification

- Each of the three tables removed **individually** → the coverage test names exactly that table
  (checked per-table, not just in aggregate).
- The scan `t.Fatal`s if it reads zero files or finds zero tables, so a broken scan cannot report
  success.
- `PYTHON=.venv/bin/python bash ci/check_go.sh all` → **rc=0, 118 ok packages, 0 FAIL**.

## Deployment note

Applying this requires the pinned River migration to re-run so the coordinator GRANTs are
re-emitted. **A hand-run `GRANT` is not a valid workaround** — `CheckRolePosture` proves the role
holds the declared posture *"no more, no less, by any route"*, so granting an undeclared privilege
makes readiness **fail**.

Closes CHAOS-3935

### DIFF
--- a/internal/storage/postgres/coordinator_posture_scheduler_coverage_test.go
+++ b/internal/storage/postgres/coordinator_posture_scheduler_coverage_test.go
@@ -1,0 +1,132 @@
+package postgres
+
+import (
+	"os"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strings"
+	"testing"
+)
+
+// The coordinator posture must cover every table the scheduler's SQL names.
+//
+// fixed_engine_statement_privileges_integration_test.go already proves that the
+// statements IT LISTS can run as the real restricted role. That is the right
+// check, and it caught the FOR UPDATE and ON CONFLICT traps its header
+// describes. What it cannot catch is a producer nobody added to the list: the
+// statement set is hand-maintained, so it shares an author with the code it
+// audits, exactly the way the posture and the grant list used to share one.
+//
+// The scheduled-report producer (internal/scheduler/fixed/reports.go) was added
+// after that suite and never got entries. Its three tables -- saved_reports,
+// scheduled_report_occurrences, report_runs -- were therefore absent from
+// coordinatorPosture, both the readiness assertion and the derived GRANTs
+// agreed they were not needed, and every scheduled_reports_dispatch cycle in
+// production failed with:
+//
+//	permission denied for table saved_reports (SQLSTATE 42501)
+//
+// This test derives the table set from the source instead of from a list, so a
+// producer added tomorrow either lands in the posture or fails here. It
+// deliberately proves only COVERAGE, never sufficiency of privileges: whether
+// SELECT is enough or a row lock demands UPDATE is what the integration suite
+// executes for real. Coverage is the part a static check can own completely.
+func TestCoordinatorPostureCoversSchedulerSQLTables(t *testing.T) {
+	declared := make(map[string]struct{})
+	posture := CoordinatorPosture()
+	for _, table := range posture.RequiredTables {
+		declared[table.TableName] = struct{}{}
+	}
+	// A column-scoped grant is a deliberate refusal to grant table-wide, not an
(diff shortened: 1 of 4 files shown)


chrisgeo:
## Deployment ordering (verified against the OCI prod compose)

**Ordering is safe.** `go-river-migrate` runs `dev-health-worker-migrate && dev-health-worker-migrate --check`, and every Go worker declares `depends_on: go-river-migrate: service_completed_successfully`. The migration emits the new coordinator GRANTs before any readiness check runs, so there is no window where a worker boots against the old grant set.

**Non-obvious: the grants ship in the RUNNER image.** `go-river-migrate` uses `ghcr.io/full-chaos/dev-hops-runner:latest`, not `dev-health-go-worker` — the migrate binary is baked into the Python runner image by `docker/Dockerfile:87` (`COPY --from=go-migrator-builder /out/dev-health-worker-migrate`).

So this change requires **`dev-hops-runner` to be rebuilt**, not only the Go images. Rebuilding just the Go images would leave the old grant list in effect while the new binaries assert the tightened posture — `CheckCoordinatorAuthorization` would then fail and the scheduler, reconciler, and workerctl would all refuse to start. That is strictly worse than the current state, where only scheduled reports are broken.

`cmd/dev-health-worker-migrate/main.go:103-104` passes `CoordinatorRole` and `CoordinatorGrants` (derived from `CoordinatorPosture()`), so the rebuilt runner picks up the three new tables with no further wiring.

PR #1802, opened 2026-08-19
fix(contracts): resolve contract artifacts from installed distributions (CHAOS-3933)

chrisgeo:
## Summary

`provider_unit_route.py` and `dispatch_routes.py` located their contract JSON via `Path(__file__).resolve().parents[3] / "contracts" / ...`. The `parents[3]` count is the repo root in a source checkout, but resolves to `/usr/local/lib/python3.14` in an installed wheel, because the intermediate directory is `site-packages` rather than `src`. `pyproject.toml` shipped only `contracts/jobs/v1`, so even a correct resolver would have had nothing to find for the other two contracts.

This was silent until the Go sync-dispatch bridge moved dispatch onto the installed `api` service: every production `dispatch_sync_run` started returning 500 with `FileNotFoundError` on `contracts/provider-matrix/v1/matrix.json`.

Resolution and packaging are one contract — either half fixed alone still fails. This adds `contract_artifacts.contract_directory(*parts)`, a shared resolver anchored on the `dev_health_ops` PACKAGE directory rather than a per-call-site `parents[N]` count, so a module moving one level deeper cannot silently break resolution again. It prefers a checkout root and falls back to `sysconfig.get_path("data")/contracts/...`, matching where `[tool.setuptools.data-files]` now places `contracts/provider-matrix/v1` and `contracts/sync-dispatch/v1` in an installed distribution.

## Verification

- **Control A**: reverting `provider_unit_route.py` to `parents[3]` fails ONLY `test_loaders_resolve_from_the_installed_layout`; the other 7 tests in the new suite still pass. That is the proof a checkout-only test is structurally blind to this bug.
- **Control B**: removing the `pyproject.toml` data-files entries fails both `test_runtime_contract_tree_is_packaged` cases.
- **Empirical**: `uv build --wheel` produces a wheel containing `dev_health_ops-*.data/data/contracts/provider-matrix/v1/matrix.json` and `.../sync-dispatch/v1/transport-routes.json` — which install to `sysconfig`'s data path, exactly where the resolver's fallback looks.

## Gate note

Ran `ci/local_validate.sh` locally. Result:

```
GATE_STAGE_MANIFEST result=FAILED declared=8 executed=7 declared_ids=lint_format,lint_check,typecheck,ch_probe,ch_scratch_create,ch_migrate,unit_suite,ch_argmax_proof executed_ids=lint_format,lint_check,typecheck,ch_probe,ch_scratch_create,ch_migrate,unit_suite failed_at=unit suite (FULL, not subset)
```

- `lint_format`, `lint_check`, `typecheck`, `ch_probe`, `ch_scratch_create`, `ch_migrate` all PASSED.
- `unit_suite` FAILED: 20 pre-existing failures, all in `tests/test_sync_units.py` (route/family dispatch tests), unrelated to this diff.
- I confirmed these reproduce on unmodified `main` code: reverted the 3 changed source files to their pre-PR versions in this worktree and reran — identical 20 failures.
- `main`'s CI (GitHub Actions) is GREEN at the branch base commit — Lint, Typecheck, Integration Tests, and Live Backend E2E all succeeded there. So this is a local-environment-specific difference, not a broken `main`.
- This PR does **not** claim "gate passed" — the above is exactly what passed and what did not locally. PR CI is the adjudicating venue for this change.

## Operational note

The OCI host currently carries a STOPGAP bind-mount (`./contracts-overlay:/usr/local/lib/python3.14/contracts:ro` on the `api` service) that must be removed once this ships, along with `~/fullchaos.dev/contracts-overlay/`.

Closes CHAOS-3933

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -161,6 +161,21 @@ include-package-data = true
 ]
 "contracts/jobs/v1/examples" = ["contracts/jobs/v1/examples/*.json"]
 "contracts/jobs/v1/schemas" = ["contracts/jobs/v1/schemas/*.json"]
+# provider-matrix and sync-dispatch ship for the same reason contracts/jobs
+# does: dev_health_ops reads them at RUNTIME, not only in a checkout.
+# Omitting them made every production sync dispatch fail with a
+# FileNotFoundError once the Go bridge moved dispatch onto the installed
+# api service (CHAOS-3933). Keep this in step with the resolver fallbacks in
+# workers/provider_unit_route.py and sync/dispatch_routes.py -- packaging and
+# resolution are one contract, and either half alone still fails.
+"contracts/provider-matrix/v1" = [
+  "contracts/provider-matrix/v1/*.json",
+  "contracts/provider-matrix/v1/*.md",
+]
+"contracts/sync-dispatch/v1" = [
+  "contracts/sync-dispatch/v1/*.json",
+  "contracts/sync-dispatch/v1/*.md",
+]
 
 [tool.ruff]
 line-length = 88
--- a/src/dev_health_ops/contract_artifacts.py
+++ b/src/dev_health_ops/contract_artifacts.py
@@ -1,0 +1,57 @@
+"""Resolve checked-in contract artifacts in a checkout OR an installed distribution.
+
+Every consumer of a ``contracts/`` artifact needs the same two-step answer, and
+getting it wrong is silent until runtime. ``provider_unit_route.py`` resolved
+its matrix as ``Path(__file__).resolve().parents[3] / "contracts" / ...``, which
+is the repository root in a source checkout and ``/usr/local/lib/python3.14`` in
+an installed wheel, because the intermediate directory is ``site-packages``
+rather than ``src``. Nothing failed until the Go sync-dispatch bridge moved
+dispatch onto the installed ``api`` service, at which point every production
+``dispatch_sync_run`` returned 500 (CHAOS-3933).
+
+Two properties this module exists to hold:
+
+* The anchor is the PACKAGE directory, not the calling module. Counting
+  ``parents[N]`` per call site means the count is a function of where the file
+  happens to live, so moving a module one directory deeper breaks resolution
+  with no other symptom. ``dev_health_ops`` is always one level below ``src``.
+* Resolution and packaging are ONE contract. A path this function can compute
+  is still useless if ``pyproject.toml`` does not ship the tree, so the two are
+  documented against each other in both places.
+"""
+
+from __future__ import annotations
+
+import sysconfig
+from pathlib import Path
+
+# .../src/dev_health_ops -> parents[1] is the repository root in a checkout.
+# In an installed distribution this resolves to the interpreter's lib directory,
+# which will not contain "contracts", so the checkout branch simply misses and
+# the installed branch answers.
+_PACKAGE_ROOT = Path(__file__).resolve().parent
+_CHECKOUT_ROOT = _PACKAGE_ROOT.parents[1]
+
+
+def contract_directory(*parts: str) -> Path:
+    """Return a ``contracts/`` subdirectory, preferring a source checkout.
+
+    ``parts`` is the path BELOW ``contracts/`` -- for example
+    ``contract_directory("provider-matrix", "v1")``.
+
+    The checkout answer wins when it exists so that a developer editing a
+    contract sees the edit without reinstalling. The installed answer is
+    ``sysconfig.get_path("data")/contracts/...``, matching where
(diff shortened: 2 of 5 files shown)


chrisgeo:
## Correction to the Gate note above

The earlier note said the local full suite showed **20 pre-existing failures in `tests/test_sync_units.py`** that reproduce on unmodified main code. **That was wrong, and the cause was this PR.**

### What actually happened

`test_loaders_resolve_from_the_installed_layout` used `importlib.reload(provider_unit_route)` to recompute an import-time path under a patched checkout anchor. Reload installs a **new** module object while every earlier `from ... import` in the session keeps binding the old one, so the two diverge for the rest of the process.

Measured directly:

| Run | Result |
|---|---|
| `tests/test_sync_units.py` alone | **193 passed** |
| this file first, then `test_sync_units.py` | **1 failed**, 200 passed |
| after the fix (subprocess probe) | **201 passed** |

The casualty was `test_dispatch_sync_run_mixed_transport_routes_every_ready_pair_independently` — a test this PR never touches.

### Why the original control missed it

The control reverted the three modified source files and deleted `contract_artifacts.py`, then re-ran and saw the same failures. But it **left this test file in place**, so the polluting reload still ran. The failures duly reproduced "on main code" and looked like someone else's problem.

Reverting the *subject* of a change while leaving the *test* that causes the effect cannot distinguish the two.

Corroborating evidence that main was never broken: main's CI is green at `a5f171979` across Lint, Typecheck, Integration Tests and Live Backend E2E, and the failing test passes in isolation. A full-suite-only failure is an ordering signal, not a code signal.

### Fix

`ef61a6a02` runs the installed-layout probe in a **subprocess**. It proves the identical property — with no `contracts/` under the checkout anchor, both loaders must resolve under sysconfig's data path — while mutating nothing in the shared interpreter.

The negative control still fires: reverting the loader to `Path(__file__).resolve().parents[3]` fails **exactly** `test_loaders_resolve_from_the_installed_layout` and nothing else. That specificity is the point — the other seven tests in this file pass either way, which is why a checkout-only test cannot see this class of bug.

chrisgeo:
## Amendment: my correction above overclaimed

The correction said the 20 failures "were wrong, and the cause was this PR." That is **too strong**. There were **two independent problems**, and only one of them is this PR's.

### 1. The 20 failures — environment, not this PR

`ops/.env` (lines 159-160) sets `GO_PROVIDER_ROUTES=all` and `DEV_HEALTH_ENV=local`, pulled in by `ops/.envrc`'s `dotenv_if_exists .env` under direnv. A shell that passes through `ops/` before running a worktree's tests inherits both, and they survive the subsequent `cd <worktree> && pytest`.

`ProviderUnitRouteSwitches.from_environment()` sets
`local_all_routes = (GO_PROVIDER_ROUTES == "all" and DEV_HEALTH_ENV == "local")`
(`provider_unit_route.py:406-407`). When true, `_switch_enabled()` uses a "route everything to Go/River locally" alias table, so `routes_to_river()` returns True for any matrix-ready pair **regardless of the per-provider `WORKER_*_ENABLED` switch the tests exist to gate on**.

Concretely, in `test_dispatch_default_off_aggregate_route_has_only_the_celery_writer[gitlab-deployments]`: dispatch returned `{"status": "dispatched", "queued_units": 1}` as expected, but the celery `run_sync_unit.s().apply_async()` spy was `[]` — the unit went down the River/outbox path instead.

Verified by unsetting both vars: `env -u GO_PROVIDER_ROUTES -u DEV_HEALTH_ENV` over the same 24 cases → **24 passed, 0 failed**.

CI never sources `ops/.env`, which is why main is green. This is a local-dev convenience preset colliding with unit tests that assume a clean environment. **Not caused by this PR, and not fixed here.**

### 2. The reload pollution — this PR's, and real

Separately and genuinely mine: `importlib.reload` in the installed-layout test broke `test_dispatch_sync_run_mixed_transport_routes_every_ready_pair_independently`. Measured in a shell where **neither** env var was set — `env | grep -E '^(GO_PROVIDER_ROUTES|DEV_HEALTH_ENV)='` returned nothing — so this is a distinct effect, not a restatement of the above:

| Run (clean shell) | Result |
|---|---|
| `tests/test_sync_units.py` alone | 193 passed |
| this file first, then it | **1 failed**, 200 passed |
| after `ef61a6a02` | 201 passed |

One failure, not twenty. Fixed by moving the probe into a subprocess.

### Why both diagnoses looked like the whole story

Two shells, two different environments. The lane's shell inherited the direnv preset and saw 20 failures; mine did not and saw zero — until my own test file ran. Each of us then generalised from a single environment. Worth remembering the next time a "pre-existing failure" and a "my change broke it" explanation both fit: they can both be right about different things.

PR #1810, opened 2026-08-20
fix(deploy): stop the PgBouncer log firehose and size the River queue pool from a model that counts notifier sessions (CHAOS-3944, CHAOS-3945)

chrisgeo:
Lands in the tracked repo the two pooler fixes already applied and verified on the live host, plus the budget-model gap that let the second one happen silently.

## CHAOS-3944 — connection-log firehose, and a healthcheck probing a nonexistent user

PgBouncer defaults `log_connections=1` and `log_disconnections=1`. Against a worker fleet that cycles connections continuously that is a permanent firehose with no diagnostic value: **~36 lines/minute across the three production poolers, against 3/minute with them off**. All three now set `LOG_CONNECTIONS=0`, `LOG_DISCONNECTIONS=0`, `LOG_POOLER_ERRORS=1`.

`log_stats` is deliberately **kept on**, with a comment saying why so it does not get "tidied" away. That once-a-minute line is the only real telemetry PgBouncer emits without an exporter, and it is what made a live pool-exhaustion incident diagnosable: the queue pooler read `wait 5683209 us` (5.7 s mean connection wait) during the stall and `wait 16136 us` after recovery.

The three production poolers had **no healthcheck at all**. They now have one, naming each pooler's own role and database — and the comment says exactly what it proves. An earlier revision of this PR claimed `-U`/`-d` makes it a "can-serve" check. Measured against this image, it does not: `pg_isready` exits 0 for a nonexistent role and a nonexistent database alike, and exits non-zero only when nothing answers the port. What `-U`/`-d` buys is that the probe stops logging `no such user: postgres` every interval — which matters now that `LOG_POOLER_ERRORS=1` promotes that to a warning you want to stay meaningful.

A probe that really proved serving would have to open a session and run a query. That is deliberately rejected on the session-mode endpoints: the pool is sized to the declared fleet and has no spare slot for a probe, and under exactly the saturation such a probe is meant to detect it would block, time out, mark the pooler unhealthy and restart it — dropping every live session at the worst moment.

## CHAOS-3945 — the queue pool budget under-counted every River worker by a third

`internal/deploymentcontract` already computed declared client footprints and rejected non-positive headroom. It counted only each replica's queue-control pgxpool. A **started** River client also holds one long-lived notifier `LISTEN` session, outside that pool. That is how `devhealth_queue` came to sit at exactly `DEFAULT_POOL_SIZE` — leader election failing with `error beginning transaction: timeout: context deadline exceeded` — while every declared maximum in the manifest still validated.

The model is extended rather than replaced or duplicated. The notifier term is charged **per runtime**, traced to the code rather than assumed:

| runtime | opens queue pool | starts River client | queue cost / replica |
| --- | --- | --- | ---: |
| `river` (sync, heavy, ops, sync-provider) | yes | yes — `cmd/dev-health-worker/river_process.go:26` | 2 + 1 |
| `control` (reconciler, scheduler) | yes | no — `*Tx` APIs only | 2 |
| `stream` (external, ingest, pagerduty) | no | no | 0 |
| `workerctl` CLI | yes, short-lived | no | 2 |

Declared queue-session maximum is therefore **34, not 26**.

Headroom must also absorb a rolling restart. Reserving one replica would hold only for a serialized replacement — every process is rendered as its own independently rolled unit (Helm gives each its own Deployment with `maxSurge: 1`, `maxUnavailable: 0`), so one image or config change surges **all** of them together. The reserve is the sum across processes, and the same defect applied to the coordinator pool, which sat at headroom 1. That derives:

- queue `DEFAULT_POOL_SIZE = 34 + 16 = 50` (was 27)
- coordinator `DEFAULT_POOL_SIZE = 10 + 4 = 14` (was 11)
- declared server footprint 119 of 200

set consistently across the manifest, the production Compose file and the Helm values — the deployment-contract test binds all three.

`server_max_connections` rises 100 → 200. That is a requirement on a PostgreSQL the stack cannot read, and raising it silently would be its own defect, so it is now stated where an operator meets it — the production install page, the databases guide, and the Compose `.env` example — each with the `psql ... -Atc 'SHOW max_connections'` command to confirm it. A live admission check needs a runtime probe and is not attempted here.

`ADMIN_USERS` is now set on both River poolers to their own `DB_USER`. `admin_users` otherwise defaults to `postgres`, so `SHOW POOLS` / `SHOW CLIENTS` / `SHOW STATS` were refused on exactly the two endpoints where saturation matters — pool exhaustion had to be inferred from `pg_stat_activity` and an elector stack trace.

The prose budget comment is rewritten to enumerate every service by name and runtime, with the arithmetic shown. The previous version accounted for "two replicas of each of the three River worker groups plus the reconciler and scheduler" and went stale silently when `go-worker-sync-provider` and the three stream runners were added. It now also says explicitly that it is not the enforcement — the validator is.

## Guards added

- a river replica is charged exactly one notifier session and control/stream/CLI clients are charged none;
- a manifest declaring zero notifier sessions is rejected, so the budget cannot quietly revert to pool-only arithmetic;
- a session pool that cannot absorb a fleet-wide rolling restart is rejected — headroom of one is not headroom.

## Review

An adversarial review of the first draft returned three findings, all real, all addressed above: the false "can-serve" claim, the single-replica surge reserve, and the undocumented `max_connections` requirement. It cleared the runtime tracing behind the notifier arithmetic (reconciler constructs a River client but with no queues, so River 0.40's `willExecuteJobs()` is false and no notifier or elector is created), confirmed `MinConns` is 0 so there is no hidden pool floor, and confirmed the new tests are mechanism-sensitive rather than outcome-shaped.

## Deliberately not changed

`deploy/helm/dev-health/templates/go-pgbouncer.yaml` probes are `tcpSocket`, which has the identical port-open-is-not-can-serve weakness. Changing them has a different blast radius — a failing liveness probe restarts the pooler and drops every session on a session-mode endpoint — so the limitation is recorded in a comment referencing CHAOS-3944 and left for its own change.

Live pool **occupancy** is also out of scope here. This model proves declared maxima, which is a build-time property of checked-in files; occupancy is a runtime property and needs a scrape of the PgBouncer admin console (now reachable, which was the prerequisite). That belongs in an observability change, not in a manifest validator that has no runtime to observe.

### DIFF
--- a/cmd/dev-health-workerctl/main_test.go
+++ b/cmd/dev-health-workerctl/main_test.go
@@ -347,13 +347,18 @@ func TestManifestQueueStatusSourceCombinesFreshQueueAndPresenceState(t *testing.
 		t.Fatalf("custom queue status = %#v", status.Groups[1])
 	}
 	// sync-provider is its own declared process (CHAOS-3926), so its
-	// max_replicas x per-process connections count separately from sync: queue
-	// session 22 -> 26 against a pool raised 23 -> 27, server 89 -> 93. Both
-	// processes ship at zero replicas, so runtime usage is unchanged; this is
-	// the declared worst case the budget has to cover.
-	if status.ConnectionBudget.QueueSession != (connectionBudgetStatus{Used: 26, Limit: 27, Headroom: 1}) ||
+	// max_replicas x per-process connections count separately from sync.
+	//
+	// CHAOS-3945 then raised the queue-session figure again, from 26 to 34:
+	// a started River client holds one notifier LISTEN session outside its
+	// queue-control pool, so each of the four "river" runtime processes costs
+	// 3 per replica, not 2. The pool went 27 -> 37, which is that declared 34
+	// plus one whole replica of rolling-restart overlap, and the server
+	// requirement 100 -> 200. Every process ships at zero replicas, so runtime
+	// usage is unchanged; this is the declared worst case the budget covers.
+	if status.ConnectionBudget.QueueSession != (connectionBudgetStatus{Used: 34, Limit: 37, Headroom: 3}) ||
 		status.ConnectionBudget.CoordinatorSession != (connectionBudgetStatus{Used: 10, Limit: 11, Headroom: 1}) ||
-		status.ConnectionBudget.Server != (connectionBudgetStatus{Used: 93, Limit: 100, Headroom: 7}) {
+		status.ConnectionBudget.Server != (connectionBudgetStatus{Used: 103, Limit: 200, Headroom: 97}) {
 		t.Fatalf("connection budget = %#v", status.ConnectionBudget)
 	}
 }
--- a/cmd/dev-health-workerctl/main_test.go
+++ b/cmd/dev-health-workerctl/main_test.go
@@ -352,13 +352,17 @@ func TestManifestQueueStatusSourceCombinesFreshQueueAndPresenceState(t *testing.
 	// CHAOS-3945 then raised the queue-session figure again, from 26 to 34:
 	// a started River client holds one notifier LISTEN session outside its
 	// queue-control pool, so each of the four "river" runtime processes costs
-	// 3 per replica, not 2. The pool went 27 -> 37, which is that declared 34
-	// plus one whole replica of rolling-restart overlap, and the server
-	// requirement 100 -> 200. Every process ships at zero replicas, so runtime
-	// usage is unchanged; this is the declared worst case the budget covers.
-	if status.ConnectionBudget.QueueSession != (connectionBudgetStatus{Used: 34, Limit: 37, Headroom: 3}) ||
-		status.ConnectionBudget.CoordinatorSession != (connectionBudgetStatus{Used: 10, Limit: 11, Headroom: 1}) ||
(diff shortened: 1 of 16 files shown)


chrisgeo:
Superseded by a fresh PR from the same branch: GitHub stopped delivering `pull_request` events to this PR after a close/reopen I used to retrigger a stuck run, so it can no longer report checks. Same branch, same commits.

chrisgeo:
Superseded by **#1813**, which is this same branch rebased onto current `main` and reduced to the additive delta over #1809.

Nothing is lost. #1810 carried duplicate compose/healthcheck work that #1809 had already landed — and landed better (`STATS_USERS` rather than `ADMIN_USERS`; a real `psql` healthcheck rather than `pg_isready`). Everything #1810 had that #1809 did not is in #1813: the `internal/deploymentcontract` budget model that counts River notifier sessions, the fleet-wide rolling-restart reserve for both session pools, and the `docs/operate/` pages including the `server_max_connections` requirement an operator has to meet.

Closed rather than rebased because GitHub had stopped delivering `pull_request` events to this PR, so it could no longer report checks.

PR #1811, opened 2026-08-20
test(worker): assert outage recovery converges instead of a 90s ceiling (CHAOS-3959)

chrisgeo:
## Problem

`TestMultiReplicaFleetSurvivesDatabaseOutage` cut the database for 15s and then
asserted recovery against a fixed 90-second wall clock. On a slower runner it
failed inside a shared integration shard, reddening PRs whose diffs cannot
reach `cmd/dev-health-worker` at all (CHAOS-3959: two consecutive failures on a
PR touching only `internal/providersync`, which is a different shard target).

## Why the ceiling could not work

Recovery from a mid-flight outage is not one operation, it is a cascade of
River-scheduled steps. Verified against the real test with River's logger
turned up:

1. The waiting handler fails acquiring the fleet concurrency lease the moment
   the link is cut (`result=discard error_category=budget`).
2. River tries to persist that error, and the `BatchCompleter` cannot reach the
   database either. It retries three times (~7s of backoff, up to ~37s if the
   attempts also time out), logs `Too many errors; giving up`, **requeues the
   batch**, and starts over. Until one of those rounds lands, the job sits in
   `running` with no error recorded — exactly the state CI reported.
3. Once the error lands the job becomes retryable, scheduled at `attempt^4`
   seconds after the previous attempt: 1s, 16s, 81s.
4. The retried handler then queues behind the fleet lease that the parked
   handler still holds.

How many of those intervals are in play depends on where inside the outage
window each transition fell, which is a function of runner speed. 90s was built
from those same constants, so it could not distinguish *recovery did not
happen* from *recovery has not happened yet*.

No contract states a recovery-latency bound. `contracts/jobs/v1` gives
`system.heartbeat` a 30s per-attempt timeout and an attempt count; there is no
fleet-recovery SLO, so the test keeps no latency assertion and the test name
stays honest about what it asserts.

## Change

`waitForCompletedAfterOutage` now waits for the terminal shape and makes a
*stalled* cascade the failure signal:

- **Stall (2m)** — no job changed state, attempt, or errors for longer than the
  longest interval River schedules (the 81s third-attempt backoff). A lost
  transition freezes the rows and trips this.
- **Budget (4m)** — backstop for a cascade that keeps churning without ever
  converging.

Both messages carry the full job snapshots and say which case fired, so a real
regression is diagnosable rather than a test that hangs. The test context goes
6m → 8m so the backstop, not a context expiry, is what reports a failure.

## Verification

`ci/check_go.sh all` never runs integration tests (`check_integration_vet` only;
CHAOS-3948), so this was run explicitly.

Fixed test, three consecutive runs, `rc=0` each:

```
--- PASS: TestMultiReplicaFleetSurvivesDatabaseOutage (51.68s)
--- PASS: TestMultiReplicaFleetSurvivesDatabaseOutage (51.78s)
--- PASS: TestMultiReplicaFleetSurvivesDatabaseOutage (52.41s)
```

(pre-change baseline on the same machine: 52.47s — the change costs nothing on
the happy path.)

Both directions were proven with throwaway probes that reuse the real fixture
and are **not** part of this diff:

- **Recovery genuinely does not happen** — cut the database and never restore
  it. The new assertion fails, on the stall branch, naming the frozen states.
- **Recovery is merely slow** — hold the database down for 100s past release,
  i.e. the CI case. The old 90s ceiling fails; the new assertion passes.

See the PR discussion for the captured output of each.

### DIFF
--- a/cmd/dev-health-worker/multi_replica_outage_integration_test.go
+++ b/cmd/dev-health-worker/multi_replica_outage_integration_test.go
@@ -4,6 +4,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"log/slog"
 	"net"
@@ -11,6 +12,7 @@ import (
 	"net/http/httptest"
 	"net/url"
 	"path/filepath"
+	"strings"
 	"sync"
 	"testing"
 	"time"
@@ -44,7 +46,13 @@ const workerPresenceTTLProof = 35 * time.Second
 // first failed presence heartbeat was fatal to the process.
 func TestMultiReplicaFleetSurvivesDatabaseOutage(t *testing.T) {
 	t.Chdir(filepath.Join("..", ".."))
-	ctx, cancel := context.WithTimeout(context.Background(), 6*time.Minute)
+	// Wide enough that the outage-recovery backstop below, not this context, is
+	// always what reports a fleet that never recovers: a context expiry here
+	// would surface as an unrelated query error from the admin pool. The worst
+	// passing path is setup plus the 15s outage, the full 4m budget, the 35s
+	// presence proof and the drain, so this leaves several minutes of slack on
+	// top of it. `go test -timeout` remains the outer guard.
+	ctx, cancel := context.WithTimeout(context.Background(), 12*time.Minute)
 	t.Cleanup(cancel)
 
 	postgres, err := containers.StartPostgres(ctx)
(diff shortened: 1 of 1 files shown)


chrisgeo:
## Verification

`ci/check_go.sh all` never runs integration tests (`check_integration_vet` only — CHAOS-3948), so this was run explicitly. All local runs on the same machine.

### Fixed test, happy path

| run | result |
| -- | -- |
| pre-change baseline | `PASS 52.47s` |
| fixed, run 1 | `PASS 51.78s` rc=0 |
| fixed, run 2 | `PASS 51.68s` rc=0 |
| fixed, run 3 | `PASS 52.41s` rc=0 |
| fixed, after review fixes | `PASS 51.53s` rc=0 |

The change costs nothing on the happy path.

### Both directions

Proven with throwaway probes that reuse the real fixture and are **not** part of this diff.

**1. Recovery is merely slow — the CI case.** Ordinary 15s outage, both original jobs recover normally, then one more heartbeat is inserted with a future `ScheduledAt` so the job set needs ~101s to reach its terminal shape. No fault anywhere; every job succeeds.

Old 90s ceiling — **FAIL**, rc=1, the CI message verbatim, on a job that had simply not run yet:

```
jobs still incomplete 1m30s after release, past River's completer retry ceiling:
  1 {State:"completed"} 2 {State:"completed"} 3 {State:"scheduled", Attempt:0}
--- FAIL: TestProbeSlowRecoveryOldCeiling (106.94s)
```

(106.94s total against the 107.87s and 108.23s CI failures.)

New assertion, same shape — **PASS**, rc=0:

```
convergence assertion returned after 1m40.997002459s
--- PASS: TestProbeSlowRecoveryConvergence (117.48s)
```

**2. Recovery genuinely does not happen.** Cut the database and never restore it — **FAIL**, rc=1, on the lost-transition branch, naming the frozen states:

```
outage recovery never started: no job changed state, attempt, or error in the 4m0s
since release -- the transition was lost, not delayed:
  1 {State:"running", Attempt:1} 2 {State:"available", Attempt:0}
--- FAIL: TestProbePermanentOutage (267.24s)
```

**3. A fleet that genuinely fails still fails.** A 115s outage blows the 30s contract timeouts and discards both jobs; the new assertion does not rubber-stamp it — it reports the other branch, with the discard errors:

```
outage recovery never converged in 4m0s: jobs were still changing (last change 32s ago)
but 0 of 2 reached completed -- the cascade is progressing without terminating:
  ... State:"discarded", Attempt:3, ... "dev-health job failed [timeout]" ...
```

### Mechanism, confirmed with River's logger at debug

The waiting handler fails acquiring the fleet concurrency lease the moment the link is cut (`result=discard error_category=budget`); River then cannot persist that error either, and the BatchCompleter logs `Too many errors; giving up`, requeues the batch, and starts over. Until a round lands, the job sits in `running` with no error recorded — exactly the state CI reported.

### Review

Adversarially reviewed with `codex exec -m gpt-5.6-luna -c model_reasoning_effort=xhigh`. Verdict: approve with findings. Two were applied — the comment overstated River's ownership of the fleet lease (it is ours, `internal/jobruntime/budget_postgres.go`) and cited an 81s third-attempt backoff these jobs can never reach, since `MaxAttempts=3` caps them at the 1s and 16s retries; and the admin-pool error paths now name the job and how far into the wait they failed. The remaining finding — that the test no longer bounds latency, so a regression completing at 3m30s would pass — is the deliberate trade, since no contract states a recovery-latency bound.

chrisgeo:
## Adversarial review — dispositions

Three passes with `codex exec -m gpt-5.6-luna -c model_reasoning_effort=xhigh` (a first review, a re-review of the reworked assertion, and a verification pass on the fixes).

**Applied**

1. The comment claimed River owned the fleet concurrency lease. It is ours — `internal/jobruntime/budget_postgres.go`, a 100ms poll inside the job's 30s contract deadline. Reworded.
2. The comment derived the budget from an 81s third-attempt backoff that these jobs can never reach: `MaxAttempts=multiReplicaRetryAttempts` caps them at the ~1s and ~16s retries and discards the third failure. The derivation now runs off the completer's ~38s give-up round instead, which is the real longest quiet gap.
3. The retry schedule needed one more step of provenance: our adapter supplies River's optional `NextRetry`, and `jobruntime.NextRetryAt` returns a zero time for any contract whose `retry_policy` is not `bounded_exponential_jitter` — `system.heartbeat`'s is `none` — so River falls back to its own schedule. Stated explicitly, so a later reader does not "correct" it to the adapter's 5s·2^(n-1) ladder.
4. The ~38s bound was asserted unconditionally; completer rounds actually repeat for as long as writes keep failing. Qualified against `databaseOutageWindow`, which fixes that at 15s, and noted that a genuinely stranded job is not waited out — River's rescuer only reclaims after an hour, far outside the budget.
5. The test context was an independent earlier deadline. Raised to 12m so the recovery budget is always what fires first.
6. Admin-pool query failures aborted with a bare error. They now name the job and how far into the wait they failed.

**Accepted, not changed**

- *The test no longer bounds latency; a regression completing at 3m30s would pass.* That is the deliberate trade. No contract states a recovery-latency bound — `contracts/jobs/v1` gives `system.heartbeat` a 30s per-attempt timeout and an attempt count, nothing about fleet recovery — and the surrounding assertions (zero terminal runs, zero cancelled/discarded, exactly one effect per job, presence live past its TTL, clean drain) are unchanged, so no semantic invariant was relaxed.
- *12m is not provably enough on an arbitrarily slower runner.* Worst passing path measured here is ~18s of setup + 15s outage + up to 4m budget + 35s proof + drain. Tripling the variable part still leaves several minutes of slack, and `go test -timeout=30m` remains the outer guard.
- *"Requeues and starts over" applies to any retryable completer error, not only an unreachable database.* True in general; in this test the cut proxy is the only source of them.

**Out of scope, flagged**

The `databaseOutageWindow` comment (unchanged by this PR) says 15s is "long enough for several budget and idempotency renewal attempts to fail". Neither renewer ticks in 15s: budget renews every 5m from a 15m lease and idempotency every 3m20s from a 10m lease. Only worker presence, at 10s, renews inside the window — which is the failure the test actually asserts on. Left alone here.

PR #1815, opened 2026-08-20
test(tooling): bump the providersync top-level-test count pin to 934 (CHAOS-3946)

chrisgeo:
## Summary
- `tests/tooling/test_go_integration_sharding.py` pins the live-discovered count of `internal/providersync` top-level tests (was 921) and integration-tagged tests (was 107) as a sentinel that forces someone to re-examine `ci/go_providersync_test_shards.tsv`'s weighting whenever the package's test surface grows.
- The pin silently drifted. Five commits merged into main since 08-18 added providersync tests without bumping it: `15e02aa7e` (+3), `84fe7690d` (+3), `d360676a7` (+3), `803041649` (+1), `206d60a49` (+3) — 13 tests total, 921 → 934.
- Nothing caught this because `go.yml`'s push trigger has been commented out since `f5e5ee4ff` (Actions concurrency budget) and it isn't a required check — it only runs against PRs that cross both a Go path filter and touch this Python test file. Main has been red-in-hiding on this specific assertion.
- Resolves an open question about a 931-vs-934 discrepancy seen across two measurements: it is **not** a platform/build-tag difference. 931 = 921 + 10 is the count after four of the five commits above had landed; 934 = 921 + 13 includes the fifth (`206d60a49`), merged later. Same discovery logic, different git commit.
- The package-level manifest `ci/go_integration_shards.tsv` needs no change — `check_go.sh` validates its package set against live discovery every run and the set is unchanged at 26 packages.
- Regenerated by rerunning `bash ci/check_go.sh integration-shard-plan` against current `origin/main` and transcribing its live counts — the same method used when this pin was first introduced in #1738. There is no separate generator script for these two literals.

## Test plan
- [x] `uv run pytest tests/tooling/test_go_integration_sharding.py -q` → 8 passed

### DIFF
--- a/tests/tooling/test_go_integration_sharding.py
+++ b/tests/tooling/test_go_integration_sharding.py
@@ -243,8 +243,8 @@ def test_shard_plan_is_exhaustive_nonempty_and_machine_readable(
 
     expected_provider_tests = _providersync_top_level_tests()
     expected_integration_tests = _providersync_integration_tagged_tests()
-    assert len(expected_provider_tests) == 921
-    assert len(expected_integration_tests) == 107
+    assert len(expected_provider_tests) == 934
+    assert len(expected_integration_tests) == 108
     assert expected_integration_tests < expected_provider_tests
 
     provider_assignments: dict[int, set[str]] = {}
@@ -260,7 +260,7 @@ def test_shard_plan_is_exhaustive_nonempty_and_machine_readable(
     provider_flattened = [
         test_name for tests in provider_assignments.values() for test_name in tests
     ]
-    assert len(provider_flattened) == len(set(provider_flattened)) == 921
+    assert len(provider_flattened) == len(set(provider_flattened)) == 934
     assert set(provider_flattened) == expected_provider_tests
     assert {
         name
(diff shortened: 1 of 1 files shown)


chrisgeo:
Superseded — folded into PR #1813 as a separate commit (eed30bf8a) per team direction, since 1813 is the Python-touching PR already carrying this fix through review. `test`/`test-matrix (3.14)` both passed here before closing, confirming the fix itself is correct; the only red check (`lint-job`) is unrelated pre-existing drift (unpinned `ruff` in lint.yml vs the lockfile version) and not caused by this change.

chrisgeo:
Reopening per updated sequencing: ruff-pin fix lands first (separate tiny PR), then this PR's lint-job re-runs against it, then this merges, then 1813 rebases to pick both up.

chrisgeo:
Re-closing — superseded again, this time for good: 1813 carries the pin commit (eed30bf8a) directly, so merging this too would double-apply the same literals. Leaving closed as the validation record (test/test-matrix passed here before the fold-in).

